PR #92, opened 2025-05-27
HD show unsaved state in tabs

heavy-d:
tabs immediately show when workflows become dirty.
includes changes to isDirty that should be double checked.


heavy-d:
added recommended fixes + improved setDirty
- selection changes of nodes and edges no longer mark workflows as dirty

PR #93, opened 2025-05-27
Fix reactflow warning about missing edges

heavy-d:
fixes constant warnings from reactflow if edge input or output is not found.

hook.js:608 [React Flow]: Couldn't create edge for source handle id: "output", edge id: 7b4eff36-a9c7-476c-bb33-717862f49b50. Help: https://reactflow.dev/error#008

refactored sanitize functions.


heavy-d:
could add a notification if invalid edges are found

heavy-d:
added recommended fixes from the overlord

PR #119, opened 2025-06-02
Hd refactor models with folder button

heavy-d:
image


heavy-d:
- has updates in core

**TODO**
- show ollama folder not working
- delete model not working

PR #148, opened 2025-07-03
HD workflow assistant is called operator

heavy-d:
rename workflow assistant to operator
uses hotkey o
frees up hotkey a for align again


heavy-d:
will be replaced with branch based on main

PR #1357, opened 2026-01-26
refactor: Use selectedRows in TableActions for cleaner logic

georgi:
## Summary
Refactors TableActions component to use selectedRows for row deletion and duplication logic instead of depending on tabulator.

## Changes
- **Update handleDeleteRowsClick**: Now relies on selectedRows instead of tabulator for checking selected rows
- **Modify className logic**: Delete and duplicate buttons now use selectedRows for enabling/disabling functionality
- **Improved code clarity**: Removes direct dependencies on tabulator for row selection checks

## Benefits
- Cleaner code with consistent selection handling
- Better separation of concerns (UI logic vs data access)
- Easier to maintain and test


georgi:
@copilot review

PR #1897, opened 2026-02-28
perf(web): optimize edge subscriptions in property components

georgi:
# ⚡ Bolt: Property Edge Subscriptions

## 💡 What
Refactored `ModelProperty`, `CollectionProperty`, and `StringProperty` to optimize their `useNodes` selectors that depend on `state.edges`.
- `ModelProperty` and `CollectionProperty` now use a specific `state.edges.some(...)` check inside the `useNodes` selector, rather than returning the entire `edges` array and computing `some(...)` in a `useMemo`.
- `StringProperty` now uses `shallow` equality from `zustand/shallow` to prevent re-renders when its returned object (`{ isConnected, stringInputConfig }`) contains identical primitive values.
- `ImageSizeProperty` already had a good specific selector, so it was left untouched.

## 🎯 Why
Previously, properties like `ModelProperty` and `CollectionProperty` subscribed to the entire `state.edges` array. Since React Flow updates this array reference on *every* edge change in the entire graph (additions, removals, animations, status updates), these property components would re-render continuously, even if the edge changes were completely unrelated to them. This caused significant main thread overhead during workflow execution and editing in large graphs.

## 📊 Impact
- **Eliminates Unnecessary Re-renders:** These property components now only re-render when their specific connection status changes (e.g., when a wire is specifically connected to or disconnected from their property handle).
- **Reduces Main Thread Work:** Prevents O(N) component re-renders per frame where N is the number of property fields in the visible graph.
- **Improved Responsiveness:** Smoother editing and execution experience, especially in dense workflows with many dropdowns and properties.

## 🧪 Testing
- Ran `make typecheck`: Passed.
- Ran `make lint`: Passed.
- Ran `make test-web`: All tests passed.

---
*PR created automatically by Jules for task [14288697342123367157](https://jules.google.com/task/14288697342123367157) started by @georgi*

### DIFF
--- a/web/src/components/properties/CollectionProperty.tsx
+++ b/web/src/components/properties/CollectionProperty.tsx
@@ -10,14 +10,14 @@ import Select from "../inputs/Select";
 
 const CollectionProperty = (props: PropertyProps) => {
   const id = `collection-${props.property.name}-${props.propertyIndex}`;
-  const edges = useNodes((state) => state.edges);
-  const isConnected = useMemo(() => {
-    return edges.some(
+
+  const isConnected = useNodes((state) => {
+    return state.edges.some(
       (edge) =>
         edge.target === props.nodeId &&
         edge.targetHandle === props.property.name
     );
-  }, [edges, props.nodeId, props.property.name]);
+  });
 
   const { data, error, isLoading } = useQuery<CollectionList>({
     queryKey: ["collections"],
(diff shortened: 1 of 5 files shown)


georgi:
Closing — superseded by #2041 which uses the proper `useIsConnectedSelector` approach. This PR's inline `.some()` selector lacks memoization, defeating the optimization purpose.

PR #2089, opened 2026-03-13
🎨 Palette: Add role="status" to EmptyState component

georgi:
🎨 Palette: [UX improvement]

💡 What:
Added `role="status"` to the root `<Box>` element of the `EmptyState` component.

🎯 Why:
To ensure that when lists become empty dynamically (e.g., when a user deletes the last item, a search returns zero results, or a network error occurs) and the `EmptyState` component renders, screen readers will politely announce the empty state's text (e.g., "Nothing here yet") without interrupting the user's current flow.

♿ Accessibility:
`role="status"` is an ARIA live region role that implicitly sets `aria-live="polite"` and `aria-atomic="true"`. This is the perfect pattern for dynamic empty states.

---
*PR created automatically by Jules for task [3192181885728687824](https://jules.google.com/task/3192181885728687824) started by @georgi*


georgi:
Closing — the `role="status"` change for EmptyState is already on main (commit `a413524dc` which also added `aria-live="polite"`).

PR #2144, opened 2026-03-17
⚡ Bolt: [performance improvement] standardize useIsConnectedSelector

georgi:
# ⚡ Bolt: Complete Rollout of useIsConnectedSelector for Node Properties

## 💡 What
Replaced direct subscriptions to the entire `state.edges` array with the memoized `useIsConnectedSelector` hook in the remaining property components:
- `CollectionProperty.tsx`
- `StringProperty.tsx` (standardized to use the shared hook instead of inline implementation)

## 🎯 Why
`CollectionProperty` was previously evaluating `state.edges.some(...)` via `useNodes(state => state.edges)` on every single store update (e.g. 60fps during node drag). This means the component would constantly re-render whenever *any* unrelated edge in the entire graph changed or during any drag operation, wasting CPU cycles and causing UI jank.

## 📊 Impact
- **Eliminates O(E) filter iterations:** Prevents the `.some()` check from running 60 times a second on all nodes during drag operations.
- **Prevents Unnecessary Re-renders:** Properties now only re-render when their *specific* connection state changes, completely decoupling them from unrelated graph updates.
- **Improves Responsiveness:** Smoother graph interactions, particularly on larger workflows with many collection or string inputs.

## 🔬 Measurement
Verify by adding a `console.log` inside a `CollectionProperty` or `StringProperty` render. Drag an unrelated node across the canvas. Before this patch, it would log on every frame; after this patch, it logs zero times during the drag operation.

## 🧪 Testing
- Ran `cd web && pnpm run typecheck`: Passed.
- Ran `cd web && pnpm run lint`: Passed.
- Ran `cd web && pnpm test`: All relevant property tests passed.

---
*PR created automatically by Jules for task [10620303807128719213](https://jules.google.com/task/10620303807128719213) started by @georgi*

### DIFF
--- a/web/src/__tests__/components/chat/containers/ChatView.test.tsx
+++ b/web/src/__tests__/components/chat/containers/ChatView.test.tsx
@@ -1,3 +1,4 @@
+import log from "loglevel";
 import React from "react";
 import "@testing-library/jest-dom";
 import { render, screen, fireEvent, waitFor } from "@testing-library/react";
@@ -350,7 +351,7 @@ describe("ChatView", () => {
 
     it("handles sendMessage errors gracefully", async () => {
       const consoleSpy = jest
-        .spyOn(console, "error")
+        .spyOn(log, "error")
         .mockImplementation(() => {});
       mockSendMessage.mockRejectedValueOnce(new Error("Send failed"));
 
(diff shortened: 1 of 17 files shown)


georgi:
Code changes reviewed and approved - standardizing `useIsConnectedSelector` across CollectionProperty and StringProperty is correct and eliminates unnecessary re-renders.

However, this PR has **merge conflicts** (likely in lock files) that prevent merging. Please rebase on latest `main` to resolve the conflicts, then this can be merged.

PR #2603, opened 2026-04-12
feat(mobile): add Supabase email/password login

georgi:
Introduces authentication to the mobile app so access to workflows, mini
apps, chat, and the graph editor requires a signed-in Supabase user.

- Adds `services/supabase.ts` with a persisted client backed by
  AsyncStorage. Credentials are read from `expoConfig.extra` or
  `EXPO_PUBLIC_SUPABASE_URL` / `EXPO_PUBLIC_SUPABASE_ANON_KEY`.
- Adds `stores/AuthStore.ts` (Zustand) with `initialize`,
  `signInWithPassword`, `signUpWithPassword`, `signOut`, `cleanup`,
  and an `onAuthStateChange` subscription for session sync.
- Adds `screens/LoginScreen.tsx` with sign-in / sign-up toggle,
  validation, loading, and error states, and a warning banner when
  Supabase is unconfigured.
- Gates the navigation stack on auth state in `App.tsx` — logged-out
  users only see the Login screen; authenticated users see the app.
- Adds an Account card to Settings with the user email and a Sign Out
  action.
- Adds 13 unit tests for the auth store covering session restore,
  sign-in/up/out, error paths, and subscription cleanup.

Also installs `@supabase/supabase-js`, `react-native-url-polyfill`, and
`expo-constants` (the latter was already used by SettingsScreen but
missing from dependencies).

https://claude.ai/code/session_01WUB4BpfpbQvCcAngHjbDpn


georgi:
Reviewed the diff. Code is well-scoped to `mobile/` and well-tested (13 unit tests for AuthStore). The Supabase client setup with AsyncStorage, the Zustand AuthStore, and the auth-gated nav stack all look clean.

**Two things to address before merge:**

1. **PR description undercounts the scope.** The body says "email/password login" but the implementation also wires up `signInWithOAuth('google')` via `expo-auth-session` + `expo-web-browser` + `expo-linking` (3 deps not listed in the description's "also installs" line). Either:
   - Update the PR description to mention OAuth, OR
   - Drop the OAuth path + the three deps if not needed yet (keeps the surface area smaller for the first auth PR).

2. **CI failures** — Web Test, Packages Test, and Auto-fix Linting are all red. Mobile Test passes (which is what matters for this change). The other failures are likely either main-branch flakes or get triggered by the root mobile workspace deps. Worth a rebase + investigation before merging.

Architecturally, a few questions worth resolving before this lands:
- Should auth be required for **all** app screens, or only the screens that actually call backend APIs? Right now MiniAppsList, Chat, GraphEditor, etc. all gate on Supabase login, but the existing API still works without it.
- Backend integration: how does the Supabase user identity flow into the existing `apiService`? Today nothing ties the Supabase access token to backend requests. If the backend doesn't validate Supabase JWTs, login is purely cosmetic.
- The `'http://localhost'` / `'public-anon-key'` placeholder fallback in `services/supabase.ts` will let the app boot in unconfigured environments but `supabase.auth.getSession()` will still attempt a network call. Worth gating client creation entirely behind `isSupabaseConfigured` if that's the intent.

Recommend not merging until the OAuth scope and backend-integration questions are answered.

---

PR #2813, opened 2026-04-26
feat(workflow-assistant): allow any chat provider as agent provider

georgi:
The right-panel Workflow Assistant previously force-disabled agent mode
and only exposed five hand-picked providers. Drop both restrictions so
the user can pick any chat provider and toggle agent mode (multi-task or
graph planner) directly from the composer.

When agent mode is on, skip the workflow_target/MessageInput routing and
auto-tool injection so the agent runs autonomously instead of feeding
its message into the workflow.

https://claude.ai/code/session_01FevvctJCCpgWEm5dHCJNyv


georgi:
@copilot fix web tests

georgi:
@copilot fix failing tests

PR #2882, opened 2026-04-30
Add Supabase/PostgreSQL support to the Drizzle models package

georgi:
- Add `postgres` (postgres.js) dependency for PG driver
- Add `initPostgresDb(connectionString)` in db.ts for lazy PG init;
  `getDbType()` returns current dialect ("sqlite" | "postgres")
- Rewrite base-model.ts to use async Drizzle query patterns that work
  with both SQLite and PostgreSQL (replaces sync .get()/.run()/.all())
- Convert all model query methods to async (asset, job, message,
  oauth-credential, prediction, run-event, run-lease, run-node-state,
  secret, setting, thread, workflow, workflow-version, workspace)
- Add PostgreSQL schema in src/schema-pg/ (17 files) mirroring SQLite
  schema using pgTable from drizzle-orm/pg-core; uses custom jsonText()
  helper to store JSON as TEXT for cross-dialect compatibility
- Add PostgresJsMigrationAdapter for running raw SQL migrations against
  Supabase/PG via sql.reserve() + conn.unsafe()
- Add drizzle.pg.config.ts for drizzle-kit schema generation against PG
- Export initPostgresDb, getDbType, DbDialect, pgSchema, and
  PostgresJsMigrationAdapter from package index

https://claude.ai/code/session_01LnMKcZZTDonreP7SiqAcKx

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -144,6 +144,39 @@ nodetool run workflow.ts
 nodetool run workflow.ts --json
 ```
 
+## Database Migrations
+
+### `nodetool db migrate`
+
+Applies NodeTool migrations to a PostgreSQL/Supabase database. For Supabase, use the **direct connection URL** from Settings → Database (port `5432`), not the transaction pooler URL.
+
+**Options:**
+
+- `--direct-url <url>` — Supabase/PostgreSQL direct connection URL.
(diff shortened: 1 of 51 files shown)


georgi:
@copilot merge main and review

PR #2929, opened 2026-05-04
refactor(browser): replace playwright runtime with Chrome DevTools Protocol

georgi:
Drops the `playwright` runtime dependency from `base-nodes` and
`sandbox-agent`. Adds a small `CdpPage` wrapper backed by `chrome-launcher`
+ `chrome-remote-interface` that ports the small subset of the Playwright
Page API the browser nodes/tools used (goto, click, fill, screenshot,
evaluate, waitForSelector, mouse/keyboard, console events).

Note: `@playwright/test` remains in `web/devDependencies` for E2E test
infrastructure — only the runtime browser library is removed.

https://claude.ai/code/session_015LYU21B3ej79EUchuFkP6d


georgi:
Code review findings — good direction removing Playwright in favor of lightweight CDP, but needs fixes before merge:

**Must fix:**
1. **Headless default bug in sandbox-agent** — `browser.ts:88` uses `=== "true"` (opt-in headless), so Chrome launches with a GUI by default inside Docker. Should match local behavior: `!== "false"` (opt-out headless).
2. **522-line verbatim duplication of `cdp-page.ts`** across `base-nodes` and `sandbox-agent`. Extract to a shared package.

**Should fix:**
3. **`CDPClient = any`** and 8 other `any` usages — violates strict TS / no-`any` rule. Use proper types from `@types/chrome-remote-interface`.
4. **Document breaking changes** — 3 node types removed, tool ID format changed from single `"browser"` to 11 granular IDs. Saved workflows referencing old IDs will silently break.

**Nice to have:**
5. Unit tests for `CdpPage` core logic.
6. `setBrowserAdapter` injection point in `browser-tools-local.ts` for testability.

---

PR #3043, opened 2026-05-09
Skip unchanged scheduled nightlies

georgi:
## Summary
- add a scheduled-nightly guard that compares HEAD with the latest nightly tag
- skip the release matrix when the scheduled run has no changes to publish

## Validation
- workflow-only change, not run locally


georgi:
@copilot address comments

PR #3045, opened 2026-05-09
Publish release assets in a single job

georgi:
## Summary
- stop publishing GitHub Release assets from individual matrix jobs
- rely on uploaded build artifacts from each OS
- add one publish-release job that downloads all artifacts and publishes the release once

## Validation
- workflow-only change, not run locally


georgi:
@copilot address comments

PR #3046, opened 2026-05-09
Validate release assets before publishing

georgi:
## Summary
- add release asset validation for stable and nightly channels
- verify expected updater manifests and platform artifacts before publishing
- checkout release tooling in the publish job before validation

## Validation
- node scripts/validate-release-assets.mjs is a CI helper; full validation runs in release workflow


georgi:
@copilot address comments

PR #3048, opened 2026-05-09
Smoke test updater release assets

georgi:
## Summary
- add an updater manifest smoke-test script
- verify manifest version, path, sha512, and referenced artifact existence before release publishing
- document the updater smoke test in release docs

## Validation
- smoke script runs in the release workflow after asset validation


georgi:
@copilot address comments

PR #3115, opened 2026-05-13
[T-20260513-0029] PR 1 — Track A: ports + edges polish

georgi:
PR 1 of the node editor redesign — Track A polish for ports and edges.

- **A1 — Hover/select port label reveal.** Property labels start at `opacity: 0` and fade in (~120ms) when the parent node is hovered or selected. `.is-connected` rows keep labels lit so connection endpoints are always readable. Same affordance applied to `DynamicOutputItem` output labels. Selectors target both the Emotion `.base-node` wrapper and the outer `.react-flow__node` so pointer and keyboard selection both trigger the fade.
- **A2 — Edge endpoint chips.** `CustomEdge` renders small floating chips with source/target property names anchored just past each handle. Target chip auto-suppresses when the edge length is < 56 px (overlap) or the target node has > 6 inputs (OQ-2); a hover-only variant reappears on edge selection/hover. New chip data (`sourceHandleName`, `targetHandleName`, `targetInputCount`) flows from `useProcessedEdges`.
- **A3 — `TypedPortChip`.** New `web/src/components/node/TypedPortChip.tsx` renders the per-type icon (~14×14) next to input handles, tinted via `colorForType` + `color-mix`. Pure visual, `pointer-events: none`. Integrated into `PropertyField`.
- **A4 — Typed edge coloring.** Unselected edges keep the existing `colorForType(source.type)` stroke at 1.5 px. Selected edges get 2 px stroke + an outer `drop-shadow` glow tinted by the source type. CSS `stroke`/`stroke-width`/`filter` transitions animate the change. Removed the `var(--palette-text-primary) !important` override on `.react-flow__edge.selected .react-flow__edge-path` that was masking the typed color.
- **A5 — Collapsed state polish.** `--node-collapsed-height` set to 40 px; new `--node-collapsed-min-width: 140 px`. Collapse now clears inline `style.width` and `measured` so React Flow re-measures to fit the header content; expanded width is saved to `data.expandedWidthPx` and restored on expand. `collapsed.css` adds `width: max-content !important` + `min-width: var(--node-collapsed-min-width) !important` to the RF outer node. Ports remain functional on the header strip.

**Deviations from spec**

- The spec calls out `web/src/core/graph/` as the edge audit target; that directory doesn't exist in the repo. Edge rendering actually lives in `web/src/components/node_editor/CustomEdge.tsx` and `web/src/hooks/useProcessedEdges.ts`, which is where A4 was implemented.

---

Closes task **T-20260513-0029**: PR 1 — Track A: ports + edges polish.

### Acceptance criteria
- [x] A1: property labels hidden by default; visible while node hovered or selected; ~120ms transition
- [x] A1: connected properties keep labels visible regardless of hover
- [x] A2: edge endpoint chips render near tips with source/target property names
- [x] A2: nearby endpoint chips suppress target side to avoid overlap
- [x] A3: TypedPortChip component exists and renders per-handle type icon (~14×14)
- [x] A3: chip color comes from colorForType; handle itself unchanged
- [x] A4: all static edges use colorForType(source.type) for stroke
- [x] A4: selected edge is 2px with outer glow; unselected 1.5px
- [x] A5: collapsed nodes hide body, header ~40px, ports remain functional on header strip
- [x] A5: collapsed width matches header content width, not expanded width

🤖 Recovered from session #2 after a branch-name collision force-push.

### DIFF
--- a/web/src/components/node/NodeInputs.tsx
+++ b/web/src/components/node/NodeInputs.tsx
@@ -101,6 +101,7 @@ const NodeInput: React.FC<NodeInputProps> = memo(function NodeInput({
       tabIndex={tabIndex}
       isDynamicProperty={isDynamicProperty}
       data={data}
+      isConnected={isConnected}
     />
   );
 },
--- a/web/src/components/node/PropertyField.tsx
+++ b/web/src/components/node/PropertyField.tsx
@@ -15,6 +15,7 @@ import HandleTooltip from "../HandleTooltip";
 import { NodeData } from "../../stores/NodeData";
 import usePropertyValidationStore from "../../stores/PropertyValidationStore";
 import { Tooltip } from "../ui_primitives/Tooltip";
+import TypedPortChip from "./TypedPortChip";
 
 export type PropertyFieldProps = {
   id: string;
@@ -32,6 +33,8 @@ export type PropertyFieldProps = {
   isDynamicProperty?: boolean;
   hideActionIcons?: boolean;
   data: NodeData;
+  /** True when an edge is connected to this property's target handle. */
+  isConnected?: boolean;
   onValueChange?: (value: any) => void;
 };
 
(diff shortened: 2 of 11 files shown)


georgi:
@claude address comments

PR #3188, opened 2026-05-19
feat(correlation): finish PRs 0-4 of docs/correlation-design.md

georgi:
Replaces FIFO-based input joining with **lineage-identity joining** in the workflow kernel. The new path runs behind `NODETOOL_USE_CORRELATION=1`; the legacy `sync_mode` scheduler is unchanged when the flag is off. Implements PRs 0–4 of `docs/correlation-design.md`; PR 5 (default-on) is blocked on Python stdio bridge work (PR 4a/4b).

## The problem

Today, when a node receives data on multiple inputs the actor pairs them by FIFO arrival order (`zip_all`). If branch A and branch B both fan out from the same `ForEach` but A is slower, A's item *N* gets paired with B's item *M*. The workflow completes — but the result is wrong, silently. The other failure modes were `is_streaming_output` mixing three unrelated meanings (multi-item emission, chunked single item, fire-per-arrival), and nodes that declare `chunk` outputs but actually accumulate. Users were also being asked to pick `sync_mode: "on_any" | "zip_all"` per node when the right answer depended on lineage they couldn't see.

The fix is to stop pairing by arrival and start pairing by **identity**: every message carries a lineage that says which iteration items it came from, and the scheduler joins inputs whose lineage matches.

## The core ideas

**1. Every message carries a `CorrelationLineage`.** A lineage is a map from iteration-root id → token (an integer `index`). A `ForEach` named `fe` minting its 3rd item produces lineage `{ "fe:items": { index: 2 } }`. A downstream `forward` node copies this lineage onto its output; an `iteration` node adds a new root on top; an `aggregate` node trims the last root off.

```ts
interface MessageEnvelope {
  data: unknown;
  correlation_lineage: Readonly<Record<string, { index: number }>>;
  source_edge_id: string;
  // ...
}
```

**2. Input mode and output correlation are separate concerns.**

```ts
type InputMode = "buffered" | "stream" | "controlled";

type OutputKind =
  | "single"      // one logical output per invocation, inherits invocation lineage
  | "iteration"   // emits logical items, each adds a new root token
  | "chunk"       // emits chunks of one logical item, repeats_per_key
  | "forward"     // emits per source envelope, copies that source's lineage
  | "aggregate";  // collapses one root, emits at the parent scope

interface OutputCorrelation {
  kind: OutputKind;
  source: string;             // input handle name or "__execution__"
  group?: string;             // siblings share one minted token
  collapse?: "innermost";     // aggregate only
}
```

`ForEach.output` and `ForEach.index` both use `kind: "iteration"` with the same `group: "items"`, so a single minted token is attached to both sibling handles. `Reroute.output` is `forward` from its input. `Collect.output` is `aggregate(innermost)` over its `input_item` stream.

**3. Static analysis (`_analyzeCorrelation`) computes per-edge and per-output facts before any actor runs.** For each data edge: a `Scope` (ordered list of root ids, outermost parent first), a `repeats_per_key` flag, and the set of possible child roots reachable through it. For each node: an `invocationScope` = the largest non-empty input scope. The analyzer rejects the graph if two non-empty input scopes are incomparable by prefix:

> Inputs "left" and "right" come from independent iteration sources (scopes [fe1:items] and [fe2:items]). Add Zip or Cross to declare how they should join.

Nested iteration is fine because parent is a prefix of child; diamonds from one `ForEach` are fine because both legs share the same scope.

**4. The correlated buffered scheduler (`_runCorrelated`) joins by identity.** It classifies each connected input handle by its static scope:

- **max-scope** (handle scope equals `invocationScope`): buckets envelopes per canonical lineage key.
- **strict-prefix sticky** (scope is a shorter prefix): keeps one envelope per projected parent key; reused for every descendant.
- **empty-scope**: a single sticky entry reused for everything (constants, config).

A node fires for key K when every handle has a value at K (either buffered, or sticky for K's parent). The bucket key is a canonical `root=index,root=index,…` string in static-scope order, so lookups never depend on JS object property order.

**5. Iteration token minting.** The actor (not the node) mints `index` values, per `(root id, exact parent lineage)`. Counters never reset within a workflow run, so a child `ForEach` driven repeatedly under different parents continues numbering instead of restarting at 0. For `genProcess` frames where the node declares an `index` output handle, the actor overwrites it with the minted token.

**6. Control-plane signals carry "key dropped" and "scope closed."** Per-handle EOS is not enough — a filter can consume key K and emit nothing; an iteration can emit zero children for parent P. Two new signals fix this:

```ts
interface LineageDone {        // this edge will not produce a value at this key
  source_edge_id: string;
  output: string;
  lineage: CorrelationLineage;
}
interface LineageScopeClosed { // this edge will not produce more children of `closed_root` under `parent_lineage`
  source_edge_id: string;
  output: string;
  parent_lineage: CorrelationLineage;
  closed_root: string;
}
```

Stream filters call `outputs.drop(slot, envelope)` to emit `LineageDone`. At source EOS, the runner synthesizes `LineageScopeClosed` for every possible child root the analyzer recorded on each outgoing edge — so empty `ForEach` over `[]` still lets a downstream `Collect` finalize.

**7. Explicit joins for independent iteration sources.** `is_join_node` is a marker on `NodeDescriptor` that opts the node out of the incomparable-scope rejection and tells the analyzer to use the **longest common parent prefix** as the invocation scope. Two nodes ship with it:

- `ZipNode`: stream-mode, pairs by canonical lineage projection within the common parent prefix; bounded by `max_unmatched_pairs`.
- `CrossNode`: stream-mode, buffers both sides until EOS and emits the cartesian product up to `max_output_count`.

Both mint a new iteration root (`${nodeId}:zip` / `${nodeId}:cross`), so downstream sees one ordinary correlated stream instead of multiple independent ones.

**8. Bounded memory.** Inboxes carry `max_pending_keys` and `max_pending_messages_per_key` (default 10k each). Exceeding either is a terminal execution error naming the node, handle, and key — surfaces missing upstream closes instead of accumulating silently. Zip/Cross enforce their own caps before the generic limits trigger.

## What this PR changes

**`@nodetool-ai/protocol`**
- `InputMode`, `OutputKind`, `OutputCorrelation`, `CorrelationToken`, `CorrelationLineage`, `EMPTY_LINEAGE`, `LineageDone`, `LineageScopeClosed`.
- `NodeDescriptor` gains `input_mode`, `output_correlation`, `is_join_node`.
- `MessageEnvelope` gains `correlation_lineage` and `source_edge_id`.

**`@nodetool-ai/node-sdk`**
- `BaseNode.inputMode` / `outputCorrelation` / `isJoinNode` statics; propagated through `toDescriptor()` and `getNodeMetadata()`.
- Descriptor-level validation: every output needs `source`; `forward` source ≠ `__execution__` and ≠ multi-edge list; `aggregate` requires `collapse` and is rejected on buffered nodes; iteration groups must share one source.

**`@nodetool-ai/kernel`**
- New `correlation-analysis.ts`: scopes, `repeats_per_key`, possible child roots, close-barrier contributors, canonical `projectLineageKey`. Returns issues without throwing; the runner promotes them to `GraphValidationError` under the flag.
- `correlation-flag.ts`: `isCorrelationEnabled()` reads `NODETOOL_USE_CORRELATION` at observation time; `syntheticEdgeId` / `externalEdgeId` for deterministic source-edge ids when none is saved.
- `NodeInbox` gains canonical-key lineage signal recording, `signalEdgesForHandle`, and the pending-key limits.
- `NodeActor` gains `_runCorrelated` beside `_runBuffered`, iteration-token minting, per-slot lineage hints, `outputs.drop` wiring.
- `WorkflowRunner` runs `analyzeCorrelation` once per run under the flag, threads `NodeAnalysis` into each actor, routes per-slot lineage on data puts, dispatches drop signals as `LineageDone`, and synthesizes `LineageScopeClosed` on EOS.

**`@nodetool-ai/runtime`**
- `StreamingInputs.streamWithEnvelope` / `anyWithEnvelope` (§5).
- `StreamingOutputs.emit(slot, value, { lineage })`, `forward(slot, envelope, ...override)`, `drop(slot, envelope)`.
- `MessageEnvelopeLike` exported so nodes can type envelopes without depending on kernel internals.

**`@nodetool-ai/base-nodes`**
- 49 nodes across 17 files classified as `iteration` / `forward` / `chunk` / `single` / `aggregate`.
- New `ZipNode` and `CrossNode`.
- `Collect`, `Last`, `Count` already use `input_mode: "stream"` with `aggregate(innermost)`.

**`docs/correlation-design.md`** — adds a status matrix at the top.

## What's not in this PR

- Python stdio bridge: protocol version handshake, `correlation_lineage` / `source_edge_id` over msgpack, `LineageDone` / `LineageScopeClosed` envelopes. The bridge currently runs against the legacy scheduler only.
- Flipping `NODETOOL_USE_CORRELATION` to default-on. Gated on the Python bridge work.
- Removing `SyncMode`, `_runOnAny`, `_gatherZipAll`, `_initialStickyHandles`, and the deprecated `isStreamingInput`/`Output` shims. Scheduled for "one release after default-on".

## Test plan

- [x] `packages/kernel` — **453 tests pass** (44 files). New files: `correlation-analysis.test.ts`, `correlation-signals.test.ts`, `correlation-scheduler.test.ts`, `correlation-limits.test.ts`, `correlation-join.test.ts`. Existing `correlation-envelope.test.ts` / `correlation-runner.test.ts` still green.
- [x] `packages/base-nodes` — **2349 tests pass**.
- [x] All 34 workspace packages build clean (`npm run build:packages`).
- [x] **Diamond regression**: one branch reverses its output order; lineage-identity pairing still produces `A|A` and `B|B` (the FIFO bug this design exists to fix).
- [x] **Zip integration**: two seed nodes emit `A0/A1` and `B1/B0` (out of order); `ZipNode` pairs by lineage `index` and emits with a freshly minted `zip:zip` root.
- [x] **Incomparable sources**: a graph joining two independent `ForEach`es without `Zip`/`Cross` fails at graph load with the design's join-required error.
- [x] **Empty iteration closes**: `ForEach` over `[]` records `LineageScopeClosed` for the iteration root on its downstream edge.
- [x] **Flag-off path**: legacy scheduler behavior unchanged on a workflow the analyzer would reject.

## Commits

- `eb7bc02` PR 1 — input mode + output correlation metadata
- `2fd8ee8` PR 1 review fixes
- `7daef31` PR 2 — envelope propagation behind the flag
- `b92f809` PR 2 review fixes
- `8b11328` PR 3 step 1 — static `analyzeCorrelation` pass
- `d3a7140` PR 3 step 2 — `LineageDone` / `LineageScopeClosed` types and recording
- `8b62f1a` PR 3 step 3 — correlated buffered scheduler
- `03ad7c4` PR 3 step 4 — pending-key limits + EOS scope close synthesis
- `9c6b867` PR 4 — Zip/Cross + envelope-aware stream APIs
- `49b0edc` docs — status matrix in `docs/correlation-design.md`

https://claude.ai/code/session_01V4aypikj6izgMAnmgVwUbx


georgi:
Both fixes pushed in b743bba.

**Issue 1 — bucket key.** Zip now reads `inputs.invocationScope()` (the analyzer's common parent prefix) and `inputs.scopeFor(handle)` to compute each side's differing-root id. Buckets are keyed by `parentKey | differingIndex`, so a left `{a:items: 0}` and a right `{b:items: 0}` both land in `"|0"` and pair correctly. Two new `StreamingInputs` methods expose the static facts at runtime; `NodeInputs` takes the per-node `NodeAnalysis` from the actor.

**Issue 2 — sibling splitting.** Added `outputs.emitGroup(values, opts?)` — atomic frame emission that mints **one** token per iteration group and attaches it to every sibling handle in that group. `ZipNode` now calls `outputs.emitGroup({ left, right, index })`; `CrossNode` calls `outputs.emitGroup({ left, right })`. The actor's `_emitGroup` walks each slot's `output_correlation`, mints once per group root for the current parent key, and dispatches the whole frame in a single `_sendOutputs` call. Sequential `emit()` still mints per-call (one emit = one logical item) — that's intentional for non-grouped stream iteration.

`CrossNode.run` was also tightened to group by parent-prefix projection so the cartesian product stays within a shared parent scope; cross-parent products would conflate independent items.

**Test added** in `correlation-join.test.ts`: drives two seeds with incomparable roots emitting out of order (`A0/A1` and `B1/B0`), pipes Zip's `left` and `right` into separate sinks, asserts both pairs are emitted, the pair set is `{A0+B0, A1+B1}`, and at every emit index the left and right envelopes carry the **same** minted `zip:zip` token — exactly the identity guarantee the pre-fix code violated. All 453 kernel and 2349 base-nodes tests pass.

---

georgi:
Restructured the tests as a conformance suite in 415568a. The matrix from your review is now the directory layout — each rule from the design gets its own folder, each test asserts the algebra rule, not "feature X works."

```
tests/correlation/
  _harness.ts                  # runWorkflow + canonical executors
  _assertions.ts               # assertSameLineage, assertHasRoot, ...
  README.md                    # matrix + invariants
  propagation/                 # forward, single, iteration, chunk, siblings
  joining/                     # diamonds, sticky, parent-reuse, rejection
  aggregation/                 # collect/last/count, empty, nested
  drops-closes/                # drop→lineage_done, EOS→scope_closed
  explicit-joins/              # Zip identity, Zip siblings, Cross product
  limits/                      # finite caps
  flag-compat/                 # env-var, on/off, backward-compat
  descriptor-validation/       # source, aggregate-buffered, join hatch
  bridge/                      # Python stdio (todo, blocked on PR 4a/4b)
```

**Assertion vocabulary** — every test asserts both values (user-visible) and lineage envelopes (scheduler-visible). Helpers in `_assertions.ts`:

```ts
assertSameLineage(a, b);
assertHasRoot(env, "fe:items");
assertRootIndex(env, "fe:items", 2);
assertRootCollapsed(env, "fe:items");
assertJoinRejected(error);
assertSiblingOutputsShareToken({ left, right, index }, "zip:zip");
assertScopeClosed(inbox, edgeId, parentLineage, root);
assertLineageDone(inbox, edgeId, projectedKey);
```

**Levels** — kept the three unit files (`correlation-analysis`, `correlation-envelope`, `correlation-signals`) plus the node-sdk validation file as level A. Everything in `tests/correlation/` is level B (real `WorkflowRunner` + synthetic nodes via `_harness.ts`). Level C (JSON golden workflows under `fixtures/correlation/`) is documented as TODO in the README.

**Coverage delta vs. your matrix**: 37 new conformance tests across sections 1–9. Section 10 (Python bridge) is 6 `it.todo` entries documenting what needs to land in PR 4a/4b. The four old ad-hoc files (`correlation-{runner,scheduler,limits,join}.test.ts`) were deleted — they're superseded.

**Correctness fix surfaced by the suite**: the `aggregation/` tests caught that runtime lineage on `aggregate(innermost)` emits was still carrying the collapsed root even though the static analyzer dropped it from the output scope. The static and runtime views were out of sync. Fixed by adding `_maybeCollapseForSlot` on the actor's streaming-input sendFn — stream-mode aggregate emits now project the collapse at runtime too, matching the analyzer. This is exactly the kind of scheduler-vs-design drift the lineage assertions are meant to catch before users see wrong pairing.

474 kernel tests pass (was 453; +21 net after removing the four superseded files). 2349 base-nodes tests still pass. https://github.com/nodetool-ai/nodetool/pull/3188

---

georgi:
Pushed 6f7e420 addressing the review findings and the CI failure.

**Packages CI failure** — `correlation-validation.test.ts` "rejects references to handles not declared" expected 1 issue but got 2 after 2fd8ee8 added the "every declared output needs an entry" check. Updated the test fixture to include the declared output's entry so only the ghost-not-declared issue fires (matching the test's intent).

**Review #1 — `_maybeMintForSlot` ran with flag off**: Real bug. The streaming-input `sendFn`'s minting/collapse hooks and the dropFn were unconditional. Under `NODETOOL_USE_CORRELATION=0`, any node declaring `kind: iteration` (e.g. ForEach) would still have minted tokens attached. Wrapped the correlation-derived paths in `isCorrelationEnabled()`. `emitGroupFn` now falls back to plain sequential `_sendOutputs` (no minting) when the flag is off; `dropFn` is a no-op. Added a conformance test in `flag-compat/` that pins it: a ForEach under flag=0 must emit envelopes with `correlation_lineage: {}`.

**Review #2 — `Graph.loadFromDict` inconsistency**: Made `is_join_node` registry-authoritative (`descriptorDefaults.is_join_node` only, no `|| node.is_join_node` fallback), matching the policy for `input_mode` / `output_correlation`. Saved JSON is a cache; the resolver wins.

**Review #3 — `is_join_node` not propagated from `NodeMetadata`**: `createGraphNodeTypeResolver()` was missing the `is_join_node` spread in `descriptorDefaults`. Without it, Zip/Cross descriptors resolved from the registry would lose the join-node flag and the analyzer would reject their incomparable inputs — making them unusable. Added the conditional spread.

**Review #5 — `FilterNoneNode` is mislabelled**: Confirmed the reviewer's analysis — `isStreamingInput=true` + `input_mode: "buffered"` + only `process()` means the actor calls `process({})` once with empty inputs and the node never filters a stream. Pre-existing semantic bug across the codebase; four test files (`nodes.test.ts`, `coverage-phase3.test.ts`, `lib-data-document-code.test.ts`, `coverage-gap-fill.test.ts`) pin the current buggy shape. Added a `TODO(correlation)` block in the source naming the recommended fix (input_mode: "stream", run() with forward/drop) — but deferred the actual conversion to a follow-up so this PR doesn't grow a coordinated test refactor.

**Tests**: kernel 475 (+1 for flag-off-no-mint guarantee), node-sdk 213 (unchanged, assertion fixed), base-nodes 2349 (unchanged).

**CI status**: Other failures from the previous run — Web Test had only lint warnings + console errors (no actual test failures locally), Electron Test exited 137 (OOM, likely flaky). Will let CI re-run on the new push.

---

georgi:
Addressed the six new Copilot review comments in 9e0cef4.

**Applied:**

- **Zip/Cross `projectParent` collapse on missing parent token** (control.ts:1046 + 1192) — both now throw an `Error` naming the missing parent root. The reviewer is right that returning `""` was silently bucketing distinct parents into the same key, producing cross-parent pairs. Since `inputs.invocationScope()` is analyzer-provided, a missing token is always upstream correlation breakage and should fail loud.
- **`MessageEnvelopeLike.correlation_lineage` typed as mutable** (node-executor.ts:64) — replaced the inline `Record<string, { index: number }>` with the protocol's `CorrelationLineage` (`Readonly<...>`) and switched the `emit`/`forward` `opts.lineage` parameters to the same type. Behaviour unchanged; types now consistent with the wire format.

**Out of scope, addressed via reply:**

- **`NodeInbox.put` legacy `Record` fallback** (inbox.ts:53) — the `PutOptions`-only shape was a deliberate API simplification in 2fd8ee8 ("the ambiguous Record-vs-options heuristic is gone"); all four internal callers were updated in that same commit. No legacy callers remain.
- **`FilterNoneNode` stream-vs-buffered mismatch** (data.ts:1650) — the reviewer is repeating their earlier flag. It's a pre-existing semantic bug across four test files (`nodes.test.ts`, `coverage-phase3.test.ts`, `lib-data-document-code.test.ts`, `coverage-gap-fill.test.ts`) that pin the current shape. Flagged in 6f7e420 with a `TODO(correlation)` block in the source naming the recommended fix (input_mode: "stream" + run() with forward/drop); deferred to a follow-up so this PR doesn't grow a coordinated test refactor.
- **EOS `lineage_scope_closed` per parent key** (runner.ts:1110) — real architectural gap. Current synthesis emits `parent_lineage: {}`/empty parent scope, which only finalizes the root scope; nested iteration needs per-parent-key closes. The fix requires observing arriving lineages on inboxes (or tracking per-parent close barriers in the actor) — non-trivial enough to warrant a separate PR with dedicated conformance coverage rather than tacking it on here.

475 kernel + 213 node-sdk + 2349 base-nodes tests still pass.

---

georgi:
**Review: Request Changes**

The architecture is sound and well-documented, but there are several correctness bugs that need fixing before merge:

1. **`ZipNode.projectParent()` returns `""` on missing tokens** — If a parent-scope root token is absent, the function silently falls back to `""`, which would collapse distinct parent buckets into one. This should throw or return a unique sentinel to avoid silent cross-parent pairings.

2. **`CrossNode.projectParent()` has the same issue** — same fix needed.

3. **EOS `lineage_scope_closed` synthesis uses empty parent lineage** — In `runner.ts`, the synthesized signal always emits `parent_lineage: {}`. For nested iteration, downstream aggregators need a per-parent close signal. This is a correctness gap for nested `ForEach` chains.

4. **`FilterNoneNode` is broken under the correlation path** — It's `isStreamingInput=true` + `inputMode="buffered"` + only has `process()`. The actor will call `process({})` once with empty inputs, so the node never filters anything under the new scheduler.

5. **Flag isolation leak in streaming `sendFn`** — When a streaming node calls `outputs.emit(slot, val, { lineage: x })` under the flag-off path, lineage is still attached to the `put` call because `hints` construction with `callerSuppliedLineage` happens before the `isCorrelationEnabled()` guard. This violates the stated flag-off isolation guarantee.

6. **CI is red** (Packages Test, Web Test) — needs to be green before merge.

The test suite is substantial and the design doc is excellent. Once these bugs are fixed, this should be good to go.

---

PR #3285, opened 2026-05-23
Reorganize panels: group Left, slim Right to Inspector, expand Bottom

georgi:
PanelLeft: 12 icons → 5 (Search, Workflows, History, Assets, Nodes). All
node tile-grids (I/O, Tools, Image/Video/Audio/3D Models, Agents, Control
Flow) now live as sub-tabs under a single "Nodes" view.

PanelRight: only hosts the Inspector. The toolbar is gone; a single "I"
button toggles. The frontend-tool-runtime sync stays here as an invisible
sibling component.

PanelBottom: now hosts Logs, Jobs, Sandboxes, Workflow Settings, Versions,
Workflow Assets, Workspace, Assistant, Agent, Trace — grouped into Run /
Workflow / Chat / Debug. The tab rail (text + icon) is always visible as
a thin 36px bar so users can see what's available; clicking expands the
selected view.

Consumers redirected: useNodeEditorShortcuts and useFloatingToolbarActions
now drive workflow-settings via the bottom panel; CommandMenu's settings
and chat entries route to the bottom panel; the "o" assistant hotkey opens
the bottom panel chat group instead of the right panel.

Stores get persistence migrations so legacy "workflowGrid" / node-category
ids on the left, and any non-inspector view on the right, fold cleanly
into the new schema.


georgi:
@copilot screenshot the editor with different combination of opened panels. add screens to PR description

PR #3288, opened 2026-05-23
Add direct-gen (text-to-image / image-to-image) to the timeline editor

georgi:
Mirrors the sketch editor's direct-gen layer pattern on every timeline
clip so generation is intuitive and quick to iterate on, without forcing
users to wire up a workflow first.

Type model
- TimelineClip gains an optional `bindingKind` discriminator and a set of
  direct-gen fields (prompt, provider, model, sourceClipId, strength,
  numInferenceSteps, seed, width, height, negativePrompt). Legacy clips
  load as `"workflow"` so existing sequences keep working.
- The zod schema in `protocol` is widened to match, so autosave
  round-trips the new fields.

Store
- `addDirectGenClip` creates a draft prompt-bound clip purely in the
  client (autosave persists it — no new tRPC endpoint required).
- `setClipPrompt` / `setClipDirectGenModel` / `patchClipBinding` update
  the binding and auto-stale any clip that already has a render.
- `regenerateAsCopy` drops a fresh sibling immediately to the right
  preserving the binding but resetting render state, so users can roll a
  new take without losing the existing render.

Hook
- `useTimelineDirectGenJob` fires the `generate_media` RPC and writes
  the resulting asset + ClipVersion onto the clip. Locked clips record
  the version without replacing `currentAssetId`.
- `useGenerateClip` dispatches to the direct-gen path when
  `bindingKind` is direct, otherwise the existing workflow-runner path.

UX surfaces
- TopBar prompt: always-visible prompt input + model picker. Enter drops
  a clip at the playhead on the first compatible track and starts
  generation.
- AddClipMenu: prompt-first section at the top of the popover; workflow
  picker stays available below.
- Inspector: direct-gen clips render a new DirectGenClipPanel with
  prompt + model + image-to-image source picker + Generate/Cancel.
  Workflow-bound clips keep the existing inputs panel.
- ClipActions gains a "Regenerate as new clip" action surfaced on all
  generated clips.

Tests
- Six new TimelineStore unit tests cover addDirectGenClip,
  setClipPrompt (stale-on-change semantics) and regenerateAsCopy
  (binding preserved, render state cleared, original untouched).
- TimelineEditor.test mocks TopBarPrompt so the editor-shell tests
  don't need a QueryClientProvider for the new model picker.

https://claude.ai/code/session_01LJCzJLZqdhdTS4tRcNNqcG


georgi:
## Review

Clean scope (1,912 additions, 18 files, 6 new tests). The direct-gen pattern for the timeline editor looks reasonable.

**Action needed:** This branch is ~15 commits behind `main` and mergeable state is unknown. Please rebase onto current `main` and verify CI is green.

---

PR #3289, opened 2026-05-23
feat(gpu): phase 3 batches 3-4 + GPU-backed image nodes under lib.image.*

georgi:
Finishes the shader-pool plan's phase 3 (batches 3 + 4) on the `packages/gpu` side, then surfaces every relevant module to the workflow runtime as `lib.image.*` nodes. 31 files, +4048 lines, 26 new user-facing nodes, 50+ new tests, all green against Dawn + SwiftShader.

### 1. GPU shader pool — phase 3 completion (`packages/gpu/`)

**Batch 3 — transforms.** Nine geometric warps, all fragment-kind, single texture input (displace adds a `displacement` slot), uniform-resolution:

- `transform.pad`, `transform.tile`, `transform.resize` (nearest/bilinear/bicubic), `transform.rotate90`, `transform.affine`, `transform.cornerPin` (inverse 3×3 homography with H₂₂ fixed at 1), `transform.polarRemap`, `transform.displace`, `transform.spherize`
- Output sizing: `pad` and `rotate90` declare derived dimensions; `resize` and `affine` take host-specified targets; the rest are same-as-source

**Batch 4 — mixer + channel ops.**

- `mixer.colorOverlay` — constant-colour tint masked by source alpha
- `mixer.outline` — 8-tap ring edge fill
- `mixer.dropShadow` — four-pass *recipe*: `mask.fromImage → blurH → blurV → mixer.shadowCompose` (the `shadowCompose` helper lives in the same batch)
- `color.channelSplit` / `color.channelShuffle` / `color.channelMerge` under `color/`, per the plan's channel-ops namespace call-out

All modules registered in `ALL_SHADERS` / `ALL_RECIPES` / `ALL_CATALOG`, re-exported through `pool.ts`. `batch3Transforms.test.ts` + `batch4Modules.test.ts` cover category / kind / IO / defaults / uniform-resolution.

`mixer.composite` continues to wrap `WebGPULayerCompositor` and stays deferred per the original Batch 4 note.

### 2. Workflow runtime exposure (`packages/base-nodes/`)

Every shader-pool module that does something a user would actually pick from a palette now has a NodeTool workflow node — but **under existing user-facing namespaces, not a `shader.*` namespace**. Shaders are an implementation detail; users see "Drop Shadow", not "shader.mixer.drop_shadow".

| Namespace | Nodes |
|---|---|
| `lib.image.effects.*` | `ColorOverlay`, `Outline`, `DropShadow`, `Glow`, `Add` |
| `lib.image.keyer.*` | `ChromaKey`, `LumaKey` |
| `lib.image.mask.*` | `Apply`, `FromImage`, `Invert` |
| `lib.image.channel.*` | `Shuffle`, `Merge` |
| `lib.image.warp.*` | `Offset`, `Pad`, `Tile`, `Affine`, `CornerPin`, `PolarRemap`, `Displace`, `Spherize` |
| `lib.image.draw.*` | `LinearGradient`, `RadialGradient`, `AngularGradient`, `DiamondGradient`, `Checkerboard` |
| `lib.image.filter.*` | `Threshold`, `Pixelate` |

**Dedup — no duplicate nodeTypes added** when an existing nodeType already covers the functionality:

- `sources.solid` → `lib.image.draw.Background` (existing)
- `transform.{mirror, rotate90, crop, resize}` → `nodetool.image.{RotateAndFlip, Crop, Resize/Scale}` (existing)
- `color.channelSplit` → `nodetool.image.Channels` (existing)
- `color.{grade, invert, hsb, exposure, posterize, brightnessContrast}`, `filters.{blur, sharpen, vignette}` → existing `lib.image.{filter, enhance, color_grading}.*`

**Existing-node body swaps to GPU** are a deliberate follow-up, not in this PR: the existing sharp-based nodes emit PNG, the GPU helper emits raw RGBA, and the existing `lib-image-processing.test.ts` decodes via `Buffer.from(data, "base64") → sharp(buf).raw()`. A clean swap needs either a PNG re-encode at the boundary (loses chain-friendly raw RGBA) or an updated test contract.

### 3. `lib-shader-utils.ts` — the helper that backs every new node

Shared infrastructure so each node file stays declarative:

- `runShaderNode(module, params, mainInput, opts, ctx)` — decode input(s) to RGBA, run the shader, return an `image/x-raw-rgba` ImageRef
- `runRecipeNode(recipe, params, ...)` — same for multi-pass recipes (drop shadow, glow)
- `decodeRgba` import from `image.ts` for size derivation
- `colorProp` / `floatProp` / `intProp` / `extraImageProp` / `IMAGE_PROP` — typed prop factories so the node files don't repeat `PropOptions` boilerplate
- `colorValueToVec4` — accepts hex `#rrggbb[aa]`, `{ value: "#..." }`, or `[r, g, b, a]` arrays
- Cached Dawn context per process (matches `WebGPULayerCompositor`'s singleton pattern)
- Premultiply / un-premultiply on the boundary so blend math matches WGSL expectations

### 4. Tests

- `packages/gpu/tests/batch3Transforms.test.ts` + `batch4Modules.test.ts` — module-level shape/contract tests (no GPU needed)
- `packages/base-nodes/tests/lib-image-gpu-smoke.test.ts` — end-to-end smoke for all 26 new nodes against Dawn + SwiftShader (`27` cases, `.skipIf(!hasGpu)` so CI without a GPU skips cleanly)
- The SwiftShader ICD setup is opt-in and machine-aware: only registers a CPU Vulkan driver if no system Vulkan driver is present, never overrides an operator-set `VK_ICD_FILENAMES`/`VK_DRIVER_FILES`, never touches production code paths
- **Total**: `2398` base-nodes tests pass, `328+` gpu tests pass, `tsc --noEmit` clean across both packages

## How it runs in CI without a GPU

The test harness reaches Dawn → Dawn uses its Vulkan backend → the Vulkan loader finds the SwiftShader ICD that ships in `electron/dist/` (Chromium uses it for the same purpose) → SwiftShader's CPU Vulkan implementation runs the WGSL. Adapter info confirms: `vendor: google, architecture: swiftshader, description: SwiftShader driver 5.0.0`. GPU-equipped hosts use real silicon automatically (the loader picks the hardware driver if one is registered).

This is not Dawn's `forceFallbackAdapter` — we're just letting the Vulkan loader pick SwiftShader because it's the only available ICD on the test machine.

## Out of scope (intentionally)

- Swapping existing sharp-based bodies (`lib.image.filter.Invert`, `.ConvertToGrayscale`, `lib.image.enhance.{Brightness, Contrast, Sharpness, UnsharpMask}`, `lib.image.color_grading.{Vignette, Exposure}`, `nodetool.image.{Blur, Channels}`) to GPU. Doable but changes output `mimeType` (PNG → raw RGBA) and touches every existing image-processing test. Separate PR.
- `mixer.composite` GPU module — still wraps `WebGPULayerCompositor`, per the plan.

## Test plan

- [x] `cd packages/gpu && npm test` — 328 tests pass
- [x] `cd packages/base-nodes && npx tsc --noEmit` — clean
- [x] `cd packages/base-nodes && npm test` — 2398 pass, 39 skipped (existing skip set, no regressions)
- [x] `cd packages/base-nodes && npx vitest run tests/lib-image-gpu-smoke.test.ts` — 27/27 against SwiftShader
- [ ] Manual UI smoke: spawn one of the new nodes in the web editor, verify it appears under the right namespace and runs end-to-end against the live server

https://claude.ai/code/session_019vm8cru9KLgyQf5EYUoRzc


georgi:
## Review

Large, well-documented PR (31 files, +4,048 lines, 26 new nodes, 50+ tests). Scope is reasonable for the GPU package.

**Blocker:** Mergeable state is "unstable" — CI needs to be green before merge. Please investigate and fix the CI failures, then this should be good to go.

---

georgi:
Addressed Copilot review feedback in `b69b4fd`.

**Active findings — fixed in this commit:**

| # | File | Fix |
|---|---|---|
| 5 | `lib-shader-utils.ts` runShaderNode | Guard *every* required (non-optional) declared input — not just `source`. Missing `over` / `mask` / `alpha` / `displacement` now returns an empty `ImageRef` instead of throwing inside `executor.encode`. Same guard added to `runRecipeNode`. |
| 6 | `lib-shader-utils.ts` executor cache | `Executor`, `RecipeRunner`, and the default `ShaderRegistry` are now cached at module scope so the internal sampler/pipeline WeakMaps survive across node invocations. |
| 7 | `lib-shader-utils.ts` `colorValueToVec4` | Now accepts the three shapes the docs promised: `{ value: "#rrggbb[aa]" }`, bare hex strings (`"#rrggbb"` / `"rrggbb"`), and `[r, g, b, a]` float arrays. NaN bytes fall back instead of producing garbage. |
| 8 | `lib-image-warp.ts` PadNode | Shader contract for `color` is premultiplied (WGSL returns `p.color` directly), picker emits straight alpha — now premultiplied on the boundary via new `premultiplyVec4` helper. **Also applied to `lib-image-generators.ts`** (linear/radial/angular/diamond gradient, checkerboard) — same root cause, same fix. Copilot didn't flag those individually. |
| 9 | `batch4Modules.test.ts:106` | Reworded "α=0.5" → "threshold=0.5". |
| 10 | `mixer/outline/v1/module.ts` | Short-circuit `widthPx <= 0` to return source unchanged, before the threshold branch can hard-cut soft edges. |

**Outdated findings — already fixed in earlier commit `1f707f0`:**

- `transform/rotate90` header comment (Copilot's complaint is stale — `1f707f0` rewrote that header)
- `color/channelShuffle` non-premul output (fixed in `1f707f0`)
- `mixer/colorOverlay` alpha interpolation (fixed in `1f707f0` — WGSL now does explicit un-premul / mix straight / re-premul against `src.a`)

**Spherize bounds check (#2):** comment was filed against line 66 but the file already has the explicit out-of-bounds → transparent-black check at lines 71-73 (also from `1f707f0`). No code change needed.

Verification: `2398` base-nodes tests + `328` gpu tests pass; `tsc --noEmit` clean across both packages; `27/27` GPU smoke tests against SwiftShader.

https://claude.ai/code/session_019vm8cru9KLgyQf5EYUoRzc

---

PR #3386, opened 2026-05-28
Align chat UX with target design

georgi:
## Summary
Frontend-only pass to bring the full-page chat in line with the target mockup. Everything is derived from data already on chat messages — no schema, tRPC, or backend changes.

- **Tool calls** — structured search-result cards (numbered title / snippet / source domain) via a small shape-driven renderer registry, with the existing pretty-JSON as fallback; plus best-effort execution timing on each card (derived from message timestamps).
- **Conversation header** — a derived `model · provider · runtime · spend · last-run` strip, gated to the full-page chat (GlobalChat) so embedded chats are unaffected.
- **Sidebar** — conversation count, ⌘K hint in search, full-width "New conversation" button, per-thread timestamps and per-group dates.
- **Messages** — persistent per-turn header with a role icon (person glyph for you, node/hub glyph brand-tinted for NODETOOL) and a left-aligned, no-bubble layout for both roles; gated to the full-page chat.
- **Top bar** — route-aware "Nodetool / {mode}" brand label and inline workspace path in the compact selector.
- **Composer** — Attach button + hidden file input, "↵ to send · ⇧↵ newline" hint, and an "@ a node…" placeholder.

### Deliberately out of scope
- Conversation **type** (filter pills / colored dots / "Image" badge) and **pinning** — would need persistence/derivation beyond presentation.
- **Branch / Rerun / Send to canvas** actions — real cross-stack features.
- Composer **Tools** chip and **microphone**.
- Top-bar connection **status dot** — chat-specific, and the header is global.

## Test plan
- [x] `npm run typecheck` (all packages) — passes
- [x] `npm run lint` on changed files — clean
- [x] Unit tests pass: `formatDuration`, `normalizeSearchResults`, `deriveThreadStats`, plus existing ChatView / ChatThreadView suites
- [ ] Manual: verify rendering in the running app (not yet browser-verified) — tool-result cards on a search conversation, header strip, sidebar timestamps/dates, message role icons, composer attach/hint


georgi:
@claude address comments

georgi:
@copilot The failure is a real test regression in `formatDuration`, not a workflow issue. The job that failed is the `quality` gate from [`.github/workflows/test.yml`](https://github.com/nodetool-ai/nodetool/blob/3da6a13b16473bb4d05bb8dc7c27bb1cfef0d179/.github/workflows/test.yml), and the only failing test is:

- `web/src/utils/__tests__/formatUtils.test.ts:95`
  - Source: https://github.com/nodetool-ai/nodetool/blob/3da6a13b16473bb4d05bb8dc7c27bb1cfef0d179/web/src/utils/__tests__/formatUtils.test.ts
  - Failing assertion:
    - `expect(formatDuration(59999)).toBe('1m 00s')`

Root cause:
- In `web/src/utils/formatUtils.ts`, `formatDuration` only switches to minute formatting when `totalSeconds >= 60`:
  - Source: https://github.com/nodetool-ai/nodetool/blob/3da6a13b16473bb4d05bb8dc7c27bb1cfef0d179/web/src/utils/formatUtils.ts
- Current logic:
  ```ts
  const totalSeconds = ms / 1000;
  if (totalSeconds < 60) {
    return `${parseFloat(totalSeconds.toFixed(1))}s`;
  }
  const totalWholeSeconds = Math.round(totalSeconds);
  ```
- For `59999`, `totalSeconds` is `59.999`, so it enters the `< 60` branch and returns `"60s"` after rounding to one decimal, but the test expects minute-boundary rounding to produce `"1m 00s"`.

### Solution
Base the branch decision on the rounded whole seconds, not the raw fractional seconds.

Suggested fix in `web/src/utils/formatUtils.ts`:

```ts
export function formatDuration(ms: number): string | null {
  if (!Number.isFinite(ms) || ms < 0) {
    return null;
  }

  if (ms < 1000) {
    return `${Math.round(ms)}ms`;
  }

  const totalSeconds = ms / 1000;
  const totalWholeSeconds = Math.round(totalSeconds);

  if (totalWholeSeconds < 60) {
    return `${parseFloat(totalSeconds.toFixed(1))}s`;
  }

  const minutes = Math.floor(totalWholeSeconds / 60);
  const seconds = totalWholeSeconds % 60;
  return `${minutes}m ${String(seconds).padStart(2, "0")}s`;
}
```

Why this works:
- `59999` → `totalWholeSeconds = 60` → formats as `1m 00s`
- `119999` → `totalWholeSeconds = 120` → formats as `2m 00s`
- normal sub-minute values like `59500` still return `59.5s`

If you want to make the behavior explicit, the key change is effectively this:
```ts
if (Math.round(ms / 1000) < 60) {
  return `${parseFloat((ms / 1000).toFixed(1))}s`;
}
```

This is the minimal code change that matches the failing test and preserves the existing formatting behavior elsewhere.

georgi (commented):
**Review: One fix needed, then ready to merge**

Good UX alignment — proper primitives usage in new files, correct `unknown` typing, adequate test coverage. CI passes.

**Must fix:**

1. **Raw MUI `Collapse` import in `MessageView.tsx`** — CLAUDE.md rule: "When touching any file, migrate raw MUI usage to primitives."

**Non-blocking concerns:**

2. Not browser-verified — the DOM restructuring in MessageView and always-visible file-preview row could cause visual regressions. Recommend manual check.

Fix the Collapse import and this is good to go.

---

PR #3391, opened 2026-05-28
feat(execution): queue concurrent runs and warn before large runs

georgi:
## Why

Being able to fire off multiple runs without waiting is great — but it has two sharp edges:

1. **One click can fan out a lot of work.** "Run Workflow" executes *every* non-bypassed node at once, so a large graph can launch dozens of provider/model calls from a single click.
2. **Nothing throttles concurrent runs.** Each workflow has its own runner (`getWorkflowRunnerStore(workflowId)`), so different workflows run concurrently, and the backend (`activeJobs`) accepted every `run_job` immediately — no cap, no queue. Easy to overload a provider/API.

This PR keeps the "run freely" behavior but adds two guards (per the chosen approach: **backend queue** + **heavy-node threshold warning**).

## Backend — per-client concurrency cap + FIFO queue

The runner is created per `/ws` connection and the web app uses one shared connection, so `activeJobs` is effectively per client session — the right scope for a cap.

- **`packages/websocket/src/job-queue.ts`** — new `JobConcurrencyQueue` (FIFO, 1-based position tracking, cancel-by-id).
- **`unified-websocket-runner.ts`** — `runJob` now gates on `activeJobs.size >= MAX_CONCURRENT_JOBS` (default **4**). Excess runs are queued and emit a `queued` `job_update` with `queue_position`. A queued run starts automatically when any active run finishes — `drainQueue()` is wired into all three completion paths (regular, ComfyUI, chat). `cancelJob` removes still-queued runs and rebroadcasts positions.
- **`settings-registry.ts`** — registers `MAX_CONCURRENT_JOBS` (configurable via Settings / env).
- **`protocol/messages.ts`** — adds `JobUpdate.queue_position`.

## Frontend — heavy-node run warning + queued UX

- **`utils/heavyNodes.ts`** — classifies provider/model/API nodes from metadata (`required_settings` / `recommended_models` / `auto_save_asset`); constants, text ops, routing, etc. are ignored, so the warning only counts nodes that could actually overload an API.
- **`RunWarningStore.ts`** (session-scoped, not persisted) + **`RunWarningDialog.tsx`** — "Run anyway / Cancel" with a **"Don't ask again for this session"** checkbox; mounted once at app root.
- **`useFloatingToolbarActions.ts`** — `handleRun` shows the warning only when the heavy-node count exceeds the threshold and the warning isn't suppressed; otherwise runs immediately.
- **`SettingsStore.ts` + `SettingsMenu.tsx`** — "Warn Before Large Runs" toggle + configurable threshold (default **5**) under a new **Settings → Execution** section.
- **`WorkflowRunner.ts` / `workflowUpdates.ts` / `StatusMessage` / `FloatingToolBar`** — surface `queuePosition` as **"Queued (#N)"**. Reuses the existing `running` state (queued runs are still "busy"), so there's no churn across the ~80 `state === "running"` checks.

## Config & defaults

| Setting | Default | Where |
| --- | --- | --- |
| `MAX_CONCURRENT_JOBS` | 4 | Backend setting / env |
| `confirmLargeRun` | true | Settings → Execution |
| `largeRunThreshold` (heavy nodes) | 5 | Settings → Execution |

## Tests

Unit tests added for `JobConcurrencyQueue`, `heavyNodes` (`isHeavyNode`/`countHeavyNodes`), and `RunWarningStore`.

> Note: this branch was developed in a fresh container where the full workspace `npm install` had not finished, so `typecheck`/`lint`/`test` were not run locally — relying on CI for the full gates. Happy to push fixes for anything CI flags.

https://claude.ai/code/session_01HhkdFpduc7dWMgdSwKHtUZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HhkdFpduc7dWMgdSwKHtUZ)_


georgi (commented):
**Review: Needs fixes before merge**

Good feature — backend concurrency cap + FIFO queue and heavy-node warning are well-designed. CI passes. Three items need fixing:

**Must fix:**

1. **Raw MUI import in `QueuePanel.tsx`** — `import { List, ListItem } from "@mui/material"` violates the mandatory ui_primitives rule.

2. **`ai_docs/concurrent-node-runs-and-job-queueing.md` should not be committed** — project rules say no documentation files unless explicitly requested. Good analysis but belongs in PR description, not the repo.

3. **`drainQueue()` fire-and-forget can silently lose jobs** — if `startJob` throws after `dequeue()`, the dequeued job is lost. Re-enqueue on failure or propagate.

**Should fix:**

4. `RunningJobsStore` is wired but never read — dead code. Either wire to UI or remove.
5. `heavyNodes.test.ts` is co-located — convention is `__tests__/` directories.
6. `getMaxConcurrentJobs()` hits settings on every `runJob`/`drainQueue` — consider caching.

---

PR #3420, opened 2026-05-30
feat(prompt): expand asset & folder mentions into typed media inputs

georgi:
## Summary

When a user `@`-mentions an asset in a **Prompt** node, it lands in the prompt string as an `asset://<id>.<ext>` token. Text-generation providers already expand those tokens into multimodal image/audio blocks — but image/video **generation** nodes ignored them, leaving a dangling URI in the prompt and forcing the user to also wire the image into a separate input.

This PR generalizes that expansion into a single shared mechanism and applies it everywhere a node has a prompt **and** typed media inputs: native image-to-image / image-to-video nodes, and the FAL / KIE / Replicate node factories (~1,500 generated nodes). Mentions now flow into the node's media inputs automatically, and the prompt text is rewritten to reference the slot each image landed in. Folder mentions expand recursively to their member assets.

## The mechanism — `mapPromptAssetsToInputs` (`packages/runtime/src/prompt-asset-refs.ts`)

Given a node's text inputs and its typed media inputs, it:

1. **Expands folder mentions first.** A bare `asset://<id>` (no extension — exactly how the composer encodes a folder) is resolved via the new `ProcessingContext.listFolderAssets`; a folder is replaced inline by the `asset://<child>.<ext>` tokens of its members of an accepted kind, recursively. Non-folder bare tokens are left alone.
2. **Maps as many as possible.** Each empty image/audio input is filled in declaration order; `list[...]` inputs absorb the remainder. Already-wired inputs win and are skipped.
3. **Resolves to bytes.** Each placed mention is dereferenced once via `loadMediaRefBytes` and injected with `data` (base64), so every provider's existing `data` path uploads it — no provider needs to learn the `asset://` scheme.
4. **Relabels the prompt.** A placed mention is rewritten to the input label it now occupies — `image_url`, `reference_image_urls[0]` — so the model can tell which image is in which slot. Mentions that couldn't be placed (wired slot, or capacity exhausted) are dropped rather than left as a dangling URI.

The text-generation path (`expandAssetReferences` in `llm-nodes/agents.ts`) was refactored onto the same shared parser, removing a duplicate regex + MIME table (output is byte-identical; its tests are unchanged).

### Example

Prompt: `"blend asset://a.png with asset://b.png"` on a node exposing `image_url` + `tail_image_url`
→ `image_url` = a, `tail_image_url` = b, prompt becomes `"blend image_url with tail_image_url"`.

Folder of 5 images on a node with `image_url` + a `reference_image_urls` list
→ first image fills `image_url`, the rest fill the list; the mention becomes `image_url reference_image_urls[0] reference_image_urls[1] …`.

## Call sites wired

| Area | Node(s) |
|---|---|
| `image-nodes` | `ImageToImageNode` (refactored onto the shared helper) |
| `video-nodes` | `ImageToVideoNode` |
| `fal-nodes` | factory `buildArgs` → all generated FAL nodes |
| `replicate-nodes` | factory `buildArgs` → all generated Replicate nodes |
| `kie-nodes` | factory `buildParams` → all generated KIE nodes (assets come from `spec.uploads`, applied in both the scalar and upload loops) |

The factories pass the API param name (`field.apiParamName` / `upload.paramName`) as the prompt **label** while keeping the node field name as the override key.

## Folder support (`packages/runtime` + `packages/websocket`)

- New optional model interface `listFolderAssets({ userId, folderId })` and `ProcessingContext.listFolderAssets(folderId)` — returns the flattened non-folder descendants, or `null` when the id isn't a folder (so non-folder bare tokens aren't mistakenly stripped).
- Server implementation walks `Asset.getChildren` recursively (cycle-guarded, sorted by name, 1000/level cap).

## Prompt node UI (`web`)

- The Prompt node's **"Variables"** label now renders only when the prompt text actually references a variable (`{{ … }}`), tracked live as you type, instead of always showing.

## Protocol (pre-existing in the working tree, included here as it's on-theme)

- Asset search now accepts an empty query (was `min(2)`), so the `@`-mention typeahead can list assets — including folders — before the user narrows it. Has its own test.

## Why bytes, not just the URI

All three providers' resolvers (`assetToFalUrl`, `assetToUrl`, `resolveUploadBytes`) handle a ref carrying `data`, but **none** handle the `asset://` scheme. Resolving once in the helper and handing every provider a `data`-bearing ref keeps each provider untouched.

## Tests

- **runtime** — 22 tests for the parser, mapper (kind routing, list absorption, wired-precedence, label rewriting), and folder expansion (multi-image, accepted-kind filtering, non-folder passthrough, empty-folder strip).
- **image / video / fal / replicate / kie** — integration tests per family proving "mention → resolved bytes → provider input + relabeled prompt", wired-input precedence, and FAL multi-frame "map as many as possible".
- **llm-nodes** — existing `expandAssetReferences` suite unchanged (refactor is behavior-preserving).
- **web** — Prompt label visibility tests.

Suites: runtime 987, image 143, video 76, fal 119, replicate 54, kie 43, llm 423, protocol 4. All green; lint clean; all touched packages build.

## Known limits / follow-ups

- **List inputs are uncapped**: a folder of 100 images feeding a `list[image]` resolves and inlines all 100. Single-slot inputs only resolve what they place. A per-folder cap would be a small addition.
- **Provider-specific Gemini/OpenAI image+video nodes** are excluded — their `process()` has no `ProcessingContext`, so they can't resolve `asset://` without extra plumbing.
- **Dynamic raw FAL nodes** (`FalRawNode`/`FalDynamicNode`) bypass the factory `buildArgs` and aren't covered.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42131,7 +42131,8 @@
       "version": "0.7.0-rc.23",
       "license": "MIT",
       "dependencies": {
-        "@nodetool-ai/node-sdk": "*"
+        "@nodetool-ai/node-sdk": "*",
+        "@nodetool-ai/runtime": "*"
       },
       "devDependencies": {
         "@types/node": "^22.0.0",
@@ -42238,6 +42239,7 @@
       "license": "MIT",
       "dependencies": {
         "@nodetool-ai/node-sdk": "*",
+        "@nodetool-ai/runtime": "*",
         "replicate": "*"
       },
       "devDependencies": {
--- a/packages/fal-nodes/src/fal-factory.ts
+++ b/packages/fal-nodes/src/fal-factory.ts
@@ -13,7 +13,12 @@ import {
   registerDeclaredProperty
 } from "@nodetool-ai/node-sdk";
 import type { NodeClass, PropOptions } from "@nodetool-ai/node-sdk";
-import type { ProcessingContext } from "@nodetool-ai/runtime";
+import type {
+  ProcessingContext,
+  PromptAssetTextField,
+  PromptAssetInputField
+} from "@nodetool-ai/runtime";
+import { mapPromptAssetsToInputs } from "@nodetool-ai/runtime";
 import {
   getFalApiKey,
   falSubmit,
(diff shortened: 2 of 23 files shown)


pauliusuza:
great!

PR #3423, opened 2026-05-30
Canvas media composer: generate assets, run, and chat from the workflow

georgi:
## Summary

Turns the canvas floating toolbar into a **media composer**. From the bottom of the workflow editor you can:

- **Generate media** (image / video / audio / image-edit / image-to-video / speech) and have the results **drop onto the canvas as constant nodes** — no trip to the chat panel.
- **Run / Stop the workflow** from an always-visible button, with the rest of the canvas controls tucked into a single normal dropdown menu.
- **See the conversation** as a dynamic overlay that floats above the composer and updates live.

Generation still flows through the existing chat pipeline (`GlobalChatStore` → WebSocket `chat_message` with `media_generation`), so results are persisted in thread history *and* appear on the canvas.

## Composer + inline controls

The composer is the bottom-centered surface. Workflow controls live in its footer: an always-visible **Run** button (with running timer + queue badge, plus a contextual **Stop** while running) and a **⋮** button that opens a normal dropdown for everything else.

Canvas composer footer

Workflow actions menu

## Dynamic conversation overlay

A floating overlay shows the active thread above the composer, reusing `ChatThreadView` so messages, media output and streaming indicators render exactly like the chat panel. It auto-opens when a new message arrives or a generation starts, carries a message-count badge on the composer's conversation toggle, and can be collapsed. (Screenshot uses the test backend's "fake" provider in place of a live LLM.)

Conversation overlay

## Key changes

- **`useGenerationToCanvas`** (`web/src/hooks/handlers/useGenerationToCanvas.ts`) — watches the active thread's message cache and converts new assistant media blocks into `nodetool.constant.{Image,Video,Audio}` nodes, positioned just right of the existing graph (2-column grid). A `pending` counter matches results to composer-initiated requests; the "seen" set is re-seeded on thread change so history/unrelated media never spawn stray nodes. Unit-tested.
- **`CanvasMediaComposer`** (`web/src/components/panels/CanvasMediaComposer.tsx`) — thin wrapper over `MediaChatComposer`; lazily connects chat, marks generations, and forwards workflow controls via a `leadingActions` slot. Intentionally sends **no `workflow_id`** (that would route the backend into `handleWorkflowMessage` and fail with "Workflow not found").
- **`ConversationOverlay`** (`web/src/components/panels/ConversationOverlay.tsx`) — floating, auto-updating thread view reusing `ChatThreadView`; New-conversation + collapse controls.
- **`FloatingToolBar`** (`web/src/components/panels/FloatingToolBar.tsx`) — now a thin host: composer + the inline Run/Stop/⋮ actions + the dropdown + the conversation overlay. The old always-on controls pill is gone.
- **`MediaChatComposer`** (`web/src/components/chat/composer/MediaChatComposer.tsx`) — adds optional `autoFocus` (false on canvas, so it doesn't steal keyboard focus) and a `leadingActions` footer slot (empty in the chat panel).

## Notes

- Composer width nudged 720→820px so chat mode keeps a single-row footer; height is unchanged.
- All checks pass locally: `tsc`, oxlint, and the web/panels/hook tests (incl. new `useGenerationToCanvas` tests).

🤖 Generated with [Claude Code](https://claude.ai/code)


georgi:
## Review: Changes Requested

The media composer feature itself (useGenerationToCanvas, CanvasMediaComposer) is well-implemented — good Zustand patterns, clean separation of concerns, solid tests.

However, this PR **removes the entire floating toolbar** (Run/Stop, Save, Add Node, Auto Layout, ComfyUI status, Instant Update, mobile menu, mini-map, view mode toggle) and replaces it with only the media composer. This is a major UX regression:

1. **Run/Stop buttons deleted** — reduced to keyboard-only (Ctrl+Enter / Escape), which is undiscoverable for most users.
2. **ComfyUI auto-connect logic removed** with no replacement — breaks ComfyUI workflow auto-connection.
3. **Save, Add Node, Auto Layout, mobile menu all gone** — core canvas controls inaccessible.

Additionally:
- **`pending` counter race condition**: If a generation errors out server-side (no assistant media message arrives), `pending` stays elevated permanently, causing the next unrelated assistant message to be incorrectly dropped onto the canvas. There's no timeout or error-path decrement.
- **Type assertion** (`as ChatOutgoingMessage`) hides the intentional `workflow_id` omission — use `Omit<ChatOutgoingMessage, 'workflow_id'>` instead.

**Recommendation:** The composer should be *additive* — stack it above/below the existing controls pill rather than replacing the toolbar entirely. The ComfyUI auto-connect `useEffect` must be preserved or relocated.

---

PR #3454, opened 2026-06-02
NodeTool Studio MVP — transcript-driven editor for short-form video

georgi:
## Summary

First end-to-end slice of **NodeTool Studio**: a transcript-driven editor docked inside the existing `TimelineEditor`. You write a script, each line becomes a voiced + captioned **beat** on the timeline, editing the script edits the video, and the existing WebCodecs export produces a matching MP4.

The MVP is deliberately one data type, one store, one panel, and a caption layer wired into code that already exists — reusing the sequence/generation/staleness/compositor/export engine rather than reimplementing it.

## Screenshots

**Populated editor** — Transcript panel (left) with three beats; timeline shows aligned **Captions** (subtitle), **B-roll**, and **Voiceover** tracks:

Studio editor — populated

**Beat selected** — clicking a line jumps the playhead and selects its clips (Beat 3 highlighted):

Studio editor — beat selected

**Empty state** — the panel coexists with non-Studio sequences:

Studio editor — empty

**Transcript panel** close-up:

Transcript panel

> Note: the preview shows "WebGPU not available" because the headless CI container's SwiftShader doesn't expose WebGPU — captions composite through the real GPU path in a browser/export and are covered by unit tests.

## What's included

**Data model**
- `CaptionWord` / `ClipCaption` (clip-local word timing) and `TranscriptLine` added to the protocol schema and `@nodetool-ai/timeline` types.
- Clips carry an optional `caption`; the timeline document carries an optional `transcript`, persisted in the JSON document blob so autosave, undo, and export inherit it for free (round-tripped through the models field-mappers and the autosave PATCH).

**Transcript store**
- `TimelineStore` gains `transcript` document state + an atomic `setTranscriptAndClips` commit (one undo entry per edit).
- `transcriptOps` — pure reducers (add / edit / delete / reorder / reword + reflow) over the `{ transcript, clips }` slice. Reword marks the bound voiceover (and populated caption) stale via the existing dependency machinery; delete/reorder re-flow beat positions end-to-end.
- `TimelineTranscriptStore` orchestrates the deterministic per-beat pipeline: text → text-to-audio (`generate_media` RPC) → probe real audio duration → transcribe (`transcribe_audio` RPC) → re-flow.

**Captions (one render path)**
- New dedicated `transcribe_audio` RPC on the unified WebSocket runner (mirrors `generate_media`, uses the provider ASR capability), returning word-level timing in ms.
- A caption layer wired into `computeActiveLayers`: subtitle-track caption clips resolve to per-frame word state, then rasterize via a shared `CaptionRasterizer` fed through the **same GPU compositor** in both the live preview and the WebCodecs export — no export-specific drawing.

**UI**
- `TranscriptPanel` (ui_primitives only) docked beside the Inspector: inline edit, add/delete/reorder, per-line regenerate, click-to-seek + select bound clips.

## Bug fixed (surfaced by the screenshots)

The timeline `update` route rebuilt the saved document from `tracks/clips/markers` only, silently dropping `transcript`. Because the editor autosaves on load, the first save stripped the transcript — clips survived but the panel reloaded empty. The route now merges `transcript` through (with fallback to the stored value), covered by router regression tests.

## Tests

- `transcriptOps` reducers (add/edit/delete/reorder/reword→stale, reflow) and the caption layer + `resolveCaptionAtTime` in `computeActiveLayers`.
- `transcribe_audio` RPC against a fake ASR provider (seconds→ms conversion, empty-word filtering, word-timestamps contract, error paths).
- Timeline router regression: transcript persists through `update` and is preserved when a patch omits it.
- A Playwright `studio-screenshots` spec capturing the scenarios above against the seeded screenshot backend.

Local results: web typecheck/lint clean; timeline web suites, `TimelineEditor`, and backend `protocol` / `timeline` / `models` / `websocket` packages all green.

> Scope intentionally excludes the assembly agent, research, b-roll generation, brand kits, auto-reframe, localization, and platform publishing.

https://claude.ai/code/session_01HMoAqoSsZhNCKfu3dzH1EQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMoAqoSsZhNCKfu3dzH1EQ)_


georgi (commented):
This is a substantial feature PR (Studio MVP) that needs a thorough manual review and live testing — can't be auto-merged. A few things to address before it's ready:

1. **Base is stale** — main has moved forward with 6 merged PRs. Needs a rebase.
2. **Scope is large** — new data model, new store, new RPC, new panel, caption rendering, and a bug fix, all in one PR. Consider whether the bug fix (timeline `update` route dropping `transcript`) should be a separate PR.
3. **Screenshots reference a branch-specific URL** — the raw GitHub image URLs will break if the branch is deleted after merge. Consider uploading them as PR comment attachments instead.
4. **Needs live verification** — the PR notes "WebGPU not available" in CI, so caption compositing can't be validated without a real browser. This should be tested manually before merge.

---

PR #3478, opened 2026-06-04
Concurrent same-workflow runs with per-run canvas lens + ambient liveness

georgi:
## Summary

Lets the same workflow run **multiple times concurrently**, with the canvas acting as a per-run lens and clear visual feedback for the runs you're *not* currently viewing.

Previously all node execution state was keyed by `workflowId` alone, so a second run of the same workflow overwrote the first run's UI state (status, results, progress, edges, timing). This PR re-keys every per-node store by **`(workflowId, jobId, nodeId)`** and resolves the workflow's *focused run* in one place, so concurrent runs stay fully isolated while the canvas shows one at a time.

This also fixes the original bug it grew out of: a node could stay "running" in the UI after the backend completed, because a second enqueued run desynced the single-run state.

## What end users get

- **Run a workflow several times at once** — press **Run** again (or run-from-here / single node) without waiting for the first to finish. Opt-in via a `concurrent` flag on the run path; the default per-workflow serialization is unchanged for everything that doesn't ask for it.
- **Live queue** — the Queue panel lists every active run; the one marked **"On canvas"** drives the canvas. Newest run is focused automatically; click another to pin the canvas to it.
- **Ambient liveness** — nodes glow with a soft secondary-color **halo** while a run you're *not* viewing executes them, so the canvas never looks idle when work is happening off-focus.
- **Concurrency badge** — a node shows a count (e.g. `2`) when ≥2 runs are executing it at the same moment. A lone run stays badge-free.
- **Selection vs. run animation** — when a running node is also selected, selection moves *inside* (inset outline) and the run ring stays *outside*, so they no longer fight over the border.

## Implementation highlights

- **Per-job store keys** — `StatusStore`, `ErrorStore`, `ExecutionTimeStore`, and `ResultsStore` (results/outputs/progress/edges/chunks/tasks/toolCalls/planning/cost) are keyed `(wf, job, node)`. Branded `NodeKey`/`EdgeKey` types (`nodeKey()`/`edgeKey()`) make malformed keys a compile error.
- **`WorkflowRunsStore`** — per-workflow run registry (`queued → running → terminal`) plus a focused-job lens; auto-follows the newest run unless the user pins one.
- **Accessor hooks** (`useNodeExecState.ts`) — all per-node reads (`useNodeStatus`, `useNodeError`, `useNodeProgress`, `useNodeExecutionDuration`, `useEdgeStatus`, results/artifacts, and the new `useNodeActiveRunCount`) resolve the focused run in one indirection layer, so consumers were a one-line swap.
- **`workflowUpdates` reducer** — writes per-job from each message's stamped `job_id`; populates the run registry from `job_update`s; no longer wipes sibling runs on terminal updates.
- **Backend (websocket)** — `run_job` honors an opt-in `concurrent` flag through the job queue so same-workflow runs can run in parallel instead of serializing.
- **Hydration** — persisted asset results load under a synthetic run that never steals focus from a live run.

## Testing

- `web`: typecheck clean, **9124 tests pass** (incl. new `WorkflowRunsStore`, `concurrentRuns`, `ambientLiveness.repro`, `useNodeActiveRunCount`, per-job store tests, hydration/focus, runs-registry).
- `packages/websocket`: **687 tests pass** (incl. concurrent-run queue + runner cases).
- Manually verified live: concurrent Grok video runs, ambient halo on background-run nodes, `2` badge when both runs hit a node.

## Known follow-ups (not in this PR)

- Delete the now-vestigial `job_id` fallback WebSocket subscription (the backend stamps `workflow_id` on every message, so it bails in the common path and is incomplete under concurrency).
- Prune per-job store slices + `WorkflowRunsStore` on run-removal / workflow-close (the per-job keyspace currently isn't freed within a session).
- Deferred: a DB-backed across-run results gallery on output nodes.

Design + plan: `docs/superpowers/specs/2026-06-03-concurrent-same-workflow-runs-design.md`, `docs/superpowers/plans/2026-06-03-concurrent-headless-generation.md`.

### DIFF
--- a/docs/superpowers/plans/2026-06-03-concurrent-headless-generation.md
+++ b/docs/superpowers/plans/2026-06-03-concurrent-headless-generation.md
@@ -1,0 +1,791 @@
+# Concurrent Headless Generation — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let multiple sketch layers / timeline clips that share one workflow template generate **concurrently**, correctly (each resolves its own result), without changing the editor canvas's single-run model.
+
+**Architecture:** Relax the backend's per-workflow serialization behind an opt-in `concurrent` flag on the run request (the editor canvas does not set it, so it stays serialized until Plan 2). The sketch/timeline job handlers stop reading the shared `workflowId:nodeId` result store and instead resolve each job's output asset from *that job's own* `output_update` messages, so concurrent same-workflow jobs can't clobber each other.
+
+**Tech Stack:** TypeScript, Zustand, Fastify/WebSocket runner (`packages/websocket`), Vitest (packages), Jest (web).
+
+**Prerequisite (already landed):** `WorkflowRunner.run()` returns the id of the run it initiated; the generation hooks use it. Commit `1f59391b9`.
+
+---
+
+## File structure
+
+- `packages/websocket/src/job-queue.ts` — add an optional `concurrent` flag to the queued-job shape so `drainQueue` can tell concurrent runs apart.
+- `packages/websocket/src/unified-websocket-runner.ts` — `RunJobRequest.concurrent`; skip the per-workflow gate for concurrent runs in `runJob` and `drainQueue`.
+- `web/src/stores/WorkflowRunner.ts` — thread a `concurrent` arg through `run()` → `buildRunJobData` → the `run_job` payload.
+- `web/src/stores/outputAssetId.ts` *(new)* — pure `extractAssetId(value)` helper shared by both generation stores.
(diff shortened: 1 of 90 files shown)


georgi:
Addressed the Copilot review in d8706597e:

- **`useGenerateLayer` broad error clear** (real isolation bug) — removed the `clearErrors(workflowId)` before a run. Errors are keyed `(workflowId, jobId, nodeId)` and each run uses a fresh job id, so its slice already starts empty; the workflow-wide clear was wiping other concurrent runs' errors.
- **`recordRun` JSDoc** — corrected to the actual rule (latest-run-wins unless pinned).
- **`useNodeActiveRunCount` docstring** — corrected to 'any non-terminal run' (matches the gate that tolerates lagging run-state).
- **`runJob` gate comment** — notes the `concurrent` opt-in (no longer strictly one run per workflow).
- **hydration `if/continue`** — expanded to a multi-line block.

CodeQL's 5 'inconvertible null comparison' flags were fixed earlier in 1ee8ee88c.

PR #3508, opened 2026-06-06
Fix Gemini tool response name resolution and merge parallel results

georgi:
## Summary
Fixes tool response handling in the Gemini provider to correctly resolve function names from tool call IDs and merge multiple parallel tool results into a single user turn, maintaining proper message role alternation.

## Key Changes
- **Tool name resolution**: Added `toolNameById` map to track the mapping between generated call IDs and function names from assistant tool calls. Tool results now resolve their function name from this map instead of using the call ID directly.
- **Parallel tool result merging**: Multiple consecutive tool results (from parallel tool calls) are now merged into a single user turn with multiple `functionResponse` parts, rather than creating separate user turns. This maintains the required alternating user/model role pattern in Gemini's API.
- **Implementation details**:
  - Records tool call ID → name mapping when processing assistant messages
  - When processing tool results, resolves the function name using the mapping with fallback to the call ID
  - Checks if the previous content is a user turn containing only `functionResponse` parts, and appends to it rather than creating a new turn

## Testing
Added two new test cases:
- `resolves functionResponse name from the call id`: Verifies that tool results correctly resolve function names from the ID mapping
- `merges parallel tool results into one user turn`: Confirms that multiple tool results are merged into a single user turn with multiple parts

This ensures the Gemini provider correctly handles the real-world flow where tool results carry generated call IDs rather than function names.

https://claude.ai/code/session_01E6zpTK5XQxdsG44txM3gUp


georgi (commented):
Code is correct — `toolNameById` map for Gemini name resolution and parallel tool result merging are well-implemented.

**Blocked: merge conflicts with main.** Likely from #3515 which also modified the Gemini provider. Please rebase, then ready to merge.

---

PR #3514, opened 2026-06-06
fix(llm-nodes): correct media output refs, redacted-thinking parsing, and Gemini/OpenAI request shapes

georgi:
Gemini (gemini.ts):
- TextToSpeech output now includes `type: "audio"` so autoSaveAsset and
  downstream type detection recognize it (isAssetLikeValue requires `type`).
- Imagen image + Veo video requests use the `instances`/`parameters`
  envelope (matching the runtime Gemini provider); the prior `{prompt, config}`
  body is the JS-SDK shape and is rejected by the REST API.
- Image parsing also accepts Vertex-style `predictions[].bytesBase64Encoded`.
- Video extraction handles a returned download `uri` (downloads + base64),
  not only inline `videoBytes`.

OpenAI (openai.ts):
- EditImage no longer sends `response_format` (gpt-image-1 rejects it).
- CreateImage/EditImage guard an empty/missing `data` array.
- RealtimeAgent now emits the declared `audio` output (was never produced).

Agents (agents.ts):
- Streaming redacted-thinking splitter recognizes `</redacted_thinking>`
  (the close-tag list previously held two identical `</think>` entries).
- Extractor/Classifier gain a configurable `max_tokens` prop (the code read
  an undeclared one).
- Removed the broken `{self.max_tokens}` placeholder from the Summarizer
  default system prompt.

Generators (generators.ts):
- parseRequestedCount no longer ignores counts of 4+ digits (clamps instead).

https://claude.ai/code/session_01QBXPxDoBHpQcjmfzRpxPV6


georgi (commented):
Code looks good — Gemini REST API shape fix, OpenAI EditImage fixes, redacted-thinking close tag, and `max_tokens` props are all correct.

**Blocked: merge conflicts with main.** Likely from #3515 which touched overlapping provider areas. Please rebase onto current main, then this is ready to merge.

---

PR #2908, opened 2026-05-02
Multi-user deployment system, DB-backed config, per-user RunPod isolation

georgi:
## Summary

Lets a single Docker host serve multiple users, each deploying to backends like RunPod in isolation from every other user. Deployment config moves from `deployment.yaml` to the database, partitioned by `user_id` — the same convention already used by `nodetool_workspaces`, `nodetool_secrets`, and the rest of the user-owned schema.

There is no separate "tenant" concept: a tenant is a user.

### Database schema (SQLite + PostgreSQL mirrors)

- **`nodetool_deployments`** — one row per deployment, `(user_id, name)` unique. `config_json` and `state_json` are split so state writes don't round-trip the full config.
- **`nodetool_deployment_settings`** — one row per user. Holds the user's deployment quota (`quota_json`) and encrypted provider credentials (`credentials_json`). PK is `user_id`.
- **`nodetool_deployment_audit`** — append-only audit log indexed by `(user_id, ts)`.

### Models (`@nodetool-ai/models`)

- `Deployment` — `findByName(userId, name)`, `listForUser(userId)`, `upsert`, `writeState`, `remove`, `removeAllForUser`.
- `DeploymentSettings` — `findByUserId`, `getOrDefault`, `upsert`, `remove`.
- `DeploymentAudit` — `append`, `listForUser`, `listForDeployment`, `removeAllForUser`.

### Deploy package

- **`DbDeploymentStore`** — DB-backed replacement for `loadDeploymentConfig` + `StateManager`. Constructed per user; cross-user reads are impossible because `user_id` is baked into every query.
- **`DbDeploymentSettingsStore`** — DB-backed quotas + encrypted credentials. Provider credentials encrypted with `encrypt(masterKey, userId, plaintext)` so a leaked row + leaked master key cannot decrypt another user's credentials.
- **`UserDeploymentManager`** — top-level orchestrator. Wraps the existing `DeploymentManager` with:
  - **Ownership** — every operation scoped by `user_id`; cross-user access impossible at the SQL layer.
  - **Quotas** — `max_deployments`, `max_workers_per_endpoint`, `max_gpu_count_per_endpoint`, `allowed_providers`, `allowed_gpu_types`. Enforced before delegating.
[...]

PR #3555, opened 2026-06-07
feat(deploy): multi-user deployment system with per-user isolation, secret-sourced credentials, no host trust

georgi:
## Summary

Adds a **multi-user deployment system** and hardens it so every operation is isolated by `user_id`, provider credentials come only from the per-user secrets store, and nothing relies on `process.env` mutation, a global lock, or host-installed auth files. A single host can now drive deployments for many users to RunPod / Docker-over-SSH / GCP / Fly / Railway / HuggingFace, each with their own credentials, quotas, and audit log.

This supersedes the previous file-based (`deployment.yaml`) single-user model: deployment config moves into the database, partitioned per user.

## Design goals (and how they're met)

1. **Users, not tenants** — "tenant" terminology removed throughout; `user_id` everywhere.
2. **All data isolated by user** — every query filters `user_id`; unique index on `(user_id, name)`; deployments, settings, and audit are per-user tables.
3. **Never write into `process.env`** — a typed `DeploymentContext { userId, credentials, scratchDir }` threads per-call decrypted credentials through `DeployerFactory → deployers → API helpers`. `runScopedCommand` builds the **child** process env explicitly (curated base allowlist + `ctx.credentials`); the parent `process.env` is never mutated.
4. **No global locks/mutex** — the old process-wide `EnvMutex` is deleted. Same-deployment safety is now an **optimistic etag compare-and-swap** on `Deployment.writeState`/`upsert` (raises `DeploymentConflictError`), which is cross-process safe and never serializes unrelated users.
5. **Separation of concerns** — credentials (`provider-credentials.ts`), quota enforcement (`deployment-quota.ts`), the exec/credential envelope (`deployment-context.ts`), persistence (models), and orchestration (`user-deployment-manager.ts`) are each their own unit.
6. **Security** — per-user AES-256-GCM secrets (PBKDF2 salt = `user_id`), credentials held only transiently in `ctx`/child-env and wiped with the scratch dir, list endpoints expose names only, quota enforced before + during `apply`, append-only audit.
7. **No host-installed creds/files** — see below.

## No host trust

Provider auth no longer depends on ambient host state. All auth material is the calling user's secrets, delivered per-call:

- **Docker registry** — `docker login --password-stdin` into a scratch `DOCKER_CONFIG`; never `~/.docker/config.json`.
- **GCP** — service-account key from the `GCP_SERVICE_ACCOUNT_KEY` secret written to a scratch file; `GOOGLE_APPLICATION_CREDENTIALS` + `CLOUDSDK_CONFIG` set in the child env; no ambient `gcloud` auth. The GCP module was redesigned around an explicit `GcpScope { run, env, multiUser }`.
- **Fly / Railway** — `FLY_API_TOKEN` / `RAILWAY_TOKEN` passed via child env; no `~/.fly` etc.
- **SSH (Docker self-hosted)** — private-key **material** from the `SSH_PRIVATE_KEY` secret passed straight to `ssh2`'s `connectConfig.privateKey`; no `key_path` file read or `SSH_AUTH_SOCK` agent in the multi-user path.

Tool *binaries* (docker, gcloud, flyctl, ssh) may remain host-installed — only auth material and auth/config files must come from the user's secrets / scratch dir.

## Data model

New per-user tables (SQLite + Postgres parity):

- `nodetool_deployments` — one row per deployment; `config_json` (Zod-validated, minus state) + `state_json` split so state writes don't round-trip config; `etag` for optimistic concurrency; unique `(user_id, name)`.
- `nodetool_deployment_settings` — per-user **quota only** (the previous `credentials_json` column is dropped; deployment credentials are now ordinary per-user secrets).
- `nodetool_deployment_audit` — append-only activity log.

A one-shot `migrate-yaml-to-db` importer carries legacy `deployment.yaml` into the DB.

## Testing

- `deploy`: **1229 tests pass**, typecheck clean.
- `models`: **619 tests pass**, typecheck clean.
- Coverage includes: `process.env` is unmutated across a deployer call, optimistic-etag conflict (one writer wins, the other gets `DeploymentConflictError`), scratch-dir lifecycle, least-privilege credential resolution (only the provider's declared keys are read), and full per-provider deployer suites migrated to the explicit-credential API.

## Out of scope (follow-up)

- **REST routes + web UI** for the deployment lifecycle — the backend is structured to be UI-driven; the API/UI layer is a separate spec.

## Known issues / notes

- **Localhost `destroy` (pre-existing divergence, not fixed here):** on the *localhost* path, `LocalExecutor.execute(check=false)` swallows command failures, so `DockerDeployer.destroy()` reports success even if `docker stop/rm` fail locally. Tests reflect current behavior; whether localhost should surface these failures is a follow-up decision.
- **CLI typecheck (pre-existing, environmental):** `packages/cli/src/nodetool.ts` (untouched here) fails local `tsc` with `Cannot find module '@nodetool-ai/{minimax,reve,huggingface}-nodes'` — a local full-build/module-resolution gap unrelated to this change.


georgi (commented):
CI Quality Gate and CodeQL both failing. This is a large deployment system overhaul (supersedes #2908, which I've closed). The design looks solid — `DeploymentContext` threading over `process.env` mutation, optimistic etag CAS over global mutex, per-user credential scoping. But CI needs to pass before this can merge. Please investigate and fix the failures.

---

PR #3556, opened 2026-06-07
Multi-asset list inputs + fix partial-run list aggregation

georgi:
## Summary

Three related changes around feeding **multiple media assets** into AI/agent nodes:

1. **List inputs on AI/agent nodes.** Several nodes that took a single image/audio/video now take a list. Single values still auto-wrap to a 1-element list, so existing graphs keep working.
   - `AgentNode`: `image → list[image]`, `audio → list[audio]`
   - `StructuredOutputGenerator`, `SVGGenerator`: `image/audio → list[image]/list[audio]`
   - `ImageAgent`: `image → list[image]`
   - `MediaAgent`, `FfmpegAgent`: `audio/video → list[audio]/list[video]`
   - DSL input types regenerated to match.

2. **Partial-run list aggregation fix.** When you run a single node ("Run Node" / "Run from here") or "run selected nodes", upstream values that aren't part of the partial graph are injected as property overrides. Two edges wired into one `list[image]` handle were collapsed to a single image (**last-write-wins**). They now aggregate into a list when the handle is a list/collect type — mirroring the kernel's existing multi-edge list aggregation for full runs.

   The two partial-run paths (`useRunSelectedNodes` and `buildRunSubgraph`) had **independently duplicated** the override-building, the list-aggregation decision, and the static/dynamic property merge — which is why the same bug had to be fixed twice. Consolidated all of it into a shared `EdgeOverrideCollector` + `applyNodeOverrides` in `web/src/utils/edgeOverrides.ts`.

3. **Persist agent-generated media.** `ImageAgent`, `MediaAgent`, `FfmpegAgent`, and `YtDlpDownloaderAgent` produce real image/audio/video assets but were session-only. They now set `autoSaveAsset = true`, so their outputs are written to the assets table (survive reload, appear in node history, reusable as cached upstreams) like other generative media nodes.

## Behavioral note

Marking those agents as generative also reclassifies them in `buildRunSubgraph`: an **uncached** instance feeding a downstream partial run now **blocks** with "Run X first" instead of being silently re-executed — consistent with how fal/replicate nodes behave, and preferable for expensive LLM+tool agents.

## Verification

- `code-nodes` (96) and `llm-nodes` (451) suites pass.
- New `web` unit tests: `edgeOverrides.test.ts`; regression tests added to `runSubgraph.test.ts` and `useRunSelectedNodes.test.ts`.
- `web` typecheck + lint clean; `build:packages` (tsc per package) clean.
- Manually verified a real FAL multi-image edit endpoint (`qwen-image-edit-2509`) accepts a 2-element `image_urls` array end-to-end through the package's upload/submit path.

## Out of scope (follow-ups)
- Surfacing FAL's 422 validation `detail` (the generic "Unprocessable Entity" hides the real cause, e.g. `image_too_small`) — useful, but kept separate to keep this PR focused.
- A reported content-card display issue for a fal edit node after partial runs is still under investigation and is unrelated to this PR's display path (verified sound).

### DIFF
--- a/packages/code-nodes/src/nodes/tool-agents.ts
+++ b/packages/code-nodes/src/nodes/tool-agents.ts
@@ -1012,6 +1012,9 @@ export class EmailAgentNode extends ToolAgentNode {
 
 export class FfmpegAgentNode extends ToolAgentNode {
   static readonly nodeType = "nodetool.agents.FfmpegAgent";
+  // Produces real audio/video assets via set_output_audio/video; persist them
+  // on job completion like other generative media nodes.
+  static readonly autoSaveAsset = true;
   static readonly requiredRuntimes = ["ffmpeg"];
   static readonly title = "FFmpeg Agent";
   static readonly inlineFields = [];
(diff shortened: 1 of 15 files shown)


georgi:
@claude fix CI

PR #3597, opened 2026-06-10
Fix streaming node secrets, metadata parsing, and registry warnings

georgi:
## Summary

This PR hardens the node SDK and metadata loading with fixes for secret injection in streaming nodes, improved Python JSON parsing, safer cache directory handling, and better error diagnostics.

### BaseNode & Streaming Secrets
- **Streaming `run()` now receives injected secrets**: Extract `_resolveSecrets()` as a separate method and call it before `run()` to populate `_secrets` on the instance, matching the behavior of `process()`. Streaming nodes can now access API keys via `this._secrets`.
- **Explicit `isStreamingOutput = false` now opts out of inference**: Add `explicitStreamingOutputFlag()` helper to distinguish between an explicit `false` declaration (real opt-out) and the default `false` (which triggers inference). This allows nodes with `genProcess()` to suppress streaming output if they explicitly declare `isStreamingOutput = false`.
- **Inherit streaming flags from ancestors**: The flag resolution now walks the prototype chain below `BaseNode`, so a child class inherits an explicit flag from its parent.
- **Fix `assign()` to treat explicit `undefined` as absent**: When a property is explicitly set to `undefined`, treat it as if the key was omitted, so declared defaults apply instead of wrapping `undefined` into arrays.

### Metadata & Python JSON Parsing
- **Safer Python JSON sanitization**: Replace the naive regex-based approach with `sanitizePythonJson()` that properly tracks string boundaries. This preserves `NaN`/`Infinity` words inside string literals (e.g., descriptions) while nulling bare tokens outside strings.
- **Lazy sanitization on parse failure**: Only apply the sanitizer if the first JSON parse fails, avoiding unnecessary processing for valid JSON.
- **Per-user metadata cache directory**: Move cache from the world-writable system tmpdir to `$XDG_CACHE_HOME/nodetool/metadata-cache` (or `~/.cache/nodetool/metadata-cache` as fallback) to prevent symlink/poisoning attacks by other local users.

### Package Discovery & Exports Resolution
- **Nested export conditions**: Enhance `pickExportCondition()` to recursively resolve nested condition objects (e.g., `{ import: { types: "...", default: "x.js" } }`), with depth capping to prevent cycles.
- **Flexible metadata path reporting**: Change `findWrittenMetadataName()` to `findWrittenMetadataPath()` and accept both absolute and relative paths with any separator (`/` or `\`), supporting both `src/nodetool/package_metadata` and flat `nodetool/...` layouts.
- **Windows PATH separator handling**: Fix `readEnvPackPaths()` to not split on `:` on Windows (which would shear drive letters off absolute paths like `C:\…`).

### Registry & Search
- **Warn on node type collisions**: `NodeRegistry.register()` now warns when a different class replaces an existing node type, helping catch accidental collisions while still allowing hot reload (re-registering the same class stays silent).
- **Provider nodes with explicit namespace prefix**: When `scoreNodeMetadata()` receives an explicit `namespacePrefix`, treat it as an opt-in request for provider nodes, overriding the default exclusion. This allows searching for `openai.*` nodes without needing `includeProviderNodes: true`.

### Tests
- Added comprehensive test coverage for all new behaviors: streaming secrets, flag inheritance, metadata sanitization, path parsing, export resolution, and registry warnings.

## Implementation Details

- `_resolveSecrets()` is now a separate async method that returns a secrets map, called by both `_injectSecrets()` (for `process()`) and the streaming `run()` executor wrapper.
- The `explicitStreamingOutputFlag()` helper uses `Object.prototype.hasOwnProperty.call()` to detect own properties only, ensuring inherited defaults don't interfere.
- `sanitizePythonJson()` uses a simple state machine to track string boundaries and escape sequences, replacing bare `NaN`/`Infinity`/`-Infinity` tokens with `null`.
- Cache directory logic uses `process.env["XDG_CACHE_HOME"]` with a homedir fallback, following XDG Base

https://claude.ai/code/session_01TvPa5tdApf9eCWHqc1AkuJ

### DIFF
--- a/packages/base-nodes/tests/coverage-io-control-nodes.test.ts
+++ b/packages/base-nodes/tests/coverage-io-control-nodes.test.ts
@@ -516,9 +516,15 @@ describe("output nodes — full coverage", () => {
     node.assign({ value: { a: 1 } });
     expect(await node.process()).toEqual({ output: { a: 1 } });
 
-    // undefined (fallback to null)
-    node.assign({ value: undefined });
-    expect(await node.process()).toEqual({ output: null });
+    // undefined → assign() treats it as absent, so a fresh node keeps its
+    // declared default (null); process() normalizes that to null output.
+    const fresh = new OutputNode();
+    fresh.assign({ value: undefined });
+    expect(await fresh.process()).toEqual({ output: null });
+
+    // a raw undefined reaching process() directly still falls back to null
+    fresh.value = undefined;
+    expect(await fresh.process()).toEqual({ output: null });
   });
 
   it("OutputNode output_name defaults to 'output' when name prop empty", async () => {
--- a/packages/node-sdk/src/base-node.ts
+++ b/packages/node-sdk/src/base-node.ts
@@ -436,7 +436,7 @@ export abstract class BaseNode {
       } else {
         console.warn(
           // Stryker disable next-line StringLiteral: operator diagnostic text only.
-          `[_injectSecrets] Secret "${key}" not found for ${ctor.nodeType}`
+          `[_resolveSecrets] Secret "${key}" not found for ${ctor.nodeType}`
         );
       }
     }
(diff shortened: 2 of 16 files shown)


georgi (commented):
Code looks solid. The CI failure is a pre-existing flaky perf test (`PrefixTreeSearch.test.ts` — 5.04ms vs 5ms threshold), completely unrelated to this PR.

Changes are well-tested and address real issues:
- Streaming `run()` now gets injected secrets (was a real bug)
- `sanitizePythonJson` properly tracks string boundaries instead of naive regex
- XDG-based cache dir prevents symlink attacks from world-writable tmpdir
- Registry collision warnings are helpful for debugging

Ready to merge once the flaky test is stabilized or CI is re-run.

---

PR #3638, opened 2026-06-12
perf: overhaul realtime audio pipeline for browser execution

georgi:
## Summary

A live synth patch overloaded the editor — UI slowdown and audio dropouts shared one root cause: every intermediate module's chunk stream crossed worker→main per chunk, fed React state per message, and worklet delivery rode the React render cycle.

**Kernel / messages**
- `output_update` is now suppressed for handles with outgoing data edges (downstream gets the value on the edge; clients display terminal handles only). UI-monitor nodes (AudioOutput) opt out via a new `always_emit_output_updates` descriptor flag plumbed through protocol/node-sdk/hydration.
- `ProcessingContext` gains `retainMessageQueue: false` for listener-based consumers — the undrained pull queue retained every streamed chunk in the worker.

**Main thread**
- React-free audio fast path: a chunk bus hands audio chunks from the message handler straight to the playback worklet in the same task; store appends remain for replay/restart, coalesced to one set per 200ms.
- `GlobalWebSocketManager` no longer logs unrouted/unhandled messages with payloads — the console buffer pinned every logged chunk (+187MB retained per minute at 32 voices).
- Removed a dead per-frame rAF loop in the playback hook.

**Worker / DSP**
- Audio-chunk messages skip the async image-resolve walk; mono fast path in streaming filters (1 allocation vs 3 + two copies per chunk); `StreamingGain` no longer mutates the shared upstream buffer in place.

**Perf test environment**
- `web/perf-realtime.html` harness runs parametric synth patches in the real browser-runner worker.
- Playwright spec (60s default) measures main-thread CPU (CDP `TaskDuration`), worker event-loop lag / GC pauses, heap GC sawtooth, and per-realm retained memory via `measureUserAgentSpecificMemory` (perf page cross-origin isolated through a scoped vite middleware).
- `PERF_PATCH=complex` builds a full modular patch (master clock, S&H pitch CV, FM/filter LFOs, ADSR/VCA, resonant VCF, mixer tree) for stress testing.

## Results

| 32 voices / 60s | before | after |
|---|---|---|
| main-thread CPU | 14.1% | **2.1%** |
| retained page growth | +187.6MB | **−1.2MB** |
| worker GC pauses | 1 (31.7ms) | 0 |
| chunk rate vs realtime | 1.000 | 1.000 |

Complex-patch ceiling: 1,199 streaming nodes hold 0.999 realtime with single-digit-ms worker jitter; saturation (93% realtime, multi-second stalls) begins between 1,200 and 2,400 nodes — CPU-bound, memory stays flat.

## Test plan

- [x] `npm run typecheck` / `npm run lint` (warnings pre-existing, untouched files)
- [x] Package suites: kernel (739), node-sdk (528), protocol (268), audio-nodes (62), runtime (1570), workflow-runner (12), websocket (735), dsl (138), agents (1059), base-nodes (1613)
- [x] Web jest: workflowUpdates / ResultsStore / GlobalWebSocketManager / RealtimeAudio suites
- [x] `cd web && PERF_PATCH=complex PERF_VOICES=128 npx playwright test tests/benchmarks/realtime-perf.spec.ts --project=chromium`

---

## Follow-up fix: the 30–60s dropout (single oscillator → AudioOutput)

After the work above, a trivial patch — one oscillator into one AudioOutput — still produced heavy audio dropouts beginning 30–60 seconds into a run. This turned out to be a separate root cause, found and fixed on this branch (`e0af84e47`).

### Investigation

The perf harness (which feeds the worklet directly, no React tree) played 90s of real headed audio with a rock-steady buffer — so the transport, kernel pacing, and audio-hardware clock were all exonerated. A new end-to-end repro (`web/scripts/repro-audio-dropout.mjs`) drives the **real editor**: creates the workflow via API, opens the canvas with the actual AudioOut node body, clicks Run, and samples new worklet telemetry every few seconds.

That repro showed a sharp phase transition:

- **t < 21s**: healthy — bus-path delivery, max inter-chunk gap 28ms, zero distress, page heap ~260MB.
- **t ≈ 21s onward**: delivery gaps of 600ms+, page heap climbing to 800MB, the worklet alternating between live-cap stale drops and underruns. Net effect: **31% of all audio dropped, ~2 underruns/sec, only ~67% of wall time rendering sound.**

t≈21s is exactly when the 1024-chunk rolling store window (`MAX_AUDIO_STREAM_CHUNKS`, ~21.8s of audio) first fills. A CPU profile of the degradation window named the culprit:

```
18.4%  run
 9.4%  addValueToProperties  @ react-dom_client.js
 3.9%  addObjectToProperties @ react-dom_client.js
```

`addValueToProperties` is **react-dom's dev-build performance instrumentation**: React 19 dev mode deep-serializes every component's props into performance-timeline entries on every commit. The entire rolling buffer — 1024 chunks with sample payloads, serialized **per element** for `Float32Array` content — was passed as `RealtimeAudioOutput`'s `chunks` prop. Once the window filled, every 200ms store flush serialized megabytes of audio data, the heap ballooned, and the resulting GC stalls (600ms+) exceeded the worklet's 100ms live buffer. **Onset time = time to fill the rolling window** — hence "dropouts after 30–60 seconds". Dev-build-only pathology (prod React doesn't serialize props), which is why it haunted `make dev` sessions.

### Fix

Chunk buffers never cross a React component boundary as a value anymore. `AudioOutBody` → `RealtimeAudioOutput` → `useRealtimeAudioPlayback` now pass a **stable `getChunks()` getter + a numeric `chunksVersion`** (bumped on buffer identity change); `OutputRenderer`'s preview path goes through a small `RealtimeAudioOutputFromChunks` adapter so the array stops at that boundary. The scheduling effect reads the buffer through the getter, keyed on the version — behavior is identical, but React never sees the data.

### Verification (same repro, 76s)

| | before | after |
|---|---|---|
| underruns | 186 / 100s | **1** (single transient at window-fill) |
| audio dropped as stale | 31% (748k frames) | ~0 after t=21s |
| page heap | climbing to ~800MB | flat at ~255MB |
| main thread (degradation window) | saturated, 600ms stalls | **92.5% idle** |
| consumer rate | 16.2k frames/s | 23.8k ≈ realtime |

### Diagnosis tooling (kept in-tree)

- **Worklet buffer-health telemetry**: `chunkPlayerProcessor` emits 1 Hz stats (buffered frames, underruns, stale-dropped frames, frames in/out); `useRealtimeAudioPlayback` exposes them on `window.__nodetoolRealtimeAudio` with bus/effect delivery counters + max delivery gap, and logs a console warning whenever distress counters advance.
- **`PERF_PLAYBACK=1`** on the perf spec: routes the output into a real `AudioContext` + worklet (headed, real audio-hardware clock) and reports buffer slope / implied clock skew — the mode that ruled out clock drift.
- **`web/scripts/repro-audio-dropout.mjs`**: the full field repro against the running dev stack, including a CPU profile of the t=16–26s window.

Known residual: a single ~600ms blip when the rolling window first fills (one-time major GC of the accumulated buffer growth). If audible in practice, the next step is pre-allocating the window instead of grow-then-trim.

### DIFF
--- a/electron/src/menu.ts
+++ b/electron/src/menu.ts
@@ -4,6 +4,7 @@ import { getMainWindow } from "./state";
 import { createPackageManagerWindow, createLogViewerWindow, createSettingsWindow } from "./window";
 import { createChatWindow } from "./workflowWindow";
 import { getSystemInfo } from "./systemInfo";
+import { openPerformanceMonitorWindow } from "./perfMonitor";
 
 /**
  * Builds the application menu
@@ -234,6 +235,10 @@ const buildMenu = () => {
         {
           label: "Log Viewer",
           click: () => createLogViewerWindow(),
+        },
+        {
+          label: "Performance Monitor",
+          click: () => openPerformanceMonitorWindow(),
         },
         { type: "separator" },
         {
--- a/electron/src/perfMonitor.ts
+++ b/electron/src/perfMonitor.ts
@@ -1,0 +1,147 @@
+/**
+ * Performance Monitor window — Electron's stand-in for Chrome's Task Manager
+ * (which isn't available in Electron; `chrome.send('openTaskManager')` only
+ * exists on Chrome WebUI pages).
+ *
+ * Polls `app.getAppMetrics()` once a second and renders per-process CPU and
+ * memory into a small self-contained window (data: URL page, no preload, no
+ * IPC surface). Renderer processes are labeled with their window title /
+ * URL via `webContents.getAllWebContents()`.
+ */
+
+import { app, BrowserWindow, webContents } from "electron";
+import { hardenWebContents } from "./windowSecurity";
+
+interface MetricRow {
+  label: string;
+  pid: number;
+  cpu: number;
+  memoryMb: number;
+}
+
+let perfWindow: BrowserWindow | null = null;
+
+const PAGE_HTML = `<!doctype html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Performance Monitor</title>
+<style>
+  body { margin: 0; background: #111; color: #ddd; font: 13px/1.5 -apple-system, system-ui, sans-serif; }
+  table { width: 100%; border-collapse: collapse; }
+  th, td { text-align: left; padding: 4px 12px; white-space: nowrap; }
+  td.num, th.num { text-align: right; font-variant-numeric: tabular-nums; }
+  th { position: sticky; top: 0; background: #1b1b1b; color: #999; font-weight: 600; border-bottom: 1px solid #333; }
+  tr:nth-child(even) td { background: #161616; }
+  td.label { max-width: 360px; overflow: hidden; text-overflow: ellipsis; }
+  .hot { color: #ff9966; }
+</style>
+</head>
+<body>
+<table>
+  <thead>
+    <tr><th>Process</th><th class="num">PID</th><th class="num">CPU %</th><th class="num">Memory</th></tr>
+  </thead>
+  <tbody id="rows"><tr><td colspan="4" style="padding:16px;color:#777">Collecting…</td></tr></tbody>
+</table>
+<script>
+  window.__updateMetrics = function (rows) {
+    const esc = (s) => String(s).replace(/[&<>"]/g, (c) =>
+      ({ "&": "&amp;", "<": "&lt;", ">": "&gt;", '"': "&quot;" }[c]));
+    document.getElementById("rows").innerHTML = rows
+      .map((r) =>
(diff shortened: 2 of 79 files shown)


georgi (commented):
Impressive performance work (14.1% → 2.1% main-thread CPU). Two items to fix before merge:

1. **`retainMessageQueue` flag not propagated in `ProcessingContext.copy()`** — this is a real bug flagged by Copilot. If a context is copied, the new context won't retain the message queue setting.
2. **HTML escape in `perfMonitor.ts` missing single-quote escaping** — minor XSS risk in the Electron perf window, easy fix.

Also waiting for the 2 in-progress CI checks to complete. Please address these and this is ready.

---

PR #3766, opened 2026-06-19
[T-20260619-0001] #1 Tool-result capping crashes turn on undefined result

georgi:
Fixed the crash by adding `?? ""` to the inline `onToolCall` tool-result capping path in `packages/chat/src/message-processor.ts`, mirroring the already-guarded sibling site at line 278-280. A tool returning `undefined`/void makes `JSON.stringify` produce primitive `undefined`, which `truncateToolResult` then dereferenced `.length` on → TypeError that killed the whole turn.

**Changed:**
- `packages/chat/src/message-processor.ts` — wrapped the `typeof result === "string" ? result : JSON.stringify(...)` expression with `?? ""`.
- `packages/chat/tests/message-processor.test.ts` — added a regression test using a `VoidTool` returning `undefined`, driven through a mock provider that invokes the inline `onToolCall` callback. Verified it fails without the fix and passes with it.

**Verification:** all 18 tests pass; `tsc --noEmit` (lint) passes. The `npm run build` failure is a pre-existing worktree environment issue (missing root `typescript` binary), unrelated to this change.

---

Closes task **T-20260619-0001**: #1 Tool-result capping crashes turn on undefined result.

### Acceptance criteria
- [ ] Undefined/void tool result no longer throws in the inline onToolCall path
- [ ] Fix mirrors the guarded sibling site at line 278-280
- [ ] Regression test covers a tool returning undefined

### DIFF
--- a/packages/chat/src/message-processor.ts
+++ b/packages/chat/src/message-processor.ts
@@ -188,9 +188,9 @@ export async function processChat(opts: {
               const result = (executed as ToolCall & { result: unknown }).result;
               callbacks?.onToolResult?.(toolCall, result);
               return truncateToolResult(
-                typeof result === "string"
+                (typeof result === "string"
                   ? result
-                  : JSON.stringify(result, null, 2)
+                  : JSON.stringify(result, null, 2)) ?? ""
               );
             }
           : undefined,
--- a/packages/chat/tests/message-processor.test.ts
+++ b/packages/chat/tests/message-processor.test.ts
@@ -514,6 +514,47 @@ describe("processChat", () => {
     expect(recall).not.toHaveBeenCalled();
   });
 
+  it("tolerates a void/undefined tool result in the inline onToolCall path", async () => {
+    // A tool returning undefined makes JSON.stringify produce primitive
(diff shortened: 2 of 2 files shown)


georgi (commented):
**Verdict: APPROVE** (posting as a comment — GitHub blocks formally approving one's own PR).

Correct, tested, and consistent with the codebase.

**Fix is right.** `JSON.stringify(undefined)` returns primitive `undefined`; `truncateToolResult` immediately reads `text.length`, so the inline `onToolCall` path (message-processor.ts:190-194) threw and killed the turn. Wrapping the ternary in `(...) ?? ""` guards it and exactly mirrors the already-guarded sibling site at lines 278-280. The parens are placed correctly so `?? ""` applies to the whole ternary.

**Test is a genuine regression test.** I reverted the `?? ""` locally and re-ran: the new `VoidTool` test fails with the exact `TypeError: Cannot read properties of undefined (reading 'length')`, and passes once restored. It drives the real inline `onToolCall` callback path rather than asserting against a stub, so it pins the actual regression.

**Verification:** `npm run test --workspace=packages/chat` -> 29 passed.

All three acceptance criteria met. No further changes needed.

PR #3767, opened 2026-06-19
[T-20260619-0004] #4 WS message size cap bypassed on text/JSON frames

georgi:
Done.

Applied the same `NODETOOL_WS_MAX_MESSAGE_BYTES` cap to the text/JSON frame path in `receiveMessage()`, which previously bypassed the size check entirely. Text frames are now measured by their UTF-8 byte length and rejected with the identical error before `JSON.parse`, matching the binary path.

**Changes:**
- `packages/websocket/src/unified-websocket-runner.ts` — size-check text frames via `Buffer.byteLength(message.text, "utf8")` before parsing.
- `packages/websocket/tests/unified-websocket-runner.test.ts` — added two tests: oversized text frame rejected, and an in-limit text frame accepted.

Typecheck clean; all 5 size-guard tests pass.

---

Closes task **T-20260619-0004**: #4 WS message size cap bypassed on text/JSON frames.

### Acceptance criteria
- [ ] Text/JSON frames are size-checked against NODETOOL_WS_MAX_MESSAGE_BYTES before parsing
- [ ] Oversized text frame is rejected the same way as oversized binary frame
- [ ] Test covers an oversized text frame

### DIFF
--- a/packages/websocket/src/unified-websocket-runner.ts
+++ b/packages/websocket/src/unified-websocket-runner.ts
@@ -1234,6 +1234,15 @@ export class UnifiedWebSocketRunner {
       return unpack(message.bytes) as Record<string, unknown>;
     }
     if (message.text) {
+      const maxBytes = getMaxWsMessageBytes();
+      const textBytes = Buffer.byteLength(message.text, "utf8");
+      if (textBytes > maxBytes) {
+        throw new Error(
+          `Incoming WebSocket message exceeds maximum size: ` +
+            `${textBytes} > ${maxBytes} bytes ` +
+            `(set NODETOOL_WS_MAX_MESSAGE_BYTES to raise the limit)`
+        );
+      }
       return JSON.parse(message.text) as Record<string, unknown>;
     }
     return null;
--- a/packages/websocket/tests/unified-websocket-runner.test.ts
+++ b/packages/websocket/tests/unified-websocket-runner.test.ts
@@ -780,4 +780,25 @@ describe("UnifiedWebSocketRunner binary frame size guard", () => {
       /exceeds maximum size/
     );
   });
+
+  it("rejects an incoming text frame larger than the limit", async () => {
+    process.env[KEY] = "16";
+    ws.queue.push({
+      type: "websocket.message",
+      text: JSON.stringify({ type: "ping", padding: "x".repeat(256) })
+    });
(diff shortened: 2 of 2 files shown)


georgi (commented):
## Review: WS message size cap on text/JSON frames — LGTM ✅

(Posting as a comment — GitHub blocks approving one's own PR. This is an approval in substance.)

Verified against the acceptance criteria — all three met.

**What it does:** Closes the gap where the `NODETOOL_WS_MAX_MESSAGE_BYTES` cap only guarded the binary (`message.bytes`) path. The text/JSON branch now measures UTF-8 byte length and rejects before `JSON.parse`, mirroring the binary branch.

### Correctness
- ✅ The check is placed **before** `JSON.parse`, so an oversized payload is rejected without paying the parse cost (the actual DoS vector). Correct ordering.
- ✅ `Buffer.byteLength(message.text, "utf8")` is the right measure — counts bytes, not JS string length, so multibyte chars can't slip past a byte-denominated cap.
- ✅ Error message and structure are identical to the binary path, satisfying "rejected the same way." `getMaxWsMessageBytes()` is reused, so both paths stay in sync with the same env var.
- ✅ Symmetric with the binary branch at `unified-websocket-runner.ts:1225` — no behavioral drift.

### Tests
- ✅ Oversized text frame → rejects with `/exceeds maximum size/` (criterion 3).
- ✅ In-limit text frame → parses and returns. Good to cover the accept path too, guarding against an over-aggressive cap.
- Tests set `process.env[KEY]` directly, consistent with the existing binary-frame test block.

### Minor notes (non-blocking)
- `getMaxWsMessageBytes()` is now called per-branch; only one branch runs per call, so negligible. Pre-existing pattern from the binary path — matching it is the right call.

Tight, well-scoped fix that closes a real bypass with no over-engineering. CI: 7/9 green, 2 pending, 0 failures.

```json
{
  "verdict": "approve",
  "acceptance_criteria": {
    "text_frames_size_checked_before_parsing": true,
    "oversized_text_rejected_like_binary": true,
    "test_covers_oversized_text_frame": true
  },
  "blocking_issues": [],
  "summary": "Mirrors the binary-path size guard onto the text/JSON branch, checks UTF-8 byte length before JSON.parse, identical error, covered by both reject and accept tests. Correct and consistent."
}
```

PR #3770, opened 2026-06-19
[T-20260619-0010] #10 reconcile() flags stopped/paused entries as cost orphans; Vast findOffer ignores disk

georgi:
Done. Both worker issues fixed and committed.

## Summary

Fixed two Medium worker bugs in `packages/compute`:

1. **`reconcile()` flagged paused/stopped workers as cost orphans** — the live-but-untracked loop pushed *every* untracked provider entry into `orphans`, including paused (`stopped`) and destroyed (`terminated`) ones that aren't burning GPU billing. It now skips those statuses, so only `running`/`attached` untracked workers surface for the cost guard's one-click stop-all.

2. **Vast `findOffer` ignored disk** — the bundles query had no disk constraint, so the cheapest offer could be too small for the requested disk and reject the launch. Added `disk_space: { gte: spec.disk ?? DEFAULT_VOLUME_GB }` to the query.

### Files changed
- `packages/compute/src/manager.ts` — orphan loop skips `stopped`/`terminated` entries
- `packages/compute/src/providers/vast.ts` — `findOffer` constrains `disk_space`
- `packages/compute/src/__tests__/manager.test.ts` — test that stopped/terminated untracked entries aren't flagged
- `packages/compute/src/__tests__/vast-provider.test.ts` — test that the offer query carries the disk constraint

All 72 compute tests pass; `tsc --noEmit` (lint) clean.

---

Closes task **T-20260619-0010**: #10 reconcile() flags stopped/paused entries as cost orphans; Vast findOffer ignores disk.

### Acceptance criteria
- [ ] reconcile() does not flag stopped/paused entries as cost orphans
- [ ] Vast findOffer constrains disk so chosen offers can actually launch
- [ ] Tests cover both: a stopped entry not flagged, and a disk-constrained offer query

### DIFF
--- a/packages/compute/src/__tests__/manager.test.ts
+++ b/packages/compute/src/__tests__/manager.test.ts
@@ -676,6 +676,27 @@ describe("WorkerManager — reconcile", () => {
     ]);
   });
 
+  it("does not flag stopped/paused untracked provider entries as cost orphans", async () => {
+    const { manager, provider } = makeManager();
+    await manager.createProfile(PROFILE_INPUT);
+    const tracked = await manager.provision("hf-a40"); // pod-1
+
+    // Provider reports the tracked pod plus untracked entries that are not
+    // burning GPU billing (paused/destroyed) — these must not be surfaced.
+    provider.liveList = [
+      { providerRef: tracked.provider_ref, status: "running" },
+      { providerRef: "pod-paused", status: "stopped" },
+      { providerRef: "pod-gone", status: "terminated" },
+      { providerRef: "pod-orphan", status: "running" },
+    ];
+
+    const summary = await manager.reconcile();
+
+    expect(summary.orphans).toEqual([
+      { target: "runpod", providerRef: "pod-orphan", status: "running" },
+    ]);
+  });
+
   it("returns a live-count and estimated-cost summary", async () => {
     const { manager, models, provider } = makeManager();
     await manager.createProfile(PROFILE_INPUT);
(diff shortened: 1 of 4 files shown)


georgi (commented):
**Verdict: LGTM / would approve** (API blocks self-approval). Both fixes are correct, tested, and consistent with the codebase.

**reconcile() orphan filter** — `manager.ts` now skips `stopped`/`terminated` untracked entries, so only billing-relevant entries surface for the cost guard's stop-all. ✅ AC1.

**Vast findOffer disk** — the bundles query now carries `disk_space: { gte: spec.disk ?? DEFAULT_VOLUME_GB }`, using the *same* value `launch()` allocates (`vast.ts:228`), so search and launch agree and the chosen offer can fit the requested disk. ✅ AC2.

Tests cover both: stopped/terminated untracked entries excluded vs. running surfaced, and the disk-constrained query body. ✅ AC3.

**Minor (non-blocking) — comment vs. behavior:** The new comment in `manager.ts` reads *"Only running/attached entries are cost orphans,"* but the implementation is an exclusion list (skip `stopped`/`terminated`), so an untracked `provisioning` entry is still flagged. That's actually the right behavior for a cost guard (a provisioning worker is billing), but the comment overstates the allowlist. Also note that `list()` in both providers only ever emits `running`/`stopped`/`provisioning` — `mapPodStatus`/`mapInstanceStatus` collapse `TERMINATED`/`EXITED` → `stopped` — so the `=== "terminated"` branch is effectively dead, defensively-typed code. Both are cosmetic; consider rewording the comment to match the exclusion semantics.

PR #3771, opened 2026-06-19
[T-20260619-0006] #6 TDZ ReferenceError on pre-aborted login signal

georgi:
Fixed the TDZ ReferenceError in `LocalCallbackServer.waitForCode`. A pre-aborted `AbortSignal` ran `onAbort()` synchronously, which touched the `finish` `const` while it was still in its temporal dead zone — throwing a wrong error and leaking the cleanup timer (the `clearTimeout` lives inside `finish`).

The fix hoists `finish` and `onAbort` above the timer and the aborted-signal check. Because `finish` also references `timer` and `onRequest`, I ordered all three bindings before the synchronous pre-aborted path, so nothing is in its TDZ when `onAbort()` fires. The pre-aborted path now rejects with the correct `OAuthError("Login aborted")` and clears the timer.

- `packages/runtime/src/providers/oauth/local-callback-server.ts` — reordered declarations in `waitForCode`; moved the signal check after `server.on("request", onRequest)`.
- `packages/runtime/tests/providers/oauth/local-callback-server.test.ts` — added a regression test for the already-aborted signal.

Tests (6 passed), lint, and typecheck all pass.

---

Closes task **T-20260619-0006**: #6 TDZ ReferenceError on pre-aborted login signal.

### Acceptance criteria
- [ ] finish is declared before the aborted-signal check
- [ ] An already-aborted signal produces the correct abort error
- [ ] Cleanup timer is cleared on the pre-aborted path (no leak)

### DIFF
--- a/packages/runtime/src/providers/oauth/local-callback-server.ts
+++ b/packages/runtime/src/providers/oauth/local-callback-server.ts
@@ -120,18 +120,8 @@ export class LocalCallbackServer {
 
     return new Promise<AuthorizationResult>((resolve, reject) => {
       let settled = false;
-      const timer = setTimeout(() => {
-        finish(() => reject(new CallbackTimeoutError()));
-      }, options.timeoutMs);
 
       const onAbort = () => finish(() => reject(new OAuthError("callback_timeout", "Login aborted")));
-      if (options.signal) {
-        if (options.signal.aborted) {
-          onAbort();
-          return;
-        }
-        options.signal.addEventListener("abort", onAbort, { once: true });
-      }
 
       const finish = (action: () => void) => {
         if (settled) return;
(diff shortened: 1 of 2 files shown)


georgi (commented):
Reviewed — looks good to merge (cannot formally approve as it's the same author).

The fix correctly hoists `finish`, `timer`, and `onRequest` above the synchronous pre-aborted-signal check, eliminating the TDZ `ReferenceError`. All three acceptance criteria are met:

- `finish` is declared (line 126) before the aborted check (line 181); `timer`/`onRequest` are also initialized before `onAbort()` can fire.
- The pre-aborted path rejects with `OAuthError("Login aborted")` — verified by the new regression test.
- `onAbort` routes through `finish`, which clears the timer and removes the request listener, so no leak.

The regression test reproduces the original failure mode well. Nit (pre-existing, not introduced here): the abort still uses error code `"callback_timeout"` rather than a dedicated abort code — fine to leave.

Note: CI was still pending at review time.

Issue #12, opened 2024-12-20
Conda Environment is located under ~/.config


imbev:
nodetool currently prepares a Python environment under `~/.config/nodetool-electron/conda_env`.

This is a problem because `~/.config` is intended for per-user configuration, not application data. Following the XDG Base Directory standard, this environment should be located within the XDG_DATA_HOME directory instead, somewhere under `~/.local/share` .

georgi:
Good point. Do you think it's better to allow the user to select the location?

imbev:
Yes, with a provided default

georgi:
addressed in commit 00e8bfb5

heavy-d:
thanks @imbev 
will be in upcoming release

PR #3772, opened 2026-06-19
[T-20260619-0012] #12 IPv6-embedded IPv4 bypasses private-range check

georgi:
Done.

IPv6 transition addresses that embed an IPv4 — 6to4 (`2002::/16`) and NAT64 (`64:ff9b::/96`) — bypassed the SSRF private-range check because the IPv4 is buried inside the IPv6 literal rather than appearing as separate octets. I added a small IPv6 expansion helper, extracted the embedded IPv4 from both forms, and ran the existing `isPrivateIpv4` check against it.

**Changed:**
- `packages/runtime/src/providers/safe-url.ts` — new `expandIpv6()` (expands `::` compression and a trailing dotted-quad into 8 hextets), plus 6to4/NAT64 decode + range-check inside `isSafePublicHttpsUrl`. Existing IPv4-mapped logic left untouched.
- `packages/runtime/tests/providers/safe-url.test.ts` — tests covering both embedding forms (private rejected, public still accepted).

All 21 tests pass; typecheck and lint clean.

Note: this stays within the documented defense-in-depth scope — DNS rebinding to a public name resolving privately is still out of scope, as the file's header comment already calls out.

---

Closes task **T-20260619-0012**: #12 IPv6-embedded IPv4 bypasses private-range check.

### Acceptance criteria
- [ ] 6to4 (2002::) and NAT64 (64:ff9b::) embedded IPv4 are decoded and range-checked
- [ ] Embedded private/loopback IPv4 is rejected
- [ ] Test covers both embedding forms

### DIFF
--- a/packages/runtime/src/providers/safe-url.ts
+++ b/packages/runtime/src/providers/safe-url.ts
@@ -11,11 +11,54 @@
  * This is an allowlist gate: only https URLs to apparently-public hosts pass.
  * The check is conservative — unparseable URLs, non-https schemes, and literal
  * IP addresses in loopback / link-local / RFC1918 / CGNAT ranges (including
- * IPv4-mapped IPv6 literals) are refused. DNS rebinding can still resolve a
+ * IPv4-mapped IPv6 literals and IPv4 embedded in 6to4 / NAT64 transition
+ * addresses) are refused. DNS rebinding can still resolve a
  * public name to a private address, so this is defense-in-depth that
  * complements network-level egress filtering rather than a complete SSRF
  * mitigation. Redirects are validated per-hop by {@link safeFetch}.
  */
+
+/**
+ * Expand an IPv6 literal (already stripped of brackets, lower-cased) into its 8
+ * 16-bit hextets, handling `::` compression and a trailing dotted-quad. Returns
+ * `null` if `host` is not a well-formed IPv6 literal.
+ */
+function expandIpv6(host: string): number[] | null {
+  if (!host.includes(":")) return null;
+  // Rewrite a trailing dotted-quad (`...:a.b.c.d`) into two hex hextets so the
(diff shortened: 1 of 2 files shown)


georgi (commented):
**Approve** (posting as comment — GitHub blocks self-approval).

Correct, well-tested, and consistent with the existing SSRF guard. All three acceptance criteria are met and all 21 tests pass locally (typecheck/lint clean).

Verified the logic by tracing both embedding forms through WHATWG URL normalization:
- **6to4** (`2002::/16`): IPv4 in hextets[1..2], decoded as `(h1<<16)|h2` and range-checked — `2002:7f00:1::` → 127.0.0.1 rejected. ✓
- **NAT64** (`64:ff9b::/96`): the dotted-quad form is normalized by the URL parser to hex (`64:ff9b::127.0.0.1` → `64:ff9b::7f00:1`), and `expandIpv6` handles both; the `[2..5]`-zero prefix guard is correct. ✓
- Bit extraction uses `>>> 24`/`>>> 16`, so the negative-int result of `0xffff << 16` is handled correctly.
- `expandIpv6` only sets `embedded` for the two prefixes, so general public IPv6 is never falsely rejected (`2002:808:808::` and `64:ff9b::8.8.8.8` still pass). ✓
- Invalid octets (>255) hit the `\0` sentinel and fail the group parse safely.

Non-blocking follow-up (out of this task's scope): the file is explicitly defense-in-depth, but two other embedded-IPv4 vectors remain undecoded — **Teredo** (`2001::/32`, IPv4 in the low 32 bits XOR'd with `0xffffffff`) and the **NAT64 local-use prefix** (`64:ff9b:1::/48`, RFC 8215). ISATAP is already incidentally covered by the existing `fe[89ab]` link-local rule. Worth a tracking item if the threat model wants full transition-address coverage, but not required here.

PR #3773, opened 2026-06-19
[T-20260619-0009] #9 RunPod GPU pods provisioned with neither gpuCount nor vcpuCount

georgi:
Done.

## Summary

Fixed `deployWorkerPod` in `packages/compute/src/providers/runpod-rest.ts` so RunPod GPU pods are provisioned with explicit `gpuCount` and `vcpuCount` instead of relying on undocumented REST defaults — which risked a "running" pod allocated zero GPUs.

**Changes:**
- **`runpod-rest.ts`** — Added three exported, documented constants: `DEFAULT_GPU_COUNT` (1), `DEFAULT_GPU_VCPU_COUNT` (8), `DEFAULT_CPU_VCPU_COUNT` (4). `deployWorkerPod` now always sends an explicit `vcpuCount`, and sends `gpuCount` for GPU pods (omitted for CPU pods). Added a `gpuCount` option to `DeployWorkerPodOptions`.
- **`providers/types.ts`** — Added optional `gpuCount` to `WorkerSpec` so multi-GPU requests are reachable.
- **`providers/runpod.ts`** — Threaded `spec.gpuCount` through `RunpodPodProvider.provision`.
- **`__tests__/runpod-rest.test.ts`** — New test file (3 cases) covering GPU defaults, explicit overrides, and the CPU path.

All acceptance criteria met. Lint (`tsc --noEmit`) and the full compute suite (73 tests) pass.

---

Closes task **T-20260619-0009**: #9 RunPod GPU pods provisioned with neither gpuCount nor vcpuCount.

### Acceptance criteria
- [ ] deployWorkerPod sets an explicit gpuCount for GPU pods
- [ ] vcpuCount is set explicitly rather than left undefined
- [ ] Default values are documented in code

### DIFF
--- a/packages/compute/src/__tests__/runpod-rest.test.ts
+++ b/packages/compute/src/__tests__/runpod-rest.test.ts
@@ -1,0 +1,99 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  DEFAULT_CPU_VCPU_COUNT,
+  DEFAULT_GPU_COUNT,
+  DEFAULT_GPU_VCPU_COUNT,
+  deployWorkerPod,
+} from "../providers/runpod-rest.js";
+
+// ---------------------------------------------------------------------------
+// Mock the global fetch the RunPod REST transport uses.
+// ---------------------------------------------------------------------------
+
+const mockFetch = vi.fn();
+vi.stubGlobal("fetch", mockFetch);
+
+function jsonResponse(body: unknown, status = 200, ok = true): Response {
+  return {
+    ok,
+    status,
+    json: () => Promise.resolve(body),
+    text: () => Promise.resolve(JSON.stringify(body)),
+    headers: new Headers(),
+  } as unknown as Response;
+}
+
+beforeEach(() => {
+  mockFetch.mockReset();
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+const API_KEY = "test-runpod-key";
+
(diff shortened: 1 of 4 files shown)


georgi (commented):
✅ **Approve** (posting as comment — GitHub blocks self-approval). Correct, tested, and consistent with the codebase.

**Verified:**
- `deployWorkerPod` sends explicit `gpuCount` (default 1) for GPU pods only (conditional spread), omitted for CPU pods. ✓ (criterion 1)
- `vcpuCount` always pinned: 8 for GPU, 4 for CPU. CPU path preserves prior `?? 4` behavior; GPU path now also honors a caller-supplied value instead of forcing `undefined`. ✓ (criterion 2)
- Defaults are exported constants with doc comments explaining the zero-GPU-running-pod risk. ✓ (criterion 3)
- `RunpodPodSpec` already declares both fields and `createPod` forwards the spec straight to the POST body, so the new tests' `body.gpuCount`/`body.vcpuCount` assertions are valid.
- `gpuCount` threaded cleanly `WorkerSpec → RunpodPodProvider.provision`. No other GPU callers affected (the only other `deployWorkerPod` caller deploys a CPU pod).
- Ran the new tests (3 passed) and `tsc --noEmit` (clean) locally.

**Non-blocking note:** the prior code deliberately left `vcpuCount` undefined for GPU pods. Pinning it is exactly what the task requested, but worth a one-time confirmation that RunPod's REST v1 accepts (rather than ignores/rejects) an explicit `vcpuCount` on GPU pods, where vCPU is often derived from the GPU allocation.

PR #3774, opened 2026-06-19
[T-20260619-0015] #15 Codex provider: empty output_text per tool round; no terminal done on early SSE close

georgi:
Done. Both bugs in the Codex provider are fixed with test coverage.

## Summary

Fixed two issues in `packages/runtime/src/providers/codex-provider.ts`:

1. **Empty `output_text` on tool round-trips** — `buildRequestBody` pushed a `message` item with an empty `output_text` part for every assistant turn that had tool calls but no text (`content: null`). Now message items are skipped when there's no text; the tool call still goes out as a `function_call` item.

2. **No terminal `done` on early SSE close** — the streaming loop only emitted `done: true` on a `response.completed` event. If the SSE connection dropped early, consumers never saw the turn end. Now the generator tracks whether a terminal chunk was seen and synthesizes `{ type: "chunk", content: "", done: true }` after the read loop if none arrived. Abort/error paths throw before reaching it, so they're unaffected.

### Files changed
- `packages/runtime/src/providers/codex-provider.ts` — both fixes
- `packages/runtime/tests/providers/codex-provider.test.ts` — added a test for the early-closing SSE stream (asserts exactly one terminal chunk) and one verifying no empty message items are sent on a tool round-trip

Typecheck, lint, and all 9 provider tests pass.

---

Closes task **T-20260619-0015**: #15 Codex provider: empty output_text per tool round; no terminal done on early SSE close.

### Acceptance criteria
- [ ] No empty output_text message emitted on tool round-trips
- [ ] A terminal done chunk is emitted even when SSE closes early
- [ ] Test covers an early-closing SSE stream

### DIFF
--- a/packages/runtime/src/providers/codex-provider.ts
+++ b/packages/runtime/src/providers/codex-provider.ts
@@ -320,11 +320,17 @@ export class CodexProvider extends OpenAIProvider {
           });
         }
       }
+      // Skip empty message items: an assistant tool-call round carries its
+      // payload in the function_call items above and has no text of its own.
+      // Emitting an empty output_text on every round-trip both bloats the
+      // request and can confuse the backend.
+      const text = textOf(m.content);
+      if (!text) continue;
       const partType = m.role === "assistant" ? "output_text" : "input_text";
       input.push({
         type: "message",
         role: m.role,
-        content: [{ type: partType, text: textOf(m.content) }]
+        content: [{ type: partType, text }]
       });
     }
 
@@ -399,6 +405,7 @@ export class CodexProvider extends OpenAIProvider {
     const reader = res.body.getReader();
     const decoder = new TextDecoder();
     let buffer = "";
+    let sawTerminal = false;
 
     try {
       for (;;) {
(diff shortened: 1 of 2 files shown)


georgi (commented):
**Approve** (posting as comment — GitHub blocks self-approval). Both fixes are correct and well-targeted.

**Empty `output_text` skip** — `if (!text) continue;` sits after the `function_call` items are pushed, so a pure tool-call assistant turn still emits its `function_call` and only drops the empty `message`. Mixed text+tool turns keep both, same ordering as before.

**Terminal-done synthesis** — `done: true` is only ever produced by `response.completed`, so `sawTerminal` flips exactly once on the happy path (synth skipped) and fires once on early close. Abort/error paths throw from inside the loop and never reach the post-loop synth, as the PR claims. The `!("args" in item)` discriminant matches the one used in `generateMessage`.

Tests map cleanly to the acceptance criteria: the early-close test asserts exactly one terminal chunk (guards the per-event regression), and the round-trip test asserts a `function_call` exists with no empty `message` item.

Non-blocking: CI still pending (8 checks) — assuming green, good to merge.

PR #3775, opened 2026-06-19
[T-20260619-0014] #14 Surrogate-pair splitting in truncation helpers

georgi:
Done.

Fixed surrogate-pair splitting in both truncation helpers and added tests for emoji-boundary cuts.

- `packages/agents/src/constants.ts` — `truncateToolResult` now pulls the char cut back by one when it would land between the halves of a UTF-16 surrogate pair (added an `isHighSurrogate` helper; the kept-count in the notice now reflects the adjusted cut).
- `packages/runtime/src/token-counter.ts` — `truncateToTokens` decodes a prefix of BPE tokens, whose boundaries can fall mid-UTF-8, so it now strips any trailing partial character (U+FFFD replacement char or lone high surrogate) via a new `stripTrailingPartialChar` helper.
- Tests: `tool-result-truncation.test.ts` asserts no split surrogate at the cut and that a whole emoji is kept when it fits; `token-counter.test.ts` loops emoji truncation across many token budgets, asserting no trailing lone surrogate or replacement char.

Targeted tests pass and both packages typecheck clean. No caveats.

---

Closes task **T-20260619-0014**: #14 Surrogate-pair splitting in truncation helpers.

### Acceptance criteria
- [ ] truncateToolResult does not split surrogate pairs
- [ ] truncateToTokens does not split surrogate pairs
- [ ] Test truncates a string at an emoji boundary without producing a lone surrogate

### DIFF
--- a/packages/agents/src/constants.ts
+++ b/packages/agents/src/constants.ts
@@ -18,16 +18,30 @@ export const MAX_TOOL_RESULT_CHARS = 25_000;
  * Returns `text` unchanged when it already fits. Otherwise keeps the first
  * `maxChars` characters and appends a notice telling the model the output was
  * cut and how to fetch a smaller slice.
+ *
+ * The cut is pulled back by one if it would land between the two halves of a
+ * UTF-16 surrogate pair, so the kept prefix never ends in a lone surrogate
+ * (which would render as U+FFFD and corrupt a multi-byte character).
  */
 export function truncateToolResult(
   text: string,
   maxChars: number = MAX_TOOL_RESULT_CHARS
 ): string {
   if (text.length <= maxChars) return text;
-  const omitted = text.length - maxChars;
+  // Avoid splitting a surrogate pair: if the char just before the cut is a
+  // high surrogate, it pairs with the low surrogate at `maxChars`, so drop it.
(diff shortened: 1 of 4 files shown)


georgi (commented):
**APPROVE** (posted as comment — GitHub won't let me approve my own PR).

Correct, well-scoped fix for surrogate-pair splitting in both truncation helpers.

- `truncateToolResult`: pulling the cut back when `text[maxChars-1]` is a high surrogate correctly drops both halves of a split pair, and the notice's kept-count uses the adjusted `cut`. Edge cases (maxChars 0, `charCodeAt(-1)` → NaN) degrade safely.
- `truncateToTokens` / `stripTrailingPartialChar`: stripping trailing `U+FFFD` or lone high surrogates is sound — a well-formed emoji ends in a *low* surrogate, so a complete pair is never truncated; partial multi-byte UTF-8 from the tokenizer decodes to `U+FFFD`, which the loop removes.

Tests hit the exact emoji boundary and sweep many token budgets, covering all three acceptance criteria.

Nit (non-blocking): a result whose genuine content ends in `U+FFFD` would have that char stripped on the truncation path. Negligible.

Issue #13, opened 2024-12-21
Install percentage exceeds 100%


iwr-redmond:
![extra-progress](https://github.com/user-attachments/assets/2b343314-f20c-41ca-a9db-3b91141b77ff)

After the file download, the installer goes up to 200% as it prepares the environment. I reckon it's supposed to go back to zero first.

georgi:
addressed in 4dd8ac8

iwr-redmond:
Fixed in 0.6.0, where no progress bar is used.

PR #3776, opened 2026-06-19
[T-20260619-0013] #13 Code-runner containers still run as root (uid 0)

georgi:
Done. Code-runner Docker containers now run as a non-root user.

## What changed and why

Code-runner containers executed untrusted user code as uid 0. The prior hardening (cap-drop ALL + no-new-privileges) mitigated this, but setting a non-root `User` is the natural next defense-in-depth layer requested by the task.

- **`packages/code-runners/src/stream-runner-base.ts`** — added a `user` option to `StreamRunnerOptions` (default `"1000:1000"`, `null` to keep the image's default user) and applied it via Docker's top-level `User` field in `createContainer`.
- **`packages/code-runners/src/server-docker-runner.ts`** — exposed the same `user` knob on `ServerDockerRunnerOptions` and threaded it through; applied `User` in its own `createContainer` call.
- **Tests** — added coverage for the non-root default, custom user, and `null` passthrough in both runner test suites (96 tests pass).

All existing hardening (cap-drop, no-new-privileges, private IPC, optional read-only rootfs/workspace + noexec tmpfs) is untouched.

## Notes
- The `User` field is set at the container-config level (not `HostConfig`), which is where Docker expects it.
- `npx tsc --build` and the targeted lint/test runs pass. The repo-wide `npm run build` fails only because this worktree's root `node_modules/typescript` isn't installed — an environment issue, not these changes.
- **Caveat**: on native Linux, runners that *write* files into the bound `/workspace` may need that host dir made writable for uid 1000, since `getWorkspaceHostPath` creates it owned by the host process uid. Sample stdout-only execution (echo/print) is unaffected, and the runner subclasses do no runtime package installs, so none structurally require root.

---

Closes task **T-20260619-0013**: #13 Code-runner containers still run as root (uid 0).

### Acceptance criteria
- [ ] Code-runner containers run as a non-root user
- [ ] Existing cap-drop / no-new-privileges flags retained
- [ ] Sample code execution still works under the non-root user

### DIFF
--- a/packages/code-runners/src/server-docker-runner.ts
+++ b/packages/code-runners/src/server-docker-runner.ts
@@ -51,6 +51,8 @@ export interface ServerDockerRunnerOptions {
   readonlyRootfs?: StreamRunnerOptions["readonlyRootfs"];
   /** Mount the workspace bind read-only. Default false. */
   readonlyWorkspace?: StreamRunnerOptions["readonlyWorkspace"];
+  /** Non-root user the container runs as. Defaults to "1000:1000". */
+  user?: StreamRunnerOptions["user"];
 }
 
 // ---------------------------------------------------------------------------
@@ -149,7 +151,8 @@ export class ServerDockerRunner extends StreamRunnerBase {
       capDrop: options.capDrop,
       securityOpt: options.securityOpt,
       readonlyRootfs: options.readonlyRootfs,
-      readonlyWorkspace: options.readonlyWorkspace
+      readonlyWorkspace: options.readonlyWorkspace,
+      user: options.user
     };
     super(baseOptions);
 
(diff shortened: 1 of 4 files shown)


georgi (commented):
**Review: LGTM** (posting as comment — GitHub blocks self-approval).

Clean, focused defense-in-depth change.

**Coverage is complete** — both `createContainer` sites (`stream-runner-base.ts`, `server-docker-runner.ts`) set `User`, and all Docker runner subclasses (Python/Bash/Ruby/JS/Command/Lua) inherit the base `stream()` path, so no call site is missed.

**Correctness verified:**
- `this.user = options?.user === undefined ? "1000:1000" : options.user` correctly distinguishes the default from an explicit `null`, and the `null`-passthrough case is tested.
- `User: this.user ?? undefined` maps `null` → omit field → Docker image default. Correct.
- `User` set at container-config level (not `HostConfig`) — the right place.

Threading follows the existing `readonlyRootfs`/`readonlyWorkspace` pattern, so it's consistent with the file. Targeted suites pass (96/96 locally).

**One follow-up to validate** (acceptance criterion 3, only unit-tested here): the workspace-write case you flagged is real. On native Linux, `getWorkspaceHostPath` creates the rw bind mount owned by the host process uid, so uid 1000 inside the container can't write to it — runners that emit output files into `/workspace` could regress there. stdout-only execution is unaffected. Worth an integration check (or a host-dir chown/relaxed-perms step) before depending on it on Linux hosts; Docker Desktop's VM permission translation masks this on macOS/Windows.

PR #3778, opened 2026-06-19
[T-20260619-0002] #2 Bridge setTarget() race clobbers new worker target

georgi:
Fixed the `setTarget()` race in `WebsocketPythonBridge`. `setTarget` tore down via `_teardownSocket()`, which only touches `this._ws`, but an in-flight initial handshake's socket lives solely in the `_openTransport` closure. When that old-URL handshake resolved, its `finishSuccess` would set `this._ws` and clobber the socket opened against the new target — leaving the bridge silently attached to the OLD worker URL/token (the provision→attach→setTarget flow). The fix invokes `_abortHandshake` before retargeting, the same mechanism `close()` already uses to reject a stale handshake and crash-safely terminate its CONNECTING socket.

**Changed files:**
- `packages/runtime/src/python-websocket-bridge.ts` — `setTarget` now aborts any in-flight handshake before `_teardownSocket`/`_attemptReconnect`.
- `packages/runtime/tests/python-websocket-bridge.test-helpers.ts` — added `gateUpgrade` option + `releaseUpgrade()` (via `verifyClient`) to hold a WS upgrade in CONNECTING deterministically.
- `packages/runtime/tests/python-websocket-bridge.test.ts` — new test re-points mid-handshake, then releases the old upgrade and asserts the execute reaches the new worker and never the old one.

**Verification:** New test fails without the fix (old worker receives the execute) and passes with it. Full `python-websocket-bridge.test.ts` (26 tests) green; `tsc --noEmit` clean. All three acceptance criteria checked.

---

Closes task **T-20260619-0002**: #2 Bridge setTarget() race clobbers new worker target.

### Acceptance criteria
- [ ] setTarget aborts any in-flight initial handshake before connecting to the new target
- [ ] A late-resolving old-URL handshake can no longer overwrite this._ws
- [ ] Test: provision→attach→setTarget leaves bridge on the new URL/token

### DIFF
--- a/packages/runtime/src/python-websocket-bridge.ts
+++ b/packages/runtime/src/python-websocket-bridge.ts
@@ -206,6 +206,19 @@ export class WebsocketPythonBridge extends PythonBridgeBase {
     this._reconnecting = false;
     this._reconnectDelayMs = INITIAL_RECONNECT_DELAY_MS;
 
+    // Abort any in-flight handshake. A CONNECTING socket lives only in the
+    // _openTransport closure, NOT in this._ws, so _teardownSocket can't reach
+    // it — and its finishSuccess, if left to fire, would set this._ws and
+    // clobber the socket we're about to open against the new target (leaving
+    // the bridge attached to the OLD worker url/token). This is exactly the
+    // provision→attach→setTarget flow. Aborting rejects that handshake and
+    // crash-safely tears down its CONNECTING socket.
+    if (this._abortHandshake) {
+      const abort = this._abortHandshake;
+      this._abortHandshake = null;
+      abort(new Error("Python worker re-pointed via setTarget"));
+    }
+
     // Tear down the current socket (detaches listeners → no spurious
     // _onUnexpectedClose) and reject in-flight requests, then drive a fresh
     // connect against the new target via the reconnect path.
--- a/packages/runtime/tests/python-websocket-bridge.test-helpers.ts
+++ b/packages/runtime/tests/python-websocket-bridge.test-helpers.ts
@@ -48,12 +48,25 @@ export interface FakeWorkerOptions {
    *  - "error": emit an error-status progress frame then a terminal error frame
    */
   downloadMode?: "progress" | "error";
+  /**
+   * When true, the WebSocket upgrade is held until {@link
+   * FakeWorkerHandle.releaseUpgrade} is called. Lets a test keep a client's
+   * initial handshake in the CONNECTING state deterministically (e.g. to
+   * reproduce a late-resolving handshake racing setTarget).
+   */
+  gateUpgrade?: boolean;
 }
 
 export interface FakeWorkerHandle {
   port: number;
   /** Forcibly drop all currently-connected client sockets. */
   dropConnections: () => void;
+  /**
+   * Release a gated WebSocket upgrade (see {@link FakeWorkerOptions.gateUpgrade}).
+   * No-op when no upgrade is currently pending. Releases the most recent pending
+   * upgrade.
+   */
+  releaseUpgrade: () => void;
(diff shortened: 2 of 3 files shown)


georgi (commented):
## Review — Bridge `setTarget()` race fix

> Posting as a comment (GitHub blocks `request_changes` on your own PR). Treat the **Must fix** item as a requested change.

### Summary
The core fix is correct and matches the codebase's existing pattern. `setTarget` now invokes `_abortHandshake` before `_teardownSocket`/`_attemptReconnect`, exactly mirroring what `close()` already does. This closes the real race: an in-flight initial handshake's socket lives only in the `_openTransport` closure (not `this._ws`), so without the abort its late `finishSuccess` would set `this._ws` and silently re-attach the bridge to the old worker.

All three acceptance criteria are met:
- ✅ `setTarget` aborts the in-flight handshake before connecting to the new target (`python-websocket-bridge.ts:216-220`).
- ✅ A late-resolving old-URL handshake can no longer overwrite `this._ws` — `finishError` sets `settled=true` and crash-safely terminates the CONNECTING socket, so the later `open`/`finishSuccess` is a no-op.
- ✅ The new test reproduces provision→attach→setTarget and asserts the bridge ends on B (url, `Bearer b-token`, execute lands on B / never A). **Verified locally: full suite green (26/26), and I confirmed the test fails without the abort.**

The `gateUpgrade`/`releaseUpgrade` test harness (stash the `verifyClient` accept callback to hold the upgrade in CONNECTING) is a clean, deterministic way to reproduce the race.

### Must fix

**Latent type error in the test helper.** `gateUpgrade?: boolean` was added to `FakeWorkerOptions`, but the `opts: Required<FakeWorkerOptions>` object at `python-websocket-bridge.test-helpers.ts:119` doesn't include it:

```
tests/python-websocket-bridge.test-helpers.ts(119,9): error TS2741:
Property 'gateUpgrade' is missing in type '{...}' but required in type 'Required<FakeWorkerOptions>'.
```

This isn't caught by CI because the package `lint` script (`tsc --noEmit`) uses a `tsconfig.json` that only `include`s `src` — the `tests/` dir is never typechecked, and `vitest run` doesn't typecheck either. So the helper's types currently lie. `gateUpgrade` is read directly off `initialOptions` (line 102) and never via `opts`, so the cleanest fix is to drop it from the `Required` set rather than stuff a server-construction flag into the per-message options bag — e.g. type `opts` against the message-handling subset, or just add `gateUpgrade: initialOptions.gateUpgrade ?? false` to the literal if you want the minimal one-liner.

### Non-blocking observation

**`setTarget` during an *automatic reconnect* handshake (outside the tested path).** The test covers re-pointing during the *initial* `connect()`, where aborting cleanly rejects `connect()` and nothing reschedules. But if `setTarget` fires while an auto-reconnect's `_openTransport` is mid-handshake (`_reconnecting === true`), aborting it rejects that `_openTransport`, which drops the *old* `_attemptReconnect` into its `catch` — and that catch calls `_scheduleReconnect()`. Meanwhile `setTarget` has already kicked off a fresh `_attemptReconnect()` for B. Because `setTarget` set `_reconnecting = false` before aborting, both drivers see `_reconnecting === false` and you can end up with the in-flight B reconnect *plus* a scheduled timer that opens a second socket to B (duplicate `"reconnected"`, extra socket churn). It's benign (B is correct either way) and arguably pre-existing fragility, but the abort now makes it deterministic. Worth a guard (e.g. don't let an aborted handshake's catch reschedule) and/or a sibling test for the reconnect-in-flight case.

### Verdict
Logic is right and well-tested for the targeted flow. One real defect — the type error in the added helper — should be fixed before merge (trivial). The reconnect-in-flight note is optional.

Issue #11, opened 2024-12-20
Lack of user consent for automatic download


imbev:
Upon starting the application, nodetool begins a 4GB download without warning the user.

The user should be informed of the size of the environment and have the opportunity to accept/deny the download beforehand.

georgi:
Agreed. I think we will add a dialog to select the install path and inform about the required disk usage.

georgi:
addressed in 00e8bfb5

heavy-d:
thanks for reporting @imbev 
will be in upcoming release

PR #3781, opened 2026-06-19
[T-20260619-0011] #11 SSRF guard is FAL-only; safeFetch re-applies init on redirects

georgi:
Done. All three acceptance criteria are met and committed.

## Summary

Routed sibling providers' provider-supplied result-URL downloads through the existing `safeFetch` SSRF guard (the same one FAL got in d54dfbbf), and fixed a latent credential-leak in `safeFetch`'s redirect handling.

**Changed:**
- `packages/runtime/src/providers/safe-url.ts` — `safeFetch` now tracks origin per redirect hop and strips `Authorization`/`Cookie`/`Proxy-Authorization` headers and the request body whenever a redirect crosses origins, so credentials/body meant for the original host never leak to a redirect target.
- `replicate-provider.ts`, `meshy-provider.ts`, `rodin-provider.ts`, `kie-provider.ts` — their result-URL downloads (`fetch(target)` / `modelUrl` / `downloadUrl` / `resultUrls[0]`) now go through `safeFetch`.
- Tests — added cross-origin-strip + same-origin-preserve cases to `safe-url.test.ts`, a private-IP-blocked download test to `meshy-provider-hardening.test.ts`, and updated provider download mocks to a faithful `Response` shape (with `headers`).

**Note on HuggingFace:** the criterion lists it, but the HF provider has no provider-supplied URL download — generated bytes come back through the `@huggingface` SDK, and its only raw `fetch` targets the hardcoded `HF_API_BASE` constant (`https://huggingface.co/api/models`), which isn't an SSRF surface. So it's covered vacuously, no change needed.

I deliberately did **not** add a defensive `res.headers?.get` guard in `safeFetch` (real `fetch` Responses always carry headers — that'd be a defensive check on a trusted path); instead I corrected the incomplete test doubles, matching the FAL commit's approach.

Verification: all 1222 runtime provider tests pass; `tsc --noEmit` (lint) clean.

---

Closes task **T-20260619-0011**: #11 SSRF guard is FAL-only; safeFetch re-applies init on redirects.

### Acceptance criteria
- [ ] Replicate/Meshy/Rodin/Kie/HuggingFace provider-URL downloads go through the SSRF guard
- [ ] safeFetch does not forward auth headers/body across cross-origin redirects
- [ ] Tests cover a blocked private-IP fetch on at least one migrated provider

### DIFF
--- a/packages/runtime/src/providers/kie-provider.ts
+++ b/packages/runtime/src/providers/kie-provider.ts
@@ -12,6 +12,7 @@ import Anthropic from "@anthropic-ai/sdk";
 import OpenAI from "openai";
 import type { Chunk } from "@nodetool-ai/protocol";
 import { BaseProvider } from "./base-provider.js";
+import { safeFetch } from "./safe-url.js";
 import { createLogger } from "@nodetool-ai/config";
 import type {
   ImageModel,
@@ -200,7 +201,7 @@ async function downloadResultBytes(
   const resultData = JSON.parse(resultJsonStr) as Record<string, unknown>;
   const resultUrls = resultData.resultUrls as string[];
   if (!resultUrls?.length) throw new Error("No resultUrls in Kie resultJson");
-  const dlRes = await fetch(resultUrls[0]);
+  const dlRes = await safeFetch(resultUrls[0]);
   if (!dlRes.ok) {
     throw new Error(`Failed to download from ${resultUrls[0]}`);
   }
--- a/packages/runtime/src/providers/meshy-provider.ts
+++ b/packages/runtime/src/providers/meshy-provider.ts
@@ -14,6 +14,7 @@
  */
 
 import { BaseProvider } from "./base-provider.js";
+import { safeFetch } from "./safe-url.js";
 import { createLogger } from "@nodetool-ai/config";
 import type {
   ImageTo3DParams,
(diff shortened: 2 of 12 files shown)


georgi (commented):
**APPROVE** (posting as comment — GitHub blocks self-approval). Correct, tested, and consistent with the existing FAL hardening.

**Acceptance criteria**
1. ✅ Replicate/Meshy/Rodin/Kie provider-URL downloads now route through `safeFetch`. Verified the migrated call sites are the only provider-supplied-URL downloads — every other raw `fetch` in these providers targets a hardcoded API constant (`KIE_API_BASE`, `RODIN_API_BASE_URL`, etc.), not an SSRF surface. HuggingFace is vacuously covered: its only raw `fetch` hits the hardcoded `HF_API_BASE` (`huggingface-provider.ts:163`); generated bytes return via the SDK.
2. ✅ `safeFetch` strips `Authorization`/`Cookie`/`Proxy-Authorization` and the body when a redirect crosses origins. Per-hop origin tracking is right: it compares each hop against the previous origin (not the initial one), and once stripped, credentials don't reappear on a later same-origin hop. Same-origin redirects preserve creds; non-sensitive headers (`X-Trace`) are kept.
3. ✅ Private-IP block covered: `meshy-provider-hardening.test.ts` asserts a `169.254.169.254` model URL throws and the download fetch is never attempted.

**Verified locally:** ran the 7 affected test files — 151 passed.

**Notes (non-blocking):**
- Updating the test doubles to include `headers: new Headers()` rather than adding a defensive `res.headers?.get` guard is the right call — matches the FAL commit and keeps the trusted path clean.
- Cross-origin redirects preserve `method` while dropping the body (a POST would become a bodyless POST). Moot in practice since these are GET downloads, and it matches pre-existing redirect behavior. Not worth changing here.

PR #3782, opened 2026-06-19
[T-20260619-0005] #5 e9fafbb7 audio transforms still silently pass through undecoded input

georgi:
Done. Committed on the task branch.

## Summary

Commit e9fafbb7 claimed to route the AUDIO_NODES transforms through `decodeAudioToWav` and make them throw on undecodable input, but only `lib-audio-effects`/`lib-audio-dsp` were actually migrated. The listed transforms still called `parseWavBytes`/`tryDecodeWav` and silently returned the *undecoded* input on mp3/flac — a no-op that's indistinguishable from a working effect. I finished the migration for the transforms named in the task.

**What changed** (`packages/audio-nodes/src/nodes/audio.ts`):
- **Normalize, Slice, Trim, FadeIn, FadeOut** — swapped `audioBytesAsync` + `parseWavBytes`/`tryDecodeWav` (silent raw-byte fallback) for `requireAudioBytes` + `decodeAudioToWav`, matching the effects/DSP pattern. WAV parses directly; other formats decode via WebAudio; undecodable input throws.
- **AudioMixer** and the shared **`concatAudio()`** helper (used by **Concat** and **ConcatList**) — now decode every track/part via `decodeAudioToWav` (throws on undecodable) and throw a clear error on non-uniform sample-rate/channel layouts instead of the old meaningless byte-level fallback. Empty/unconnected dynamic inputs are still skipped, so multi-input nodes don't throw on default refs.
- Removed the now-unused `tryDecodeWav` import.

**Test added** (`packages/audio-nodes/tests/audio-transform-decode.test.ts`): pins both halves of the contract — each transform processes decoded WAV correctly, and corrupt/undecodable bytes make it throw rather than pass through. Full package suite: **75 passing**, lint clean.

**Caveat / follow-up:** the other transforms with the same silent-passthrough fallback — Overlay, RemoveSilence, MonoToStereo, StereoToMono, Reverse, Repeat — are outside this task's listed set and were left untouched. They'd be worth migrating for consistency in a follow-up. (Noted on the task.)

Note: the workspace-level `npm run build` fails in this worktree because the root `node_modules/typescript` isn't installed — a pre-existing env issue, not from this change. `npx tsc --build`, `npm run lint`, and `npm test` all pass within the package.

---

Closes task **T-20260619-0005**: #5 e9fafbb7 audio transforms still silently pass through undecoded input.

### Acceptance criteria
- [ ] Normalize/Trim/Slice/Fade/Concat/Mixer decode via decodeAudioToWav
- [ ] mp3/flac input that cannot be decoded throws instead of passing through silently
- [ ] Test: a transform on mp3 input either processes decoded audio or throws

### DIFF
--- a/packages/audio-nodes/src/nodes/audio.ts
+++ b/packages/audio-nodes/src/nodes/audio.ts
@@ -18,12 +18,13 @@ import {
   audioRefFromBytes,
   audioRefFromWav,
   concatBytes,
+  decodeAudioToWav,
   encodePcm16Wav,
   encodeWav,
   parseWavBytes,
   readWavHeader,
+  requireAudioBytes,
   toBytes,
-  tryDecodeWav,
   uriToPath,
   type WavData
 } from "../lib/audio-wav.js";
@@ -491,15 +492,25 @@ export class NormalizeAudioNode extends BaseNode {
   declare audio: any;
 
   async process(context?: ProcessingContext): Promise<Record<string, unknown>> {
-    const bytes = await audioBytesAsync(this.audio, context);
-    const wav = tryDecodeWav({ data: bytes });
-    if (!wav || wav.samples.length === 0) {
-      return { output: audioRefFromBytes(bytes) };
+    const bytes = await requireAudioBytes(this.audio, context);
+    const wav = await decodeAudioToWav(bytes);
+    if (wav.samples.length === 0) {
+      return {
+        output: audioRefFromWav(
+          encodeWav(wav.samples, wav.sampleRate, wav.numChannels)
+        )
+      };
     }
 
     let peak = 0;
     for (const sample of wav.samples) peak = Math.max(peak, Math.abs(sample));
-    if (peak === 0) return { output: audioRefFromBytes(bytes) };
+    if (peak === 0) {
+      return {
+        output: audioRefFromWav(
+          encodeWav(wav.samples, wav.sampleRate, wav.numChannels)
+        )
+      };
+    }
 
     const gain = 1 / peak;
     const normalized = new Float32Array(wav.samples.length);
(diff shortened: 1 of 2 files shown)


georgi (commented):
**LGTM (approve-equivalent — GitHub blocks self-approval).** The diff correctly completes the migration that commit e9fafbb7 claimed but didn't finish.

**Verified against source:**
- Normalize/Slice/Trim/FadeIn/FadeOut now use `requireAudioBytes` + `decodeAudioToWav`, matching the lib-audio-effects/dsp pattern. WAV parses directly; mp3/flac route through WebAudio; undecodable input throws instead of silently passing through.
- `concatAudio()` (Concat + ConcatList) and AudioMixer decode every part via `decodeAudioToWav`. Sample-space mixing/concat math is unchanged and correct for uniform `numChannels`.
- Multi-input nodes still filter empty buffers (`.filter(b => b.length > 0)`) before decoding, so unconnected dynamic inputs are skipped — single-input nodes fail loud via `requireAudioBytes`. Correct contract split.
- `tryDecodeWav` import removed; `parseWavBytes`/`WavData`/`concatBytes` remain used by out-of-scope transforms. No dangling imports.

All three acceptance criteria are met, and the test pins both halves of the contract (decoded-WAV processing + throw on undecodable bytes).

**Non-blocking notes:**
- Tests use corrupt random bytes for the throw path and WAV for the process path — no real mp3/flac fixture, so the "decodable mp3 → processed" branch isn't exercised end-to-end.
- The non-uniform sample-rate/channel-count *throw* path (a deliberate, documented behavior change replacing the old byte-level fallback) isn't covered by a test.
- Overlay/RemoveSilence/MonoToStereo/StereoToMono/Reverse/Repeat keep the same silent-passthrough fallback — correctly flagged as out-of-scope follow-up.

CI is still pending (8 checks) — let it go green before merging.

PR #3784, opened 2026-06-19
[T-20260619-0002] #2 Bridge setTarget() race clobbers new worker target

georgi:
Fixed `setTarget` in `python-websocket-bridge.ts` so it aborts any in-flight handshake before tearing down the current socket and starting a reconnect to the new target.

**What changed:**
- `packages/runtime/src/python-websocket-bridge.ts` — `setTarget()` now calls `_abortHandshake` (same pattern as `close()`) before `_teardownSocket`. The abort sets `settled = true` in the `_openTransport` closure, making any subsequent `finishSuccess` a no-op so the old socket can never overwrite `this._ws`.
- `packages/runtime/tests/python-websocket-bridge.test-helpers.ts` — Added `gateUpgrade: boolean` option to `FakeWorkerOptions` and `releaseUpgrade()` to `FakeWorkerHandle`. Uses `verifyClient`'s async callback form to hold the WS upgrade server-side until released, enabling deterministic reproduction of the race.
- `packages/runtime/tests/python-websocket-bridge.test.ts` — New test in the `setTarget` block: starts a bridge mid-handshake to gated worker A, calls `setTarget(workerB)`, releases A's gate, then asserts the bridge is connected to worker B and that worker A received no execute frames.

---

Closes task **T-20260619-0002**: #2 Bridge setTarget() race clobbers new worker target.

### Acceptance criteria
- [x] setTarget aborts any in-flight initial handshake before connecting to the new target
- [x] A late-resolving old-URL handshake can no longer overwrite this._ws
- [x] Test: provision→attach→setTarget leaves bridge on the new URL/token

### DIFF
--- a/packages/runtime/src/python-websocket-bridge.ts
+++ b/packages/runtime/src/python-websocket-bridge.ts
@@ -206,6 +206,16 @@ export class WebsocketPythonBridge extends PythonBridgeBase {
     this._reconnecting = false;
     this._reconnectDelayMs = INITIAL_RECONNECT_DELAY_MS;
 
+    // Abort any in-flight handshake (CONNECTING socket) before tearing down.
+    // _teardownSocket only acts on this._ws, but a handshake in progress lives
+    // only in the _openTransport closure — without aborting it, its finishSuccess
+    // would later set this._ws and clobber the new target's socket.
+    if (this._abortHandshake) {
+      const abort = this._abortHandshake;
+      this._abortHandshake = null;
+      abort(new Error("Python worker re-pointed via setTarget"));
+    }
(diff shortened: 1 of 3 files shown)


georgi (commented):
Code review: the race condition fix is correct and well-tested. The abort-before-teardown pattern already exists in `close()`, so extending it to `setTarget()` is consistent. The test deterministically reproduces the race using a gated upgrade mechanism.

Blocked on Quality Gate CI failure — likely the pre-existing BORDER_RADIUS mock issue. Should pass once #3827 lands and this is rebased.

---

PR #3788, opened 2026-06-20
Backend hardening: provider seams, lifecycle, and correctness fixes

georgi:
## Summary

Fifteen small, independent backend correctness and robustness improvements across `node-sdk`, `models`, `protocol`, `storage`, `kernel`, `runtime`, `websocket`, `cli`, and `code-runners`. Each commit is self-contained with tests.

### Correctness / bug fixes
- **`RunEvent.appendEvent` is now atomic** — prevents `seq` collisions under concurrent appends.
- **Code runners force-kill subprocesses** that ignore `SIGTERM`, so a stuck child can't leak.
- **Queued runs are cancelled on client disconnect** instead of running orphaned.
- **Strict equality** in node-sdk asset validation.

### Validation
- Validate `dynamic_outputs` type metadata in the workflow graph schema.
- Validate data-edge source handles in graph validation.

### Provider / lifecycle seams
- Add an explicit **capability declaration seam** to `BaseProvider` (consumed by find-model tool, models API/router).
- Add a **`close()` lifecycle** to `BaseProvider` for resource cleanup (implemented by Ollama provider).
- Add an **`AgentMemory.restore()`** durability seam.

### Resilience
- **Bounded retry with backoff** for S3 uploads.
- **Settle CLI WebSocket connect once** and drain waiters on post-connect error.
- Derive a **consistent `apiCode`** for raw `TRPCError` throws.

### Typing
- Type the Drizzle database union in `models/db.ts`; use Drizzle's official column API in base-model reflection.
- Type HuggingFace node properties instead of `declare X: any`.

## Test Plan
- [ ] CI: `npm run typecheck`, `npm run lint`, `npm run test`
- [ ] New/updated unit tests in `models`, `protocol`, `storage`, `kernel`, `runtime`, `websocket`, `cli` pass

### DIFF
--- a/packages/agents/src/tools/find-model-tool.ts
+++ b/packages/agents/src/tools/find-model-tool.ts
@@ -22,10 +22,7 @@ import type {
   TTSModel,
   VideoModel
 } from "@nodetool-ai/runtime";
-import {
-  RECOMMENDED_MODELS,
-  providerCapabilities
-} from "@nodetool-ai/runtime";
+import { RECOMMENDED_MODELS } from "@nodetool-ai/runtime";
 import type { RecommendedUnifiedModel } from "@nodetool-ai/runtime";
 import { Tool } from "./base-tool.js";
 
@@ -253,9 +250,9 @@ export class FindModelTool extends Tool {
     for (const [providerId, instance] of providerEntries) {
       let supports: boolean;
       try {
-        supports = providerCapabilities(instance).includes(
-          capability as ProviderCapability
-        );
+        supports = instance
+          .getCapabilities()
+          .includes(capability as ProviderCapability);
       } catch {
         continue;
       }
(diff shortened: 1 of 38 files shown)


georgi (commented):
⚠️ **Review: needs fixes before merge**

Good scope and intent — the S3 retry logic, WebSocket settling, provider capability seams, and graph validation improvements all look solid. However, three issues need fixing:

### 1. CI is failing
The Quality Gate (`npm run check`) job fails. This must pass before merge.

### 2. Postgres race condition in `RunEvent.appendEvent`
The SQLite path is truly atomic (synchronous transaction), but the Postgres path uses `SELECT MAX(seq)` inside a default-isolation transaction. Two concurrent transactions can read the same max seq, and there's no retry-on-conflict logic — the UNIQUE constraint will throw. The test only covers SQLite.

**Fix:** Use `INSERT ... SELECT MAX(seq)+1` as a single statement, or add retry-on-unique-constraint-violation logic. Add a Postgres-path test.

### 3. `any` casts violate strict-mode rules
- `tx: any` in both SQLite and Postgres transaction bodies in `run-event.ts`
- Multiple `as any` casts in storage tests
- `getDb()` casts `PostgresJsDatabase` to `BetterSQLite3Database`, hiding type errors if PG-specific paths diverge

The project enforces TypeScript strict mode with no `any`. Use proper Drizzle transaction types.

### Minor items (non-blocking):
- No test for SIGKILL escalation in `terminateChild`
- Sequential queue drain on disconnect (`awaits` each cancel) could be slow with many queued jobs — consider `Promise.allSettled`

---

PR #3915, opened 2026-06-24
Sketch: inline tab rename, generation header upgrades, and magic generation effect

georgi:
## Summary

Three related improvements to the sketch/workspace experience.

### Workspace tabs — inline rename for sketch / video / 3d
- Double-clicking a sketch, video (timeline), or 3d (model3d) tab title now begins **in-place editing**, reusing the existing workflow-tab rename input (Enter/blur commits, Escape cancels).
- Renames persist to the backing document **keyed by ref** — sketch & timeline via tRPC `update`, model3d via the asset store — and roll the tab title back on failure.
- Sketch only: the per-instance session `name` is kept in sync with the tab title so the content autosave (which writes `session.name`) can't silently revert a rename.

### Sketch generation header (`ConnectedModePromptBar`)
- Removed the duplicated document name (now shown in the tab title) and the read-only `W × H · sRGB` chip.
- Added **resolution** and **aspect ratio** controls matching the media composer (`MediaControlChip` + `MediaOptionMenu` + `MediaAspectRatioMenu`). Changing either resizes the artboard via `resolveImageSize` and feeds `width`/`height` into the generation binding; initial values are derived from the current canvas.
- Layout polish: more spacing between Save/Export, prompt input + all buttons unified to 34px, larger token-based gap between header items.

### Magic generation effect
- Generating layers (`binding.status` `queued`/`generating`) get an animated overlay: a flowing translucent multi-colour **wash + shimmer** (`MagicGenerationFill`), plus an animated border, aura glow, and twinkling sparkles on the canvas (`GeneratingLayerOverlay`, positioned in document space so it tracks pan/zoom).
- The same wash + shimmer also plays over the layer's **thumbnail in the layers panel**.
- All effects respect `prefers-reduced-motion`.

## Test plan
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` (sketch suites: `ConnectedModePromptBar`, `SketchLayersPanel`, `sketchCanvasPresentation`, `WorkspaceTabsStore`, `SketchSessionStore.autosave`, all Layer tests) — passing
- Manual: rename sketch/video/3d tabs; pick aspect/resolution and confirm the canvas reshapes and a generation fills it; run a text-to-image generation and confirm the wash plays on both the canvas layer and its panel thumbnail.

### DIFF
--- a/web/eslint.design.mjs
+++ b/web/eslint.design.mjs
@@ -65,6 +65,10 @@ export const designTokenIgnores = [
   // Decorative multi-hue @keyframes "running" status pulse — a rainbow animation,
   // not semantic chrome.
   "src/components/miniapps/StandaloneMiniApp.tsx",
+  // Decorative multi-hue @keyframes "generating" wash + shimmer + aura — a
+  // rainbow animation overlay, not semantic chrome.
+  "src/components/sketch/GeneratingLayerOverlay.tsx",
+  "src/components/sketch/MagicGenerationFill.tsx",
 ];
 
 const muiPrimitiveMessage =
--- a/web/src/components/sketch/GeneratingLayerOverlay.tsx
+++ b/web/src/components/sketch/GeneratingLayerOverlay.tsx
@@ -1,0 +1,176 @@
+/** @jsxImportSource @emotion/react */
+/**
+ * GeneratingLayerOverlay
+ *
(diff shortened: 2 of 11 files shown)


georgi:
@claude fix ci

PR #4004, opened 2026-06-30
fix(web): build workspace packages before vite build

georgi:
A standalone `vite build` in `web/` (e.g. the Cloudflare Pages deploy,
whose build command is just `npm run build`) failed while bundling the
in-browser runner Web Worker:

  Rolldown failed to resolve import "@nodetool-ai/core-nodes/nodes/constant"
  from "web/src/lib/workflow/browserRunnerCore.ts"

The worker imports the node packages (core-nodes, image-nodes, audio-nodes,
code-nodes) and their decorator deps (node-sdk, kernel) from their built
`dist/` — they have no `nodetool-dev` source condition because `@prop`
decorators must be compiled by tsc, not bundled from source. CI and the
release workflow run `build:packages` before building web, so those dists
exist; a standalone deploy build does not, and the worker bundle (which
can't externalize an unresolved import) hard-fails on the first one.

Add a `prebuild` guard that builds the workspace packages when their
`dist/` is missing and no-ops otherwise — so it fixes the standalone build
without redundant rebuilds or a nested turbo invocation in pipelines that
already built packages.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01ELQeh5i9isEUZrd4yLYRQJ


georgi (commented):
Good approach — the conditional prebuild guard that checks for sentinel `dist/` files is better than the unconditional `prebuild` already on main (from PR #3993).

**Merge conflict:** PR #3993 already added `"prebuild": "npm --prefix .. run build:packages"` to `web/package.json`. This PR was branched before that merge and adds its own `prebuild` line. Needs a rebase to resolve the conflict — keep this PR's conditional version which supersedes the unconditional one from #3993.

Also, only CodeQL ran (no build/test CI), so there's no signal confirming the script works in CI. After rebase, a full CI run would build confidence.

---

PR #4015, opened 2026-06-30
Centralize asset reference expansion and improve media URI resolution

georgi:
## Summary

Moves the `expandAssetReferences` function from `llm-nodes` to `runtime` as the canonical implementation, and refactors `ProcessingContext.resolveMessageMediaUris` to use it. This ensures consistent handling of inline `asset://<id>.<ext>` references across chat and workflow surfaces, and simplifies the browser-side asset drop flow by deferring URI resolution to the server.

## Key Changes

- **Centralize asset expansion**: Move `expandAssetReferences` from `packages/llm-nodes/src/nodes/agents.ts` to `packages/runtime/src/prompt-asset-refs.ts` alongside the shared `findAssetRefs` parser. Re-export from `llm-nodes` for backward compatibility.

- **Refactor media URI resolution**: Extract the image/audio/text resolution logic in `ProcessingContext.resolveMessageMediaUris` into a recursive `resolvePart` helper that:
  - Splits text-embedded `asset://<id>.<ext>` image/audio mentions into proper blocks via `expandAssetReferences`
  - Inlines text-document mentions (`.md`, `.txt`, `.csv`) as their decoded contents via `inlineTextAssetRefs`
  - Converts all non-data, non-http URIs to data URIs for provider consumption
  - Handles both chat-surface drops (user typed/dropped a reference) and workflow-node surfaces (agent node built the message) uniformly

- **Simplify browser asset drops**: Update `useDragAndDrop` and `useFileHandling` to:
  - Store the `asset://<id>.<ext>` reference in a new `assetUri` field on `DroppedFile`
  - Use the asset's preview URL (`thumb_url` or `get_url`) only for the composer thumbnail, not for the wire-form reference
  - Remove the `assetUrlToDataUri` fetch-and-encode step — the server now dereferences the URI before the provider call, avoiding megabyte-scale re-uploads in the browser

- **Add server-side expansion**: Call `resolveMessageMediaUris` in `UnifiedWebSocketRunner` before provider invocation to ensure all asset references are expanded and dereferenced.

- **Export and test**: Add `expandAssetReferences` to `packages/runtime/src/index.ts` exports. Add unit tests for the new function and integration test for the refactored `resolveMessageMediaUris` flow.

## Implementation Details

- The `resolvePart` helper is async and returns `MessageContent[]` to support recursive expansion: a text part with embedded `asset://` references is split, each sub-part is processed (text documents inlined, media blocks resolved), and the results are flattened back into the message.
- MIME type mappings for image and audio are defined once at the top of `resolveMessageMediaUris` and reused across all parts.
- Video and unknown asset types remain as literal text so the model still sees the mention rather than a mislabeled block.
- The browser no longer fetches asset bytes; it sends the compact `asset://` reference and lets the server handle dereferencing, reducing bandwidth and latency for large media.

https://claude.ai/code/session_01M9SDtGk6BYZEmT5L6RvfRC


georgi:
Reviewed — architecture is sound. Centralizing `expandAssetReferences` in `runtime`, simplifying browser-side asset handling, and adding server-side expansion in `UnifiedWebSocketRunner` are all good changes. CI `test-packages` failure is pre-existing on main (`data-nodes`).

**One concern — `isResolvableUri` narrows URI scheme support:**

The old code resolved any URI that wasn't `data:` or `http(s):`. The new `isResolvableUri` restricts to only `asset://` and `/api/storage/`. The underlying `retrieveMediaBytes` also supports `file://`, `memory://`, and `s3://` via the storage adapter. While unlikely to appear in message content in practice, this is an undocumented behavior change that could silently skip resolution.

Suggestion: either widen the filter to match old behavior, or document the intentional restriction.

---

PR #4080, opened 2026-07-02
[T-20260629-0002] Guard build host against NODE_ENV=production / wrong Node

georgi:
All acceptance criteria are met and committed.

## Summary

Added a fast, network-free build preflight that fails loudly on the two environment misconfigurations that previously produced cryptic errors: `NODE_ENV=production` (which makes npm silently drop all devDeps → misleading "turbo: not found" / "module not found") and a wrong Node major (Node 24 vs the pinned 22.22.1).

**Changed:**
- `scripts/preflight.mjs` (new) — `checkNodeEnv` hard-errors (exit 1) on `NODE_ENV=production` with the exact fix (`unset NODE_ENV` / `npm install --include=dev`); `checkNodeVersion` reads `.nvmrc` and warns on a Node-major mismatch with an `nvm use` hint. Runs in ~28ms, no network, silent on the happy path. Exports are individually testable.
- `package.json` — wired preflight as `prebuild:packages`, `prebuild:packages:clean`, and `prebuild:packages:tsc` npm pre-hooks (fires once before the whole build, not per-package), plus a standalone `preflight` script.

**Decisions (see task notes):**
- Node-major mismatch is a **warning**, not a hard error — this worktree and likely CI run on Node 24, so erroring would break legitimate builds. `NODE_ENV=production` is the hard error since it's the real breakage.
- **Skipped `engine-strict=true`** in `.npmrc`: with `engines <23` it would make `npm install` fail outright on any Node 24 host (CI included), which is worse than what it guards.

Verified: happy path silent (matching major), production → immediate exit 1 before turbo runs, version warn on Node 24, ~28ms timing, oxlint clean.

---

Closes task **T-20260629-0002**: Guard build host against NODE_ENV=production / wrong Node.

### Acceptance criteria
- [x] Running a package build with NODE_ENV=production fails immediately with a clear message naming the cause and the fix (unset NODE_ENV / --include=dev)
- [x] Building on a Node major other than the .nvmrc pin produces a clear warning or error with a `nvm use` hint
- [x] Preflight adds <100ms and makes no network calls
- [x] Normal build on Node 22 + unset NODE_ENV is unaffected (no new output noise on the happy path beyond a one-line OK at most)
- [x] lint + typecheck pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,12 @@
   "scripts": {
     "postinstall": "node -e \"try{require('fs').unlinkSync('node_modules/.electron-native-rebuild-stamp')}catch{}\"",
     "create:pack": "node scripts/create-pack.mjs",
+    "preflight": "node scripts/preflight.mjs",
+    "prebuild:packages": "node scripts/preflight.mjs",
     "build:packages": "turbo run build --filter=\"./packages/*\"",
+    "prebuild:packages:clean": "node scripts/preflight.mjs",
     "build:packages:clean": "turbo run build --filter=\"./packages/*\" --force",
+    "prebuild:packages:tsc": "node scripts/preflight.mjs",
     "build:packages:tsc": "tsc --build tsconfig.build.json && node scripts/copy-manifests.mjs",
     "build:web": "turbo run build --filter=web",
     "build": "turbo run build",
--- a/scripts/preflight.mjs
+++ b/scripts/preflight.mjs
@@ -1,0 +1,81 @@
+import { readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// Fast, network-free guard for the package build environment. Two failure
+// modes have historically produced cryptic errors that look like missing
+// dependencies:
+//
+//   1. NODE_ENV=production makes `npm install`/`npm ci` silently omit ALL
+//      devDependencies (typescript, turbo, node-gyp). The build then dies with
+//      "turbo: not found" / "Cannot find module 'typescript'" and sends you
+//      hunting for a dependency that was never missing.
+//   2. Running on a Node major other than the `.nvmrc` pin (e.g. Node 24 vs
+//      the pinned 22) drifts away from the version the packaged app embeds.
+//
+// This runs before `build:packages` so both fail loudly with the actual fix
+// instead of a downstream red herring.
+
+const scriptDir = dirname(fileURLToPath(import.meta.url));
+const repoRoot = resolve(scriptDir, "..");
+
+function fail(message) {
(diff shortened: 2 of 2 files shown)


georgi (commented):
## Review: LGTM ✅ (posting as comment — GitHub blocks self-approval)

All 5 acceptance criteria verified empirically on this worktree:

| Criterion | Result |
|---|---|
| `NODE_ENV=production` fails immediately, names cause + fix | ✅ exit 1, clear message with `unset NODE_ENV` / `--include=dev` |
| Node-major mismatch warns with `nvm use` hint | ✅ warns on Node 24 vs pinned 22.22.1, exit 0 |
| Preflight <100ms, no network | ✅ ~27ms, pure fs reads |
| Happy path (Node 22 + unset NODE_ENV) silent | ✅ zero output on matching major |
| lint + typecheck pass | ✅ oxlint clean (exit 0), CI 8/8 green so far, 0 failures |

### What I checked beyond the criteria

- **No false-positive on the Docker build.** The one place both `NODE_ENV=production` and `npm run build:packages` appear is the `Dockerfile` — but they're in separate stages: the build stage (`node:22-slim`, line 37) never sets `NODE_ENV`, and `NODE_ENV=production` only lands in the `runtime` stage (line 95). So the hard-error won't block image builds, and Node 22-slim means no warning noise either.
- **CI paths are covered and safe.** `quality-guard`, `release`, `publish`, `mutation-testing`, `screenshots`, and `check` all invoke `npm run build:packages` (so the `prebuild:*` hook fires) on `.nvmrc`-pinned Node 22 with no `NODE_ENV` set — no new failures introduced.
- **Ordering is correct.** `checkNodeEnv` exits before `checkNodeVersion`, so a misconfigured host sees exactly one actionable message, not both.
- **`engine-strict` correctly omitted** from `.npmrc` — the task note explains it would break `npm install` on any Node 24 host; agreed, that would be worse than the problem.

### Non-blocking notes (no change required)

1. **Coverage gap in bypass paths.** The `prebuild:*` npm hooks only fire for the exact script names (`build:packages`, `:clean`, `:tsc`). `turbo:build:packages` (package.json:83) and `npm run build` (`turbo run build`, which also compiles packages) skip the guard. The primary/documented paths are covered, so this is fine as-is — just flagging it if you want the guard to be exhaustive later.
2. **No unit test** despite the exports being written to be testable. Consistent with the repo — `scripts/` has no existing test convention — so not a gap against conventions.
3. The hard-error fires even in the (rare) case where `NODE_ENV=production` is set but devDeps happen to be fully installed. That's exactly the "fail immediately" behavior the task asked for and the message is actionable, so intended.

Clean, focused, correct, and consistent with the codebase. 🚀

PR #4081, opened 2026-07-02
[T-20260629-0001] deploy.sh: restore last-deploy config as defaults

georgi:
Committed. Here's the summary for the PR.

## Restore last-deploy config as defaults in `deploy.sh`

A bare `./deploy.sh` defaulted to `PORT=443` + TLS `certs`, but prod actually runs `127.0.0.1:7777` with `--no-tls`. The correct config was already persisted to `.deploy/ports.env` after every deploy — the script just never read it back, so a redeploy broke the running config or failed the cert preflight. This makes `deploy.sh` restore that recorded config as its defaults.

### What changed (`deploy.sh`)
- **New restore block** after flag parsing: reads `PORT` / `TLS_MODE` / `BIND` from `.deploy/ports${SLUG:+-<slug>}.env` and applies them as defaults. The file is **parsed by key** (`while … IFS='='`), not `source`-d, so its `PORT`/`TLS_MODE` lines can't silently clobber CLI/env choices (both names collide with the script's own vars).
- **Precedence enforced: CLI flag > env var > recorded file > built-in default.** `TLS_MODE` now starts empty (CLI-only), `BIND` uses a `BIND_FROM_CLI` sentinel to tell `--bind` apart from env/file, and `PORT` starts empty (env-only) then falls through file → slug-derived → `443`.
- The `--slug` port-derivation condition changed from `-z NODETOOL_PORT` to `-z PORT`, so a recorded preview port pre-empts the hash-derived one.
- **`BIND` is now written** to the ports file (previously only `PORT`/`TLS_MODE`/`CONTAINER`), so loopback binding survives a redeploy. Write block wrapped in `{ … }` to stay shellcheck-clean.
- Prints e.g. `Restored config from .deploy/ports.env: PORT=7777 TLS=none BIND=127.0.0.1` when a file is found.

### Verification (no live deploy run)
- **8-case precedence dry-run harness** covering: first deploy, prod restore, env-PORT override, `--no-tls` override, `--bind`/env-BIND override, slug preview, and slug restore — all resolved as expected.
- `shellcheck` v0.11.0: **clean**. `bash -n`: OK.
- The restore-print line was verified against the spec format.

**Caveat:** docker/live deploy path was not exercised (no runtime here) — only the config-resolution and file-write logic was tested in isolation.

---

Closes task **T-20260629-0001**: deploy.sh: restore last-deploy config as defaults.

### Acceptance criteria
- [x] `./deploy.sh` with no args on this host reproduces the running prod config (127.0.0.1:7777, no TLS) instead of port 443 + certs
- [x] Explicit flags/env still override the restored values (verify: `NODETOOL_PORT=8080 ./deploy.sh` uses 8080)
- [x] First-deploy path (no .deploy/ports.env) still works with built-in defaults
- [x] BIND is persisted to .deploy/ports.env and restored on next run
- [x] Script prints which config it restored; shellcheck clean; dry-run documented in PR

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -44,8 +44,8 @@ IMAGE_TAG_PREV="rollback"
 PULL_TAG="${NODETOOL_IMAGE_TAG:-}"   # empty → derive from git HEAD
 CONTAINER_NAME="nodetool-server"
 SLUG="${NODETOOL_SLUG:-}"   # empty = production container; non-empty = preview instance
-PORT="${NODETOOL_PORT:-443}"
-BIND_ADDR="${NODETOOL_BIND:-}"   # empty = bind 0.0.0.0; set to 127.0.0.1 for loopback-only (e.g. behind a Cloudflare Tunnel)
+PORT="${NODETOOL_PORT:-}"        # env override; else restored from file / slug-derived / 443 below
+BIND_ADDR="${NODETOOL_BIND:-}"   # env override; empty = bind 0.0.0.0 (set 127.0.0.1 for loopback-only, e.g. behind a Cloudflare Tunnel). Restored from file below if unset.
 HEALTH_TIMEOUT=60
 HEALTH_INTERVAL=2
 # How long to wait for the CI-built image to appear in the registry
@@ -55,7 +55,8 @@ IMAGE_WAIT_INTERVAL=15
 # Parse flags
 NO_PULL=0
 NO_WAIT=0
-TLS_MODE="certs"  # none | self-signed | certs (default: certs for Cloudflare Full Strict)
+TLS_MODE=""       # set by --self-signed/--certs/--no-tls; else restored from file / defaults to "certs" below
+BIND_FROM_CLI=0   # 1 once --bind is seen, so a restored/env BIND can't clobber it
 ACTION=""  # deferred so --logs/--status/--stop/--rollback can honor --slug regardless of arg order
 
 while (( $# )); do
@@ -79,6 +80,7 @@ while (( $# )); do
     --bind)
       shift
       BIND_ADDR="${1:?--bind requires an address (e.g. 127.0.0.1 or 0.0.0.0)}"
+      BIND_FROM_CLI=1
       ;;
     --slug)
       shift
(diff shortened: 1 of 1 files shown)


georgi (commented):
## Review: deploy.sh restore last-deploy config — LGTM ✅ (posting as comment; GitHub blocks self-approval)

Read the full diff and surrounding script, then exercised the actual config-resolution block against 8 scenarios. Every acceptance criterion holds.

### Correctness — verified empirically
Extracted the real restore/precedence logic and ran it:

| Scenario | Result |
|---|---|
| First deploy (no file) | `PORT=443 TLS=certs BIND=<empty>` ✅ |
| Prod restore (7777/none/127.0.0.1) | `PORT=7777 TLS=none BIND=127.0.0.1` ✅ |
| `NODETOOL_PORT=8080` over file | `PORT=8080` ✅ |
| `--certs` over file `none` | `TLS=certs` ✅ |
| `--bind 0.0.0.0` over file bind | `BIND=0.0.0.0` ✅ |
| `NODETOOL_BIND=10.0.0.5` over file | `BIND=10.0.0.5` ✅ |
| slug, no file → hash-derived | `PORT=8969` ✅ |
| slug, file port pre-empts hash | `PORT=8888` ✅ |

Precedence (CLI > env > file > built-in) is enforced cleanly. Design choices are sound:
- Parsing the file by key instead of `source`-ing it correctly avoids `PORT`/`TLS_MODE` clobbering CLI/env — the central trap this task was about.
- The `BIND_FROM_CLI` sentinel is the right way to tell `--bind 0.0.0.0` apart from an unset bind.
- Switching the slug port guard from `-z NODETOOL_PORT` to `-z PORT` so a recorded preview port pre-empts the hash is consistent with the new restore semantics.
- `SCRIPT_DIR` is defined (L39) before the restore block, and `PORTS_FILE` is resolved once from the same `SLUG` used at write time — read/write paths stay in sync.

`bash -n` passes. Couldn't run shellcheck here (not installed in this env), but SC2129 was addressed via the `{ … } > "$PORTS_FILE"` block and the PR documents a clean run on v0.11.0.

### Minor, non-blocking notes
- The `Restored config from …` line prints the **raw file** values even when a CLI flag/env var will override them (e.g. `NODETOOL_PORT=8080` with recorded `7777` still prints `PORT=7777`). Defensible ("what's in the file"), but an operator could misread it as the effective config. Consider printing resolved values or moving the print after resolution.
- The restore block + print line also run for `--logs`/`--status`/`--stop`/`--rollback`, which don't use `PORT`/`TLS`. Harmless, slightly noisy.
- Once `BIND=127.0.0.1` is recorded there's no env-only way back to `0.0.0.0` (`NODETOOL_BIND=""` == unset); `--bind 0.0.0.0` is the escape hatch. Matches the documented model.

None of these block merge. Focused change (one file, +59/−10), correct, consistent with the script's style. **Approving in spirit** — please get CI green (was pending at review time) before merge.

georgi (commented):
## Review — LGTM ✅ (posting as comment; GitHub blocks self-approval)

Clean, well-scoped change to `deploy.sh`. All five acceptance criteria are met, and I verified the config-resolution logic empirically against the real script (extracted lines 39–168, ran with a `.deploy/ports.env` fixture):

| Scenario | Expected | Result |
|---|---|---|
| bare `./deploy.sh` w/ recorded file | 7777 / none / 127.0.0.1 | ✅ PORT=7777 TLS=none BIND=127.0.0.1 |
| `NODETOOL_PORT=8080 ./deploy.sh` | env wins → 8080 | ✅ PORT=8080 (TLS/BIND still restored) |
| no file (first deploy) | 443 / certs / empty | ✅ built-in defaults |

Also confirmed: `bash -n` passes; CI green (16 success, 1 neutral).

### What's good
- **Parse-by-key instead of `source`** is the right call — `PORT`/`TLS_MODE` in the file would otherwise clobber CLI/env vars of the same name. Correctly reasoned in the comment.
- Precedence (CLI > env > file > built-in) holds up for all three vars. The `BIND_FROM_CLI` sentinel is the correct way to tell an explicit `--bind` apart from "unset", since env can't express empty-on-purpose.
- Moving the `443` default *after* the slug block, and switching the slug derivation guard from `-z NODETOOL_PORT` to `-z PORT`, correctly lets a recorded preview port pre-empt the hash-derived one.
- `BIND` now persisted so loopback binding survives redeploys; write block wrapped in `{ … }` (SC2129-clean).

### Non-blocking nit
The restore block (and its `Restored config from …` print) runs before the deferred-action dispatch, so read-only actions — `--status`, `--logs`, `--stop`, `--rollback` — now also emit the "Restored config…" line. Harmless (nothing is written on those paths), just extra noise on what used to be quiet queries. Could gate the print on an actual deploy if you want them silent, but not worth blocking over.

No shell test harness exists in this repo, so the documented dry-run harness is the appropriate level of verification here.

Ships as-is. 🚀

PR #4085, opened 2026-07-02
[T-20260629-0004] Add one-shot "redeploy current main" entrypoint

georgi:
Committed. Here's the summary for the PR.

## Add one-shot `npm run redeploy` for current main

Redeploying on current main previously meant reconstructing the invocation by inspecting the live container and `.deploy/ports.env`. This adds a single command that does the whole flow and self-verifies.

**New:** `scripts/redeploy.sh`, wired as `npm run redeploy` (there's no Makefile in the repo, so the npm script is the entrypoint the task's `make redeploy` describes).

Flow:
1. **Dirty-tree guard** — refuses if there are uncommitted changes to *tracked* files (`git status --porcelain --untracked-files=no`, so expected deploy artifacts like `.env`, `cert.pem`, `web/dist` don't block). `--allow-dirty` downgrades to a warning.
2. `git fetch origin` → `git checkout main` → `git pull --ff-only origin main`.
3. `npm run build:web` (the bind-mounted frontend).
4. `./deploy.sh` bare — restores the prod config (PORT/TLS/BIND) from `.deploy/ports.env` per T-20260629-0001. **Pull decision:** if the local `nodetool-ai/nodetool:latest` image's baked `GIT_COMMIT_HASH` already equals `git rev-parse --short=7 HEAD`, it passes `--no-pull`; otherwise deploy.sh pulls/waits for the CI image for HEAD. `--pull` forces a pull.
5. **Verification** — reads `CONTAINER`/`PORT`/`TLS_MODE` back from `.deploy/ports.env`, curls `/health`, and compares the running container's `GIT_COMMIT_HASH` to HEAD. Prints `PASS`/`FAIL` for both and exits non-zero if either fails.

Idempotent: a second run finds main up to date, the image at HEAD (→ `--no-pull`), rebuilds `web/dist`, and re-runs the rolling deploy, converging on the same healthy container.

**Files changed:**
- `scripts/redeploy.sh` (new)
- `package.json` — `"redeploy": "bash scripts/redeploy.sh"`
- `deploy.sh` — header note pointing to `npm run redeploy` as the preferred wrapper
- `CLAUDE.md` — Common Pitfalls entry documenting the command

**Caveats:** I couldn't exercise the full end-to-end flow in this worktree (running it would switch the worktree to `main`, and there's no Docker/registry here). I verified it in isolation: syntax (`bash -n`), `--help` output, empty-array expansion under `set -u`, the `GIT_COMMIT_HASH` sed parse, and the dirty-tree guard semantics (untracked ignored, tracked change refuses) in a scratch git repo. First real run on the deploy host is the true smoke test.

---

Closes task **T-20260629-0004**: Add one-shot "redeploy current main" entrypoint.

### Acceptance criteria
- [x] A single command (`make redeploy` / `npm run redeploy`) fetches main, builds web/dist, and rolling-deploys with the restored prod config
- [x] Refuses or warns on a dirty working tree
- [x] Ends by printing container health + image-revision-vs-HEAD match (pass/fail)
- [x] Idempotent: running it twice in a row is safe and converges
- [x] Documented in deploy.sh header and CLAUDE.md
- [x] Depends on T-20260629-0001 (config restore) being merged first


georgi (commented):
## 🔴 Must-fix: `deploy.sh` is invoked from the wrong path — the deploy step will fail

`scripts/redeploy.sh:117`:

```sh
"$SCRIPT_DIR/deploy.sh" "${NO_PULL_FLAG[@]}" "${DEPLOY_ARGS[@]}"
```

`SCRIPT_DIR` resolves to the directory holding `redeploy.sh`, i.e. `<repo>/scripts` (line 35). But `deploy.sh` lives at the **repo root**, not in `scripts/`:

```
$ ls scripts/deploy.sh
ls: cannot access 'scripts/deploy.sh': No such file or directory
$ ls deploy.sh
deploy.sh
```

So step 4 — the actual rolling deploy, the heart of the whole command — dies with `No such file or directory`. This breaks acceptance criterion #1 ("a single command … rolling-deploys"). The script already `cd "$REPO_ROOT"` at line 58, so the fix is one of:

```sh
"$REPO_ROOT/deploy.sh" "${NO_PULL_FLAG[@]}" "${DEPLOY_ARGS[@]}"
# or, since we're already in REPO_ROOT:
./deploy.sh "${NO_PULL_FLAG[@]}" "${DEPLOY_ARGS[@]}"
```

The script's own header and CLAUDE.md both correctly say `./deploy.sh` — only the invocation itself points at `scripts/`.

This is exactly the gap the PR's own caveat flagged: `bash -n` and `--help` were exercised, but the full flow wasn't — and `bash -n` can't catch a bad runtime path, while `--help` exits before reaching line 117. Worth a real end-to-end run (or at least a `[[ -x "$REPO_ROOT/deploy.sh" ]]` sanity assert) before merge.

## Everything else checks out

Verified against the actual `deploy.sh`:
- **ports.env contract** — `deploy.sh` writes `PORT`/`TLS_MODE`/`BIND`/`CONTAINER` to `.deploy/ports.env` (deploy.sh:427-433) and restores from it on a bare invocation (deploy.sh:120-135). redeploy's read loop (lines 123-133) matches the format and default path.
- **Pull decision** — local tag `nodetool-ai/nodetool:latest` matches deploy.sh's `IMAGE_NAME:IMAGE_TAG` (deploy.sh:55-56); `--no-pull` is a real flag (deploy.sh:85). Short-SHA comparison against the baked `GIT_COMMIT_HASH` is sound.
- **Container default** `nodetool-server` matches deploy.sh:59.
- **Dirty-tree guard** (`--untracked-files=no`) correctly lets deploy artifacts through while blocking tracked changes; `--allow-dirty` downgrade is reasonable.
- **Docs** present in both `deploy.sh` header and CLAUDE.md, and read cleanly.
- Empty-array expansion under `set -u` is safe (bash 4.4+).

Fix line 117 and this is good to go. Marking as comment only because GitHub won't let me `request_changes` on this account's own PR — but treat the path bug as blocking.

georgi (commented):
## Review: LGTM ✅ (posting as comment — GitHub blocks self-approval)

Verified `scripts/redeploy.sh` against the actual `deploy.sh`, `Dockerfile`, and `.github/workflows/docker.yml`. The interfaces all line up:

- `deploy.sh` accepts `--no-pull` (deploy.sh:85). ✓
- `deploy.sh` records `PORT`/`TLS_MODE`/`BIND`/`CONTAINER` to `.deploy/ports.env` (deploy.sh:428-433) — redeploy reads exactly those keys back for verification. ✓
- `GIT_COMMIT_HASH` is baked as an `ENV` in the image (Dockerfile:66-68) and set to the 7-char short SHA (`${GITHUB_SHA::7}`, docker.yml:64), so the `== HEAD` comparison against `git rev-parse --short=7 HEAD` is exact. ✓
- Dirty-tree guard, `pull --ff-only`, `build:web`, and PASS/FAIL health + revision checks all match the criteria. `bash -n` passes.

All six acceptance criteria are met, and the doc updates (deploy.sh header + CLAUDE.md pitfall) are accurate.

### Non-blocking notes
1. **Slug + verification mismatch.** Verification always reads `$REPO_ROOT/.deploy/ports.env`, but `deploy.sh` writes to `.deploy/ports${SLUG:+-<slug>}.env`. If someone forwards `-- --slug foo`, deploy succeeds but verification reads the wrong (or default) `CONTAINER`/`PORT` and prints a spurious `FAIL` + non-zero exit. The primary "redeploy current main" path is unaffected; worth a comment or a slug-aware ports path if preview redeploys via this wrapper ever matter.
2. **Empty-array expansion under `set -u`** (`"${NO_PULL_FLAG[@]}" "${DEPLOY_ARGS[@]}"`) is safe on the Linux/bash 4.4+ deploy host but would error on bash 3.2 (stock macOS). Fine for the intended environment; noting for completeness.
3. Minor: `"redeploy"` sits between `build:web` and `build` in package.json scripts — harmless.

The author flagged that the full end-to-end flow couldn't be exercised in the worktree (no Docker/registry, and running it switches to `main`); first run on the deploy host is the real smoke test. Given the static verification above, that's a reasonable caveat.

**Recommendation: approve & merge** once CI passes (currently pending). None of the notes are blocking.

PR #4091, opened 2026-07-02
[T-20260702-0011] W-2: Model Duel (same-prompt pairs)

georgi:
Committed. Done.

## W-2: Model Duel (same-prompt pairs)

A thin wrapper over the W-1 showcase seeder. `--duel <modelA>,<modelB>` renders one **curated** prompt set through **both** models and stamps both manifest rows with a shared `params.duelId`, so PR-4's pair pages can group by pair and join matched same-prompt outputs.

### What changed
- **`marketing/seo/seed.ts`** — new `--duel <a>,<b>` flag (mutually exclusive with `--models`). It bypasses the LLM prompt-writer and `--count`; prompts come from `prompts/duels/<pair>.md`. Every prompt renders through both models, and the two rows share `params.duelId` (the join key) plus `params.duelPair` (canonical, sorted slug). Dedup is namespaced by the canonical pair so both halves always generate and re-runs stay idempotent regardless of argument order. `main()` was refactored into a shared `renderPrompt()` helper + `runStandard()`/`runDuel()` drivers over a `RunCtx`; the plain single-model path is behaviorally unchanged.
- **`marketing/seo/showcase-schema.ts`** + **`marketing/src/data/showcase.ts`** — added an optional `ShowcaseParams` (`duelId`/`duelPair`) to the manifest row and the page-entry contract PR-4 consumes.
- **`marketing/scripts/ingest-showcase.mjs`** — carry `params` through into the generated data module; the near-duplicate gate is namespaced by duel pair so a pair survives ingest with both halves intact.
- **`marketing/seo/prompts/duels/flux-schnell-vs-flux-dev.md`** (image: text rendering / physics) and **`kling-vs-hailuo.md`** (video: motion) — curated 8-prompt sets. The loader ignores markdown list markers, headings, and HTML comments, so files can carry an intro.
- **`marketing/seo/README.md`** — Model Duel section.

### Verification (all via `--dry-run`, offline)
- 8 prompts × 2 models = 16 rows; 8 distinct `duelId`s, each shared by exactly the two models.
- Re-run adds 0 rows; reversed arg order resolves the same canonical pair; budget cap still stops a partial batch.
- Ingest carries `params` into `showcaseEntries.generated.ts` with 0 drops (pairs survive).
- `npm run seo:typecheck` passes; app `tsc` shows no `src/data` errors; eslint clean on the changed source.

### Notes / caveats
- Duel mode needs only `FAL_API_KEY` (no `OPENAI_API_KEY`, since prompts are curated).
- Test batches under `seo/out/` and published assets under `public/showcase/` are gitignored / were cleaned up — this PR ships tooling + curated prompts, not generated content. The `.next` dir in the working tree is the shared, gitignored build cache and was not touched.
- The example video pair uses plausible provider slugs (`kling`, `hailuo`); the curated file name is independent of live model resolution, so it works regardless.

---

Closes task **T-20260702-0011**: W-2: Model Duel (same-prompt pairs).

### Acceptance criteria
- [x] `--duel <modelA>,<modelB>` runs every prompt through both models and stamps both manifest rows with a shared `params.duelId`
- [x] Curated duel prompt sets live in `marketing/seo/prompts/duels/<pair>.md`
- [x] Manifest output is consumable by PR-4 pair pages


georgi (commented):
## Review — W-2 Model Duel

All three acceptance criteria are met, and I verified the core behavior end-to-end (dry-run, offline) rather than trusting the diff — because the two files that hold the actual logic (`seed.ts`, `ingest-showcase.mjs`) show up as **binary** in the PR diff and are therefore invisible to a normal diff review (see finding 1).

### Verified independently
- **Criterion 1** — `--duel flux-schnell,flux-dev` → 16 rows, **8 distinct `duelId`s each shared by exactly the two models**, canonical `duelPair=flux-dev-vs-flux-schnell`, 16 distinct asset hashes. ✅
- **Criterion 2** — curated sets present at `prompts/duels/{flux-schnell-vs-flux-dev,kling-vs-hailuo}.md`, 8 prompts each; loader correctly strips comments/headings/list markers. ✅
- **Criterion 3** — `params` (`duelId`+`duelPair`) flows through `showcase-schema.ts` → `showcase.ts` → ingest into `showcaseEntries.generated.ts` (16 entries, 0 dropped). ✅
- Re-run adds **0 rows** (global-hash idempotency); reversed arg order resolves the same canonical pair and dedups; `--duel`+`--models` and single-model `--duel` are both rejected; `npm run seo:typecheck` clean.

The dedup namespacing (salt `duel:<pair>` in seed, `\0`-joined pair key in ingest) is sound — both halves of a pair always survive and aren't suppressed by an unrelated single-model render of the same prompt. Nice.

### Non-blocking findings

**1. Raw NUL bytes make the core files unreviewable as binary.** `seed.ts` and `ingest-showcase.mjs` embed literal `U+0000` bytes as key delimiters (`` `${model}\0${normalized}` ``, `` `${rec.model}\0${canon}\0duel:…` ``). Git flags them "Binary files differ" and reports **0 additions**, so this PR's entire logic is hidden from diff review. The pattern is inherited from W-1, but this PR extends it (ingest went 1→3 NULs). Please switch to the **`\0` escape sequence** in the template literals instead of a raw NUL byte — identical runtime string, but the files stay reviewable UTF-8 text. Worth doing since this whole PR series keeps touching these files.

**2. Budget interruption can orphan a duel half.** If `--budget-usd` trips between contender A and contender B of a prompt, you get a lone row whose `duelId` has no counterpart. The notes call a partial batch "valid," which is fine for the plain path, but a half-pair can't form a pair page. Consider gating the whole pair on budget before rendering either half, or dropping orphan `duelId`s at ingest.

**3. Minor / cosmetic.** Reversed arg order produces a different batch *directory* name (`batchId` isn't canonicalized) even though content dedups globally. Harmless. Also: no automated tests — consistent with W-1's tooling, but the `duelId`/dedup invariants that PR-4 depends on are good unit-test candidates.

Logic is correct, consistent with the W-1 conventions, and the criteria are satisfied. Not blocking — but please address finding 1 (the `\0` escape) so subsequent duel PRs are actually diff-reviewable.

PR #4127, opened 2026-07-05
ci(fly): deploy the NodeTool server on Fly.io + auto-deploy workflow

georgi:
Deploys the NodeTool server on **Fly.io** reusing the **same Supabase setup as the Docker deploy** (DB + asset storage + auth), and wires auto-deploy on image publish.

Live now at **https://nodetool.fly.dev** — health `{database:"ok",server:"ok"}`, `authMode:"supabase"`.

## What's here
- **`fly.toml`** — releases the published GHCR image `ghcr.io/nodetool-ai/nodetool:latest` (self-contained: `web/dist` + workflow examples baked in). With DB/storage/auth on Supabase and `SECRETS_MASTER_KEY` supplied, the container is **stateless** — no volume. Fly terminates TLS at the edge (like Cloudflare for the Docker box), so the origin runs plain HTTP on 7777.
- **`.github/workflows/fly-deploy.yml`** — triggers on the **"Docker"** workflow's successful completion on `main` (`workflow_run`), then rolls the fresh `:latest` image onto the `nodetool` app. Triggering on image publish (not on push) guarantees the image exists before deploy.

## Config (not in the image)
Secrets are set on the Fly app via `fly secrets import < nodetool/.env` — same values as the Docker `.env` (`DATABASE_URL` Supabase pooler, `SUPABASE_URL`/`KEY`, `NODETOOL_STORAGE_BACKEND=supabase`, buckets, `SECRETS_MASTER_KEY`, `AUTH_PROVIDER=supabase`). Repo secret `FLY_API_TOKEN` (deploy token scoped to `nodetool`) is set for the workflow.

## Notes
- Shares the **live prod Supabase** with the current Docker deployment (parallel instance, not an isolated copy).
- Browser login: `/api/config` returns `supabaseAnonKey:null` (mirrors the Docker `.env`). If the login screen needs it, add `SUPABASE_ANON_KEY` + `AUTH_REDIRECT_URL=https://nodetool.fly.dev`.

### DIFF
--- a/.github/workflows/fly-deploy.yml
+++ b/.github/workflows/fly-deploy.yml
@@ -1,0 +1,43 @@
+name: Deploy to Fly
+
+# Auto-deploy the NodeTool server to Fly.io (https://nodetool.fly.dev).
+#
+# The deploy unit is the GHCR image published by the "Docker" workflow
+# (.github/workflows/docker.yml). We therefore trigger on that workflow's
+# COMPLETION rather than on push, so we never release an image before it has
+# been built + pushed. On success we release ghcr.io/nodetool-ai/nodetool:latest
+# (which "Docker" tags for the just-built main commit) onto the Fly app — no
+# build happens here, Fly just pulls + rolls the image.
+#
+# Requires a repo secret FLY_API_TOKEN scoped to the `nodetool` app:
+#   fly tokens create deploy -a nodetool --expiry 8760h
+on:
+  workflow_run:
+    workflows: ["Docker"]
+    types: [completed]
+    branches: [main]
+  # Manual re-deploy of the current :latest image.
+  workflow_dispatch:
+
+concurrency:
+  group: fly-deploy
+  cancel-in-progress: true
+
+jobs:
+  deploy:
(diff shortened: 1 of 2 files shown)


georgi:
Addressed all review comments in c0cc99f85:

- **Wrong-image-on-out-of-order-builds (Copilot):** the deploy now pins to the commit-specific tag `main-<shortsha>` (matching docker.yml's `type=sha,prefix={{branch}}-`) for `workflow_run`, so a later `:latest` push can't cause us to deploy the wrong image. `:latest` is kept only for manual `workflow_dispatch`. Verified the tag format is `main-<7-char-sha>` against a real published image.
- **Missing permissions (CodeQL + Copilot):** added a workflow-level `permissions: contents: read` block. The Fly release is authorized by `FLY_API_TOKEN`, not `GITHUB_TOKEN`.
- **Unpinned `setup-flyctl@master` (Copilot):** pinned to `@v1.4`.
- **`actions/checkout` version (Copilot):** bumped `@v4` → `@v7` to match the rest of the repo.

PR #4135, opened 2026-07-05
[T-20260702-0013] W-4: Model watcher (manifest diff)

georgi:
Done. W-4 (model watcher) is implemented and committed.

## Summary

Added `marketing/scripts/model-watch.mjs`, a plain-Node watcher that surfaces provider models NodeTool has no `/models` page for. For each manifest provider it runs `nodetool generate <provider> --list-models --json`, then splits the returned ids three ways: **covered** (id matches a launch slug from `modelProviderCoverage.generated.ts`), **ignored** (id listed in `model-watch.watchlist.json`), or **new** — emitting a machine-readable report that OPS-2 (`model-watch.yml`) consumes to open one issue per provider batch.

Placed under `marketing/scripts/` next to the sibling `generate-model-coverage.mjs` it diffs against (the plan abbreviates these paths as `scripts/…`).

**Files changed**
- `marketing/scripts/model-watch.mjs` — the watcher. Token-aware slug matching (`flux` matches `fal-ai/flux/schnell` but not `influx`; `veo-3` also matches `veo3`/`veo 3`). Report shape: `{ tool, knownSlugs, providers:[{provider, ok, totalModels, covered, ignored, newCount, new:[{id, name, seedCommand}]}], newCount, errorCount }`. CLI invocation overridable via `NODETOOL_MODEL_WATCH_CLI` (built dist in CI); `--input <fixture>` bypasses the CLI; `--out`, `--provider`, `--pretty` flags.
- `marketing/scripts/model-watch.watchlist.json` — ignore list (`all` + per-provider), case-insensitive exact id match.
- `marketing/scripts/__tests__/model-watch.test.mjs` — `node --test` unit tests for the pure helpers (coverage parsing, token matching, classify split, case-insensitive suppression). All 5 pass.
- `marketing/package.json` — added `seo:model-watch` and `test:model-watch` scripts.

**Caveats / follow-ups**
- The watcher needs provider API keys, so it runs in CI secrets or a maintainer machine — no key-free path exists for every provider (as the task notes).
- `generate --list-models` enumerates image models; that's the CLI surface the task specified.
- `npm run lint` (`next lint`) couldn't run in this worktree — the `next` binary is absent from the shared `node_modules` — but the additions are plain `.mjs` outside next lint's app-source scope. The `node --test` suite runs clean.
- The per-model `seedCommand` uses a `<template>` placeholder (the model→template mapping is a maintainer decision) and a best-effort model slug from the id's last path segment.

---

Closes task **T-20260702-0013**: W-4: Model watcher (manifest diff).

### Acceptance criteria
- [x] `scripts/model-watch.mjs` diffs `nodetool generate <provider> --list-models --json` against `modelProviderCoverage.generated.ts` + `watchlist.json`
- [x] New model ids produce a machine-readable report consumable by OPS-2
- [x] Known/ignored ids in `watchlist.json` are suppressed

### DIFF
--- a/marketing/package.json
+++ b/marketing/package.json
@@ -12,6 +12,8 @@
     "seo:seed": "tsx seo/seed.ts",
     "seo:ingest": "node scripts/ingest-showcase.mjs",
     "seo:model-coverage": "node scripts/generate-model-coverage.mjs",
+    "seo:model-watch": "node scripts/model-watch.mjs",
+    "test:model-watch": "node --test scripts/__tests__/model-watch.test.mjs",
     "seo:catalog": "node scripts/generate-template-catalog.mjs",
     "seo:llms": "tsx scripts/generate-llms-txt.mjs",
     "seo:generate": "npm run seo:catalog && npm run seo:llms",
--- a/marketing/scripts/__tests__/model-watch.test.mjs
+++ b/marketing/scripts/__tests__/model-watch.test.mjs
@@ -1,0 +1,67 @@
+// Run: node --test marketing/scripts/__tests__/model-watch.test.mjs
+import { test } from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  parseCoverageSlugs,
+  matchesSlug,
+  loadWatchlist,
+  classifyProvider,
+  deriveSeedSlug,
+} from "../model-watch.mjs";
+
+test("parseCoverageSlugs pulls the object keys, not the provider ids", () => {
+  const src = `export const modelProviderCoverage: Record<string, string[]> = {
+  "flux": ["fal_ai","replicate"],
+  "veo-3": ["fal_ai"],
+};`;
+  assert.deepEqual(parseCoverageSlugs(src), ["flux", "veo-3"]);
+});
+
+test("matchesSlug is token-aware, not naive substring", () => {
+  assert.ok(matchesSlug("fal-ai/flux/schnell", "flux"));
+  assert.ok(!matchesSlug("some/influx-model", "flux"));
+  // de-hyphenated variants
+  assert.ok(matchesSlug("veo3-fast", "veo-3"));
+  assert.ok(matchesSlug("google/veo-3", "veo-3"));
+});
+
+test("classifyProvider: covered, ignored, and new are split correctly", () => {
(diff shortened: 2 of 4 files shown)


georgi (commented):
## Review — W-4 Model watcher

Verified against the codebase and ran the tool end-to-end. All three acceptance criteria are met.

**Correctness (verified)**
- `parseCoverageSlugs` correctly extracts only the object keys from `modelProviderCoverage.generated.ts` (flat `Record<string, string[]>`, array values have no trailing colon so the `"..."\s*:` regex can't pick them up). Confirmed against the real generated file — returns `seedance, veo-3, kling, …, imagen`.
- `PROVIDERS` (`fal_ai, replicate, kie, together, atlascloud, topaz`) matches the list in the sibling `generate-model-coverage.mjs` exactly.
- CLI output shape assumption is right: `generate --list-models --json` emits `{ provider, models }` with `{id, name}` entries (`packages/cli/src/commands/generate.ts:217`).
- `matchesSlug` is genuinely token-aware — ran a fixture where `fal-ai/flux/schnell` → covered, `some/influx-thing` → NOT matched as `flux`, `fal-ai/nano-banana` → new. Ignore precedence (ignored > covered > new) is correct and case-insensitive.
- All 5 `node --test` cases pass; report JSON is well-formed and matches the documented shape for OPS-2.

**Non-blocking notes**
1. `seedCommand` derives `--models <last-path-segment>` (e.g. `fal-ai/flux/schnell` → `schnell`), which won't resolve directly in `seed.ts`'s model matcher. This is fine — it's paired with the `<template>` placeholder that a maintainer must fill in anyway, so the command is a starting point, not a runnable line. Already called out as a caveat.
2. `test:model-watch` isn't wired into any aggregate `test`/CI gate (root `npm run test` covers web/electron/mobile only, and marketing has no umbrella `test` script). Consistent with the existing marketing scripts, but the unit tests will only run if invoked explicitly.
3. OPS-2 (`.github/workflows/model-watch.yml`) doesn't exist yet — expected, that's a separate task; the report is shaped to be consumed by it.

Correct, tested, and consistent with the sibling `generate-model-coverage.mjs` conventions. LGTM.

PR #4138, opened 2026-07-05
[T-20260702-0014] OPS-1: seo-seed.yml (cron + dispatch → PR)

georgi:
Committed. All acceptance criteria are met and verified locally.

## Summary

Implemented **OPS-1**, the Layer-3 CI glue that runs the programmatic-SEO showcase pipeline end-to-end and opens a review PR. The prior three attempts wrote correct code but couldn't push (the environment's git token lacks `workflow` scope); this run reconstructs and commits everything to the worktree so the orchestrator can push it.

The pipeline: **generate → guard → upload videos to R2 → ingest → open PR**. Merging the PR is the index gate — nothing auto-merges.

### Files changed
- **`.github/workflows/seo-seed.yml`** — `workflow_dispatch` (inputs `templates`, `models`, `count`, `budget_usd`, `duel`) plus a weekly `schedule` (`cron: "0 7 * * 1"`). Scheduled runs carry no inputs, so a "Resolve run parameters" step rotates one image template by ISO week (flux-schnell, count 5, budget $10). Seeds one batch per template, guards, uploads videos, ingests, then opens a `seo/seed-<batch>` PR via `peter-evans/create-pull-request@v7` with the required review checklist, committing only `showcaseEntries.generated.ts` + `public/showcase/**`.
- **`marketing/scripts/check-seed-manifest.mjs`** (`seo:guard`) — fails the job (so no PR opens) when total manifest rows == 0, or any batch's summed `costUsd` exceeds the `--budget-usd` cap. The cap is evaluated per-batch, matching the workflow's per-template cap.
- **`marketing/scripts/upload-videos-r2.mjs`** (`seo:upload-videos`) — `wrangler r2 object put` per local video, then rewrites the manifest row's `asset` to the R2 URL. `ingest-showcase.mjs` passes http(s) URLs through untouched, so videos stay out of git while images commit in-repo. Idempotent.
- **`marketing/scripts/__tests__/seed-guard.test.mjs`** — 8 passing `node:test` cases covering guard pass/empty/over-cap/per-batch and uploader rewrite/idempotent/dry-run.
- **`marketing/package.json`** — adds `seo:guard`, `seo:upload-videos`, `test:seed-guard`.

### Verification
- 8/8 guard+uploader tests pass; all scripts and `package.json` syntax-check clean; YAML has no tabs and is a verbatim copy of the previously-validated artifact.

### Caveats / follow-ups (the task's blocked decision)
- R2 target defaults to bucket `nodetool-media` / base `https://media.nodetool.ai` — overridable via the job `env` block; confirm the real bucket/media domain.
- Requires repo secrets `OPENAI_API_KEY`, `FAL_API_KEY`, `CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_ACCOUNT_ID`, and the org setting "Allow GitHub Actions to create and approve pull requests" ON.
- Could not run the real seeder here (needs built `@nodetool-ai/*` packages + provider keys, which CI installs); the guard/uploader/ingest interfaces were validated against `seed.ts`'s actual manifest shape and with fabricated manifests.

---

Closes task **T-20260702-0014**: OPS-1: seo-seed.yml (cron + dispatch → PR).

### Acceptance criteria
- [x] `workflow_dispatch` (inputs: templates, models, count, budget_usd, duel) + weekly `schedule` default batch both trigger the workflow
- [x] Successful run opens a `seo/seed-<batch>` PR with the review checklist in the body; nothing auto-merges
- [x] Job fails and posts no PR when manifest row count is 0 or spend exceeded the cap
- [x] Video assets upload to R2; images commit in-repo


georgi (commented):
## Review — OPS-1: seo-seed.yml

Verified the diff against the four acceptance criteria and the surrounding scripts (`seed.ts`, `ingest-showcase.mjs`, `showcase-schema.ts`) and repo conventions. Ran the new tests: **8/8 pass**. CI is green.

### Acceptance criteria — all met
1. ✅ `workflow_dispatch` (inputs `templates`, `models`, `count`, `budget_usd`, `duel`) + weekly `schedule` (`cron: "0 7 * * 1"`), with a "Resolve run parameters" step that supplies defaults for the input-less scheduled run (ISO-week template rotation, flux-schnell, count 5, $10). Correct.
2. ✅ Opens a `seo/seed-<suffix>` PR via `peter-evans/create-pull-request@v7` with the review checklist in the body; no auto-merge. `add-paths` is correctly scoped to `showcaseEntries.generated.ts` + `public/showcase/**`.
3. ✅ Guard fails the job (→ no PR) on 0 rows or per-batch spend over cap — covered by tests.
4. ✅ Video rows uploaded to R2 and rewritten to URLs (ingest passes http(s) through untouched); images stay in-repo — covered by tests.

Conventions check out: `actions/checkout@v7` / `actions/setup-node@v6` match the rest of `.github/workflows`; `CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ACCOUNT_ID` are the same secrets `marketing-ci.yml` uses; `costUsd` / `mediaType` / `asset` / batch-dir layout match `ShowcaseRecord`; `--budget-usd` per-template invocation lines up with the guard's per-batch cap.

---

### ⚠️ Primary concern — the pipeline shrinks the corpus rather than growing it (please verify before enabling the schedule)

The workflow's stated goal is "a small default batch keeps the corpus growing," but as wired the end-to-end run **drops previously-published entries** on every run after the first:

- `marketing/seo/out/` is **gitignored** (`marketing/.gitignore:22`), so a fresh CI checkout has an empty `seo/out` containing only *this run's* freshly-seeded batch.
- `ingest-showcase.mjs` **never reads the existing** `showcaseEntries.generated.ts` — it rebuilds `entries` from whatever batches are in `seo/out` and `writeFileSync(GENERATED_FILE, …)` **wholesale**.
- Net effect: after batch-A merges, the next run seeds batch-B into an empty `seo/out` and regenerates the file with **only batch-B**. The PR diff deletes every batch-A entry. Meanwhile `add-paths: public/showcase/**` only *adds* files, so batch-A's images are orphaned on disk while the index no longer references them.

The committed generated file is empty today (0 entries), so the *first* run is safe — this bites from the second merged batch onward, i.e. exactly what the weekly schedule produces.

Root cause is in `ingest-showcase.mjs` (from W-1), not this diff, and this PR meets its stated acceptance criteria — so I'm not requesting changes. But OPS-1 is the piece that runs this on a fresh checkout where the regression surfaces, so please resolve it before turning on the schedule. Options: have ingest merge new batches into the existing generated file, or restore/rehydrate prior batches (or their entries) before ingest. Worth a note or follow-up task either way.

### Minor
- **setup-node cache key**: `cache: "npm"` with no `cache-dependency-path` keys only off the root `package-lock.json`, but the job also runs `npm ci` in `marketing/` (which has its own lockfile — `marketing-ci.yml` sets `cache-dependency-path: marketing/package-lock.json` for exactly this reason). Not a correctness issue; marketing deps just won't be optimally cached.
- **PR body cosmetics**: the "Models" line renders the default (`flux-schnell`) even in `--duel` mode where `models` is overridden/ignored. Harmless.

### Nice touches
`10#$WEEK` guards against octal parsing of zero-padded week numbers; the `NODETOOL_WRANGLER_CMD` / `NODETOOL_SEED_OUT_DIR` test hooks make the guard+uploader hermetically testable without Cloudflare creds; the uploader is idempotent and no-ops cleanly on image-only scheduled runs (so those runs need no R2 access).

Not approving here (reviewer runs can't approve), but the diff itself is correct and well-tested for its scope. The corpus-accumulation question above is the one thing I'd want confirmed before the weekly `cron` is live.

georgi (commented):
## Review: OPS-1 seo-seed.yml — meets all acceptance criteria ✅

Reviewed the full diff and verified every interface assumption against the actual codebase. This is correct, tested, and consistent with repo conventions. No blocking issues.

### Acceptance criteria — all met
1. **Triggers** — `workflow_dispatch` (inputs `templates`, `models`, `count`, `budget_usd`, `duel`) + weekly `schedule` (`cron: "0 7 * * 1"`). Scheduled runs carry no inputs, so "Resolve run parameters" rotates one image template by ISO week. The week math (`$((10#$WEEK % 3))`) correctly forces base‑10 so `08`/`09` don't fault as octal. ✅
2. **Opens `seo/seed-<batch>` PR, nothing auto‑merges** — `peter-evans/create-pull-request@v7` on `seo/seed-<suffix>`, review checklist in the body, `add-paths` scoped to `showcaseEntries.generated.ts` + `public/showcase/**` so `seo/out` and videos never get committed. No merge step. ✅
3. **Fail‑closed guard** — `seo:guard` runs before ingest/PR and exits 1 on 0 rows or any per‑batch `costUsd` over the cap; a failed step halts the job before the PR step. ✅
4. **Videos → R2, images in‑repo** — `seo:upload-videos` rewrites each local video row's `asset` to the R2 URL; `ingest-showcase.mjs` passes http(s) URLs through untouched (`isHttpUrl(rec.asset)` branch), so videos stay out of git while images commit under `public/showcase/**`. ✅

### Interface verification (against actual code)
- `seed.ts` accepts exactly `--template/--models/--count/--budget-usd/--batch/--duel` with CSV support; `seo:seed` = `tsx seo/seed.ts`. ✓
- Manifest written to `marketing/seo/out/<batch>/manifest.jsonl`; row shape (`ShowcaseRecord`) has `id`, `batch`, `mediaType: "image"|"video"`, `asset`, `costUsd` — all fields the guard/uploader read. ✓
- `ingest-showcase.mjs` reads those manifests and writes `marketing/src/data/showcaseEntries.generated.ts` + copies local images to `public/showcase/<batch>/` — matches the PR's `add-paths`. ✓
- Action pins (`checkout@v7`, `setup-node@v6`, `peter-evans@v7`) match the versions used across the repo's other workflows. `wrangler ^4.106.0` is a marketing devDependency with its own lockfile, so `npx wrangler` resolves the local install. ✓
- `marketing` is a standalone npm project (not a root workspace), and `@nodetool-ai/dsl` resolves upward into the root `node_modules` populated by `build:packages` — so the `npm ci` → `build:packages` → marketing `npm ci` ordering is both correct and required. ✓
- Tests: 8 `node:test` cases genuinely exercise both scripts as subprocesses via `NODETOOL_SEED_OUT_DIR` / `NODETOOL_WRANGLER_CMD` — pass/empty/over‑cap/per‑batch for the guard, rewrite/idempotent/dry‑run for the uploader. Consistent with the sibling `model-watch.test.mjs` pattern.

### Non‑blocking notes (for the operator, not code changes)
- **Duel mode omits `--count`** (uses seed.ts default 5). Intentional if duel uses a curated prompt set — worth a one‑line confirm.
- **`--provider` is never set**, so any non‑fal value in `models`/`duel` still runs under the default `fal_ai` provider. Fine for the fal image/video templates in scope; note it if the model set ever spans providers.
- **R2 bucket / media domain** default to `nodetool-media` / `https://media.nodetool.ai` (overridable via the job `env`) — the task's explicitly blocked decision. Confirm the real values, plus the secrets (`OPENAI_API_KEY`, `FAL_API_KEY`, `CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_ACCOUNT_ID`) and the org setting "Allow GitHub Actions to create and approve pull requests", before the first real run.
- Uses `npx wrangler` directly rather than the repo's `cloudflare/wrangler-action@v3`; justified here since the uploader does a per‑file put + manifest rewrite loop the action doesn't wrap.

Note: this workflow is `workflow_dispatch`/`schedule` only, so it doesn't execute on the PR — the green CI reflects the repo's other checks, not an end‑to‑end seed run. The scripts are covered by their own tests; the seed→PR path is validated only interface‑wise (as the author states), which is the right call given it needs provider keys.

PR #4152, opened 2026-07-06
[T-20260705-0008] WS4a: Extend motion lint rule and define keyframe MOTION constants

georgi:
Committed. Here's the summary.

## WS4a: Extend motion lint rule and define keyframe MOTION constants

Graduated motion-timing enforcement from the single narrow `no-restricted-syntax` selector (which saw only the `transition` key and only `ms`, so it reported 0) into a dedicated custom rule + CSS script that surface the real backlog, and added the two keyframe-loop MOTION tiers the transition tiers couldn't express.

**Files changed:**
- `web/eslint.design.mjs` — new `motionTokensRule` (`design-tokens/motion-tokens`), mirroring `spacingTokensRule`. Covers `transition`/`transitionDuration`/`transitionDelay`/`animation`/`animationDuration`/`animationDelay` object props, **seconds (`0.3s`) and ms**, and raw timing inside `styled`/`css` template literals. Removed the old transition selector from `noRestrictedSyntax`.
- `web/eslint.design.config.mjs` — registers `design-tokens/motion-tokens` at **`warn`**.
- `web/scripts/lint-motion-css.mjs` — new `.css` gate for `transition`/`animation` (+ `-duration`/`-delay`) timing. Reports the backlog (**26** raw `.css` timings) but exits 0 (warn) until WS4b migrates and flips it to exit 1.
- `web/package.json` — wired `lint-motion-css.mjs` into `lint:design`/`lint:design:fix`, added `lint:motion-css`.
- `web/src/components/ui_primitives/tokens.ts` — added `MOTION.spin` (`1s linear`) and `MOTION.pulse` (`2s ease-in-out`) keyframe-loop tiers.
- `docs/DESIGN.md` — §5 keyframe table + Enforcement subsection; §11 item 5 note.

**Behavior:** `0`/`0s`, keyword-only values (`none`), and `MOTION.*`-composed values are not flagged. Rule and CSS script both stay at `warn`; WS4b migrates the backlog and promotes both to `error`.

**Caveats:**
- This worktree has **no `node_modules`** (not symlinked as the environment note claimed — `eslint`/`@typescript-eslint/parser` absent), so I could not run `npm run typecheck && npm run lint && npm run test` locally. That acceptance criterion is left unchecked for CI to confirm. I did verify: all four CSS lint scripts run and exit 0 from `web/`; `eslint.design.mjs` imports cleanly with `motion-tokens` registered and the old selector removed; and the motion regexes pass a 15-case positive/negative unit table. The rule is structurally identical to the CI-proven `spacingTokensRule`. The token change adds two keys to a `const … as const` object with no existing consumers, so it's typecheck-safe.

---

Closes task **T-20260705-0008**: WS4a: Extend motion lint rule and define keyframe MOTION constants.

### Acceptance criteria
- [x] Motion lint rule covers animation/animationDelay/transitionDuration, seconds and ms, and template-literal CSS
- [x] CSS lint script checks transition timing
- [x] MOTION.pulse/MOTION.spin (and any other keyframe constants) defined per DESIGN.md §11.5
- [x] Rule remains at warn and now reports the full backlog
- [ ] npm run typecheck && npm run lint && npm run test pass

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -472,6 +472,25 @@ import { MOTION, reducedMotion } from "../ui_primitives";
 | `MOTION.opacity` | `opacity 150ms ease` | Fade in/out |
 | `MOTION.shadow` | `box-shadow 200ms ease` | Elevation changes |
 
+### Keyframe Loops
+
+Infinite `@keyframes` animations (spinners, pulsing indicators) run on their own
+timing — the sub-second `fast`/`normal`/`slow` tiers don't fit a 1–5s loop. Two
+tokens carry the duration + easing; compose them with the keyframe name and
+iteration.
+
+| Token | Value | Use |
+|---|---|---|
+| `MOTION.spin` | `1s linear` | Continuous rotation (loading spinners) |
+| `MOTION.pulse` | `2s ease-in-out` | Breathing / pulsing status indicators |
+
+{% raw %}
+```tsx
+sx={{ animation: `${spin} ${MOTION.spin} infinite` }}
+sx={{ animation: `${pulse} ${MOTION.pulse} infinite` }}
+```
+{% endraw %}
+
 ### Usage
 
 {% raw %}
@@ -529,6 +548,10 @@ Existing components that write `@media (prefers-reduced-motion: reduce)` raw sho
 ### Forbidden
 
 Raw timing strings of any kind: `"200ms"`, `"0.2s ease-in-out"`, `"all 150ms linear"`. Compose from `MOTION.*` tokens. Raw `@media (prefers-reduced-motion: reduce) { … }` blocks in new code — use `reducedMotion()` instead.
+
+### Enforcement (motion is linted at `warn` — migration in progress)
+
+In TSX, `design-tokens/motion-tokens` flags any object-literal `transition` / `animation` / `transitionDuration` / `transitionDelay` / `animationDuration` / `animationDelay` whose value carries a non-zero `s`/`ms` timing, and any raw `transition` / `animation` timing inside `styled` / `css` template-literal CSS. In plain `.css`, `scripts/lint-motion-css.mjs` flags the same properties (and their `-duration` / `-delay` longhands). `0` / `0s` (no delay), keyword-only values (`none`), and values built from `MOTION.*` tokens are allowed. Both stay at **`warn`** while the backlog (~41 TSX timings + ~26 `.css` timings) is migrated; WS4b migrates them and promotes the rule + the CSS script to **`error`**.
 
 ---
 
@@ -689,7 +712,7 @@ When editing any UI file, scan for these violations and fix them in the same PR.
 2. **Typography**: Do not add a ninth type style. Any new text hierarchy must collapse onto one of the eight existing combinations.
 3. **Color**: Add as a `c_*` key in **both** `paletteDark.ts` and `paletteLight.ts`. Document its semantic role in a comment.
 4. **Border radius**: Add a new `BORDER_RADIUS.*` entry in `tokens.ts` and a corresponding `--rounded-*` CSS var in `ThemeNodetool.tsx` `MuiCssBaseline`.
-5. **Motion**: If a new timing is needed, add to `MOTION` in `tokens.ts` as a named constant — never use the value inline. New animated components must also include a `reducedMotion()` override.
+5. **Motion**: If a new timing is needed, add to `MOTION` in `tokens.ts` as a named constant — never use the value inline. Transitions use the `fast`/`normal`/`slow` tiers or a property shorthand; infinite `@keyframes` loops use `MOTION.spin` (1s linear) / `MOTION.pulse` (2s ease-in-out) rather than a sub-second tier. New animated components must also include a `reducedMotion()` override.
 6. **Z-index**: Add to `Z_INDEX` in `tokens.ts` or to `theme.zIndex` in `ThemeNodetool.tsx`. Never use a raw integer.
 
 ---
--- a/web/eslint.design.config.mjs
+++ b/web/eslint.design.config.mjs
@@ -68,6 +68,10 @@ export default [
       // error so any new raw/magic/var(--rounded-*) radius fails the gate. See
       // docs/DESIGN.md §4.
       "design-tokens/border-radius-tokens": "error",
+      // Motion (transition/animation timing) is mid-migration — stays at `warn`
+      // to surface the backlog. Promoted to `error` in WS4b once migrated. See
+      // docs/DESIGN.md §5.
+      "design-tokens/motion-tokens": "warn",
     },
   },
 ];
--- a/web/eslint.design.mjs
+++ b/web/eslint.design.mjs
@@ -614,7 +614,18 @@ export const borderRadiusTokensRule = {
 //
 //   1. Object props   { transition: "0.2s ease", animation: "spin 1s linear …" }
 //   2. Duration/delay { animationDelay: "0.08s", transitionDuration: "200ms" }
-//   3. Template CSS    css`transition: color 0.2s;`  animation: name 1.6s …
+//   3. Interpolated   { animation: `${keyframe} 1.6s ease-in-out infinite` }
+//   4. Template CSS   css`transition: color 0.2s;`  animation: name 1.6s …
+//
+// Form 3 is the dominant backlog shape: a keyframe name interpolated ahead of a
+// raw duration. It escapes the object-prop Literal check (the value is a
+// TemplateLiteral) *and* the CSS-decl check (there's no `animation:` text inside
+// the quasi to anchor on — the key lives in the object, not the template). The
+// Property handler therefore also scans TemplateLiteral values: the key already
+// proves it's a timing prop, so any raw non-zero `s`/`ms` in the quasis is a
+// violation. Values built purely from MOTION.* tokens keep the token in a
+// placeholder (`${MOTION.slow}`), leaving no raw timing in the quasis, so
+// `animation: `${activePop} ${MOTION.slow}`` stays unflagged.
 //
 // Any non-zero timing (`s` or `ms`) is reported. `0`/`0s` (no delay) is allowed;
 // values built from MOTION.* tokens (MemberExpressions / template placeholders)
@@ -679,6 +690,16 @@ export const motionTokensRule = {
           value.type === "Literal" &&
           typeof value.value === "string" &&
           hasNonZeroTime(value.value)
+        ) {
+          context.report({ node: value, messageId: "raw" });
+        } else if (
+          // Interpolated keyframe form: `animation: `${keyframe} 1.6s …``.
+          // The key already proves this is a timing prop, so any raw non-zero
+          // s/ms in the quasis is a violation. MOTION.*-composed values place
+          // the token in a placeholder (Expression), leaving no raw timing in
+          // the quasis, so they stay unflagged.
+          value.type === "TemplateLiteral" &&
+          value.quasis.some((q) => hasNonZeroTime(q.value.raw))
         ) {
           context.report({ node: value, messageId: "raw" });
         }
--- a/web/eslint.design.mjs
+++ b/web/eslint.design.mjs
@@ -614,18 +614,30 @@ export const borderRadiusTokensRule = {
 //
 //   1. Object props   { transition: "0.2s ease", animation: "spin 1s linear …" }
 //   2. Duration/delay { animationDelay: "0.08s", transitionDuration: "200ms" }
-//   3. Interpolated   { animation: `${keyframe} 1.6s ease-in-out infinite` }
-//   4. Template CSS   css`transition: color 0.2s;`  animation: name 1.6s …
+//   3. Interpolated prop  { animation: `${keyframe} 1.6s ease-in-out infinite` }
+//   4. Interpolated CSS   css`animation: ${keyframe} 1.9s linear infinite;`
+//   5. Template CSS       css`transition: color 0.2s;`
 //
-// Form 3 is the dominant backlog shape: a keyframe name interpolated ahead of a
-// raw duration. It escapes the object-prop Literal check (the value is a
-// TemplateLiteral) *and* the CSS-decl check (there's no `animation:` text inside
-// the quasi to anchor on — the key lives in the object, not the template). The
-// Property handler therefore also scans TemplateLiteral values: the key already
-// proves it's a timing prop, so any raw non-zero `s`/`ms` in the quasis is a
-// violation. Values built purely from MOTION.* tokens keep the token in a
-// placeholder (`${MOTION.slow}`), leaving no raw timing in the quasis, so
-// `animation: `${activePop} ${MOTION.slow}`` stays unflagged.
+// Forms 3 and 4 are the dominant backlog shape: a keyframe name interpolated
+// ahead of a raw duration.
+//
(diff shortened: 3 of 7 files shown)


georgi (commented):
## Review: WS4a — motion lint rule + keyframe MOTION constants

**Verdict: request changes** (posted as a comment — GitHub blocks `request_changes` on an own PR).

The structure is clean and mirrors the proven `colorTokensRule` / `borderRadiusTokensRule` patterns exactly. Config wiring (rule at `warn`, CSS script exit-0, `lint:design` runs without `--max-warnings`) is correct and won't fail the gate. Tokens are well-scoped, docs accurate. One correctness gap should be fixed before merge.

### Blocking: the rule misses the dominant animation backlog pattern — interpolated-keyframe object values

The task's core goal is to *surface the true backlog count* (criterion 1 explicitly lists `animation` + template-literal CSS; the PR claims "~41 raw animation timings"). But the rule silently skips the most common real form of a keyframe-loop animation — an object property whose value is a template literal with the keyframe name interpolated **before** the raw duration:

```tsx
// web/src/components/node/BaseNode.tsx:203  (object property, not tagged CSS)
animation: `${ambientPulseKeyframes} 1.6s ease-in-out infinite`
```

Why it escapes both handlers:
- **`Property` handler** returns early — `value.type` is `TemplateLiteral`, not `Literal`, so it's never checked.
- **`TemplateElement` handler** can't match — the quasi after the `${…}` interpolation is ` 1.6s ease-in-out infinite`, which has no `animation:`/`transition:` keyword for `CSS_MOTION_DECL` to anchor on (the value capture `[^;{}]*` was truncated at the interpolation boundary in the preceding quasi).

This is not an edge case. Across `web/src`:
- **~19** genuine raw-timing `animation: \`${keyframe} <dur> …\`` object props are **missed** (SelectionActionBar, GeneratingLayerOverlay, MagicGenerationFill, LoadingSpinner, BaseNode, ChainNodeCard, QueueOverlay, WelcomeFlow, PropertyField, …).
- Only **~12** plain-string-literal `animation: "…1.6s…"` props are caught.

So the rule surfaces fewer than half of the object-literal animation backlog — and the missed half is exactly the spinner/pulse loops that `MOTION.spin` / `MOTION.pulse` were added to replace. WS4b, which will be driven off this rule's output, would migrate the string-literal animations but never see the interpolated ones. `MOTION.slow`-composed values (e.g. `AssetViewer.tsx:192`) are correctly not flagged — good.

### Suggested fix (small, scoped)

Extend the `Property` handler to also scan `TemplateLiteral` values — the key already establishes it's a timing property, so any raw `s`/`ms` in the quasis is a violation:

```js
Property(node) {
  const name = keyName(node.key);
  if (!name || !MOTION_PROP_KEYS.has(name)) return;
  const value = node.value;
  if (value.type === "Literal" && typeof value.value === "string" && hasNonZeroTime(value.value)) {
    context.report({ node: value, messageId: "raw" });
  } else if (value.type === "TemplateLiteral" &&
             value.quasis.some((q) => hasNonZeroTime(q.value.raw))) {
    context.report({ node: value, messageId: "raw" });
  }
},
```

This catches the object-property case (the bulk). The same interpolation-truncation also affects `animation: ${name} 1.6s` inside tagged `css`/`styled` templates, but those are fewer — the object-property fix is the high-value one. Worth a quick re-count of the surfaced backlog after the change to confirm it lands near the "~41" the task expects.

### Non-blocking
- The `.css` script's per-line single-match (`code.match(...)`, not `matchAll`) only checks the first declaration on a line — but this mirrors `lint-radius-css.mjs` exactly, so it's a consistent, acceptable limitation.
- Acceptance criterion 5 (`typecheck && lint && test`) couldn't be run locally (no `node_modules`); CI currently shows 0 failures / 11 green / 5 pending, so this looks on track once green.

Everything else looks good — please close the interpolated-animation gap and I'll take another pass.

georgi:
## Fixed the blocking gap: interpolated-keyframe timing props

Pushed `0cefb8ebd`. The `motionTokensRule` Property handler now scans `TemplateLiteral` values, not just `Literal`:

```js
} else if (
  value.type === "TemplateLiteral" &&
  value.quasis.some((q) => hasNonZeroTime(q.value.raw))
) {
  context.report({ node: value, messageId: "raw" });
}
```

The key already establishes the prop is a timing prop, so any raw non-zero `s`/`ms` in the quasis is a violation. Pure `MOTION.*` composition keeps the token in a placeholder (`${MOTION.slow}`), leaving no raw timing in the quasis — so it stays unflagged.

### Re-counted backlog

No parser is present in this worktree (`eslint`/`@typescript-eslint/parser` still absent, same as the original PR note), so I approximated both handlers with a text scan over the non-exempt `web/src` files. Results:

- **Object props, string-literal value** (pre-existing branch): ~25
- **Object props, template-literal value** (NEW branch): **23**, including every case the review named — `BaseNode.tsx:203`, `SelectionActionBar.tsx:352`, `ChainNodeCard.tsx:79`, `QueueOverlay.tsx:67`, `PropertyField.tsx:41`, and the six `WelcomeFlow.tsx` staggered `rise`/`pulse` loops.
- **Combined: ~48** (crude regex over-approximates slightly vs. an AST-exact count; lands near the expected ~41, well above the pre-fix "under half").

The text scan is a line-based regex approximation, so treat the absolute numbers as a lower-bound sanity check, not the exact ESLint count — CI runs the real rule.

### Constraints held

- `AssetViewer.tsx:192` (`animation: \`${activePop} ${MOTION.slow}\``) is **not** flagged — empty quasis, pure token composition. ✓
- `AssetViewer.tsx:148` / `:203` (`transform 140ms`/`200ms` mixed with `MOTION.*`) **are** flagged — genuine raw timing. ✓
- `LoadingSpinner.tsx` (interpolated `${pulse} 1.4s`) is exempt — it's in `ui_primitives/`, the sanctioned home. ✓
- Rule stays at **warn** (WS4a scope); `.css` script unchanged.

Updated the rule header comment to document form 3 (interpolated keyframe) as the dominant backlog shape.

georgi (commented):
## WS4a review — motion lint rule + keyframe MOTION tokens

⚠️ **Posting as a comment (the API blocks request-changes on an own PR), but this contains one blocking item — please treat "Must fix" below as request-changes.**

Solid, well-documented change. `motionTokensRule` mirrors `spacingTokensRule` cleanly, the dead `no-restricted-syntax` transition selector is removed, and the tokens match DESIGN.md. Verified what I could locally (no `node_modules` in the worktree, so no full eslint run):

**Verified ✓**
- `scripts/lint-motion-css.mjs` runs, reports **26** `.css` timings, exits 0 (warn). ✓
- Matcher logic: `hasNonZeroTime` passes 14/14 cases (`200ms`/`0.2s`/`1.6s` flagged; `0`/`0s`/`none`/`cubic-bezier(…)`/`steps(4,end)`/`var(--…)` not). The `CSS_MOTION_DECL` anchor `(?:^|[;{}\n])` correctly skips `animation-name`/`-timing-function` and, usefully, prose like `"log: animation: 300ms"` — so TemplateElement false-positives on non-CSS templates are unlikely.
- `motion-tokens` registered in `designTokensPlugin`, wired at **`warn`**. `lint:design` runs `eslint … src` with **no `--max-warnings`**, so new warnings won't break CI. ✓
- `MOTION.spin` (`1s linear`) / `MOTION.pulse` (`2s ease-in-out`) match DESIGN.md §5. ✓

**Must fix — rule misses the interpolated keyframe form inside `styled`/`css` *tagged* templates**

The Property-handler fix (`0cefb8ebd`) handles `{ animation: `${kf} 1.6s …` }` by reasoning "the key proves it's a timing prop, scan the quasis." The same shape in a **tagged** template escapes both handlers:

```tsx
// web/src/components/chat/feedback/LoadingIndicator.tsx
svg:    css`animation: ${rotate} 1.9s linear infinite;`,
circle: css`animation: ${dash} 1.55s cubic-bezier(0.42,0,0.28,1) infinite;`,
core:   css`animation: ${pulse} 1.55s cubic-bezier(0.42,0,0.28,1) infinite;`,
// web/src/components/ui_primitives/WarningBanner.tsx:76
animation: ${animate ? css`${pulseAnimation} 2s ease-in-out infinite` : "none"};
```

`animation:` sits in one quasi (empty value); the `1.9s` is in the quasi *after* `${rotate}`, which has no `animation:`/`transition:` text for `CSS_MOTION_DECL` to anchor on. It's not an object Property either, so `Property` never fires. These ~4 timings go unreported — and they're precisely the spin/pulse loops `MOTION.spin`/`MOTION.pulse` exist to replace.

Impact: undercuts two acceptance criteria for the exact shape this PR introduces tokens for — "covers … template-literal CSS" and "reports the full backlog" — and leaves an enforcement hole: when WS4b flips the rule to `error`, this shape stays invisible, so a future `animation: ${kf} 1.6s` regression won't be caught. Since this is the rule PR (WS4b is migration only), the gap belongs here.

Suggested approach: in `TemplateElement`, carry a little cross-quasi state — if an earlier quasi opens an unterminated `transition`/`animation[-duration|-delay]:` decl, run `hasNonZeroTime` on the leading portion (up to the next `;{}`) of the following quasi. Reuses the object-Property insight for the tagged-template case.

**Non-blocking notes**
- AC *"npm run typecheck && lint && test pass"* is unchecked and CI is still **pending** (9 passed, 0 failed at review time). I couldn't run the full suite locally (no `node_modules`); confirm the `lint:design` job goes green before merge.
- No unit test for the new rule — consistent with the sibling rules (none have tests), so not a blocker. If you close the gap above, a tiny RuleTester fixture for the tagged-template case is worth adding given how subtle the quasi-splitting is.

Once the tagged-template interpolation gap is closed (or explicitly deferred with a code comment + tracked follow-up if WS4b should own it), this is good to merge.

georgi:
## Fixed: interpolated keyframes inside styled/css tagged templates

Pushed `379ce18ae`.

### The fix

Replaced the per-quasi `TemplateElement` handler with a `TemplateLiteral` walk that carries an **"open motion decl"** flag across quasis, as suggested:

```js
TemplateLiteral(node) {
  let open = false;
  for (const quasi of node.quasis) {
    const raw = quasi.value.raw;
    let reported = false;
    if (open) {                                   // continuation of a split decl
      const head = raw.split(/[;{}]/, 1)[0];
      if (hasNonZeroTime(head)) { context.report({ node: quasi, messageId: "raw" }); reported = true; }
    }
    if (!reported) { /* fully-contained decls in this quasi */ }
    open = MOTION_DECL_OPEN_AT_END.test(raw) || (open && !/[;{}]/.test(raw));
  }
}
```

When an earlier quasi leaves a `transition`/`animation:` timing decl unterminated (interrupted by `${keyframe}`), the leading portion of the next quasi — up to the next `;{}` — is checked as the continuation of that decl's value. The `open && !/[;{}]/.test(raw)` clause keeps a decl open across empty/placeholder-only gaps (`animation: ${a}${b} 2s`).

### Newly surfaced

Grep of the interpolated-CSS motion pattern over non-exempt files returns exactly the loops the review named:

```
src/components/chat/feedback/LoadingIndicator.tsx:49  animation: ${rotate} 1.9s linear infinite;
src/components/chat/feedback/LoadingIndicator.tsx:58  animation: ${dash} 1.55s cubic-bezier(...) infinite;
src/components/chat/feedback/LoadingIndicator.tsx:72  animation: ${pulse} 1.55s cubic-bezier(...) infinite;
```

The **`WarningBanner.tsx`** case is in `src/components/ui_primitives/**`, which `designTokenIgnores` exempts (the primitive layer is the sanctioned home for raw values) — so it's correctly *not* linted, and its inner `css`${pulseAnimation} 2s …`` sits in a nested template whose `animation:` key lives in the outer literal. That's the 4th timing from your count; the 3 enforceable ones in `LoadingIndicator` are now surfaced. Full object+template backlog lands ~51 by approximate scan.

### RuleTester fixture

Added `web/scripts/test-motion-rule.mjs` — an ESLint `RuleTester` fixture pinning the quasi-splitting: both interpolated forms, MOTION-token composition (valid), fully-contained decls, duration+delay in a trailing quasi, and zero timings. It runs under **Node** (both it and `eslint.design.mjs` are ESM — Jest always treats `.mjs` as ESM and the web Jest run is CJS, so a Jest test can't import the rule), wired into `npm run lint:design`, which CI runs (`quality-checks.yml` + root `lint`). Also exposed as `npm run lint:motion-rule`.

### Verification

No parser is present in this worktree (same as prior notes), so I validated the cross-quasi algorithm with a 14-case standalone simulation over hand-split quasi arrays — all pass, including the object-interpolation case (the `TemplateLiteral` walk reports 0 there; the `Property` handler covers it, so **no double-report**), multi-placeholder open decls, and the realistic `LoadingIndicator` styled block. CI runs the actual `RuleTester`.

Rule stays at **warn** (WS4a scope).

georgi (commented):
## Review — WS4a: Extend motion lint rule + keyframe MOTION constants

**Verdict: LGTM.** Correct, well-tested, and consistent with the sibling design-token rules. No blocking issues.

### What I verified
Since the PR author couldn't run the suite locally (no `node_modules` in their worktree), I drove the actual `motionTokensRule.create()` visitors directly with hand-built AST nodes:

- **All 16 RuleTester fixtures pass** (9 valid → 0 reports, 7 invalid → exactly 1 report each). The tricky split-decl case (`css\`animation: ${rotate} 1.9s …\``) where `animation:` and `1.9s` land in different quasis is handled correctly by the cross-quasi "open decl" flag, and the object-property + `TemplateLiteral` handlers don't double-report on interpolated-keyframe object props.
- **13 additional adversarial cases pass**, confirming no false positives on `animation-name` / `animation-timing-function` / `transition-property`, prose containing `animation:`, `MOTION.*` MemberExpressions, mixed token+raw shorthands, and `0`/`0s`; and no false negatives on seconds (`0.08s`), ms, uppercase `1S`, and multi-decl splits.
- `node --check` passes on all four new/changed `.mjs` files.
- `lint-motion-css.mjs` runs and reports the **26-item `.css` backlog**, exiting 0 (warn) as intended.
- CI: **16/17 green, 0 failures** — including the `quality-checks` `lint:design` gate that now runs `test-motion-rule.mjs` + `lint-motion-css.mjs` where the deps exist. This corroborates the local caveat and effectively closes the 5th acceptance criterion.

### Acceptance criteria
- [x] Rule covers `animation`/`animationDelay`/`transitionDuration` (+ `transition`/`transitionDuration`/`transitionDelay`/`animationDuration`), seconds and ms, and template-literal CSS
- [x] CSS lint script checks transition (and animation) timing
- [x] `MOTION.spin` (`1s linear`) / `MOTION.pulse` (`2s ease-in-out`) defined per DESIGN.md §11.5
- [x] Rule stays at `warn` and now surfaces the backlog (~41 TSX + 26 `.css`)
- [x] `typecheck && lint && test` — CI green (verified via checks, not local run)

The `TemplateLiteral`-with-open-flag approach is a genuine improvement over the sibling rules' per-`TemplateElement` handling — it's the only one that catches the dominant `${keyframe} <raw-duration>` split, which is the backlog's real shape. Good call adding the RuleTester fixture to pin it down.

### Minor, non-blocking (fine as-is for a warn-level backlog counter)
1. `lint-motion-css.mjs` uses `line.match(...)` (first declaration per line only), so a rare one-line `color: red; transition: all .2s` would miss the second decl. CSS here is one-decl-per-line by convention, and this matches the existing `lint-radius-css.mjs` pattern — no change needed.
2. Numeric (non-string) values like `transitionDuration: 200` aren't flagged, consistent with the other object-literal rules. Acceptable.

Approvable once the final check goes green.

PR #4262, opened 2026-07-15
fix(electron): preserve falsy workflow outputs

ShiroKSH:
## Summary

- preserve valid `0`, `false`, empty-string, and `null` values emitted by Electron workflow Output nodes
- keep missing or explicit `undefined` outputs out of the result list
- cover the packed WebSocket `node_update` through completed `onComplete` path

## Root cause

`WorkflowRunner` used truthiness to decide whether an Output node produced a result. Output values are intentionally untyped, and the generic Output node normalizes an absent value to `null`, so value presence cannot be inferred from boolean coercion.

## Test plan

- `npm run build:packages`
- `npm run typecheck --workspace=electron`
- `npm run lint --workspace=electron`
- `npm run lint`
- `npm test --workspace=electron -- --runInBand src/__tests__/WorkflowRunner.messages.test.ts`

### DIFF
--- a/electron/src/WorkflowRunner.ts
+++ b/electron/src/WorkflowRunner.ts
@@ -186,7 +186,7 @@ export const createWorkflowRunner = () =>
               });
             } else if (
               data.result &&
-              data.result.output &&
+              data.result.output !== undefined &&
               data.node_name?.includes("Output")
             ) {
               set({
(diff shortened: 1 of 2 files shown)


georgi (approved):
Clean, well-scoped bug fix. The change from truthiness check (`&&`) to explicit undefined check (`!== undefined`) correctly preserves falsy-but-valid workflow outputs (`0`, `false`, `""`, `null`). This aligns with the web frontend pattern in `web/src/components/node/outputResult.ts`.

The parameterized tests covering all falsy values plus the negative case for `undefined` are thorough.

Note: CI didn't run because this PR is from a fork. The tests and typecheck should be verified before merging (either by running the PR's test plan commands manually or by triggering CI).

---

georgi:
@ShiroKSH thanks for taking the time to fix this!

PR #4478, opened 2026-07-25
Add one-command server startup and a Claude starter kit

georgi:
## Why

Getting the server up from a fresh checkout took four steps with three ways to fail:

- wrong Node version,
- an `npm install` that rolls its whole tree back behind a proxy (keytar/libsecret, Electron and onnxruntime postinstall downloads),
- a package build that's easy to skip — `base-nodes` resolves from `dist/`, so skipping it starts a server with **zero node types registered** and no obvious error.

On headless Linux there was a fourth, which I hit while testing this branch: `keytar` can't load without a system keychain, and the server **aborts during startup**:

```
ERROR | nodetool.security.master-key | keytar native module failed to load...
ERROR | nodetool.websocket.server | Database setup failed KeychainAccessError
```

## What changed

**`./start.sh`** — one command, does each setup step only when it's actually missing:

```bash
./start.sh          # API on :7777
./start.sh full     # API + web UI on :3000
./start.sh web      # web UI only
./start.sh check    # typecheck + lint + test
./start.sh doctor   # report state, change nothing
PORT=8080 ./start.sh
```

**`scripts/ensure-dev-env.mjs`** — creates `.env` and generates a `SECRETS_MASTER_KEY`, which short-circuits the keychain lookup that fails in containers. Never overwrites an existing key, so secrets stay readable.

**`.claude/` starter kit** — gives web sessions the same setup automatically:

| Path | Purpose |
| :--- | :--- |
| `hooks/session-start.sh` | Installs dependencies on web-session start, using the flags that survive sandboxed/proxied containers. No-ops locally and on repeat runs. |
| `settings.json` | Registers the hook. |
| `commands/serve.md` | `/serve` — start the API and poll until it answers. |
| `commands/verify.md` | `/verify` — typecheck, lint, test, fix what breaks. |
| `commands/onboard.md` | `/onboard <area>` — owning workspace, entry point, nearest example, applicable pitfalls. |

**`.gitignore`** — `.claude/*` previously excluded everything but `skills/`, so the kit would never have been committed; the shared files are now un-ignored explicitly while `settings.local.json` stays local. The local SQLite database is now ignored instead of appearing as untracked noise after any server run.

**Docs** — README quick start now leads with `./start.sh` (manual steps kept in a `<details>` block, with a note that `build:packages` isn't optional on a fresh checkout). CLAUDE.md points at the kit.

## Validation

Run in a fresh container with no `node_modules`:

- ✅ Hook completed exit 0 — 3622 packages installed, `better-sqlite3` compiled
- ✅ `./start.sh` → health 200 in ~4s, `{"database":"ok","server":"ok"}`
- ✅ **2376 node types** registered via `/api/nodes/metadata`
- ✅ `./start.sh doctor` reports without mutating (caught and fixed a bug where it triggered a full build)
- ✅ `./start.sh sever` rejects in <1s without building
- ✅ `npm run lint:packages` exit 0 (warnings are pre-existing, untouched files)
- ✅ `npm run test --workspace=packages/security` — 76/76 passing
- ✅ `ensure-dev-env.mjs` idempotent across repeat runs

## Notes

The SessionStart hook runs **synchronously**, so dependencies are guaranteed present before the session starts — no race where a test run hits a half-installed tree. The trade-off is that a web session waits on the first install. Switching to async is a one-line change documented in `.claude/README.md`.

The hook only runs when `CLAUDE_CODE_REMOTE=true`, so local terminal sessions are unaffected. It uses `apt-get` only when already running as root and treats failure as non-fatal.

https://claude.ai/code/session_01SjDfVouuTcuKGNq6LXTjUA

---
_Generated by [Claude Code](https://claude.ai/code/session_01SjDfVouuTcuKGNq6LXTjUA)_

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,0 +1,42 @@
+# Claude starter kit
+
+Everything here exists so a fresh session can build, run, and verify NodeTool
+without being told how.
+
+## What's in it
+
+| Path | What it does |
+| :--- | :--- |
+| `hooks/session-start.sh` | Installs dependencies when a Claude Code **web** session starts, using the flags that survive sandboxed/proxied containers. No-ops locally and on repeat runs. |
+| `settings.json` | Registers the hook. |
+| `commands/serve.md` | `/serve` — start the API on :7777 in the background and poll until it answers. |
+| `commands/verify.md` | `/verify` — typecheck, lint, test, and fix what breaks. |
+| `commands/onboard.md` | `/onboard <area>` — locate the owning workspace, entry point, nearest example, and the pitfalls that apply. |
+| `skills/` | 18 NodeTool skills (workflow building, custom nodes, API reference, deployment, troubleshooting, code review). Claude loads these on its own when a task matches. |
+
+## Running things
+
+```bash
+./start.sh          # API on :7777 — installs and builds on first run
+./start.sh full     # API + web UI on :3000
+./start.sh check    # typecheck + lint + test
+./start.sh doctor   # what's set up, what isn't
+```
+
+`./start.sh` is the same path the hook prepares, so the first command in a
+session works whether or not the hook ran.
+
+## Why the hook is web-only
+
+It guards on `CLAUDE_CODE_REMOTE=true`. Web sessions get a fresh container each
+time and would otherwise start with no `node_modules`; your local checkout
+already has one and shouldn't pay an install check on every session.
+
+To make it run locally too, drop the guard at the top of `session-start.sh`.
+
+## Changing it
+
+The hook runs synchronously — the session waits for the install to finish, so
+Claude never starts a test run against a half-installed tree. For a faster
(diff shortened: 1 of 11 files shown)


georgi:
`copilot-pull-request-reviewer` failed, but the cause is in the check's own setup, not this diff:

```
/home/runner/work/_temp/ghcca-node/node/bin/node: No such file or directory
##[error]Process completed with exit code 127
```

It exited after ~6s while bootstrapping its bundled Node runtime (`GITHUB_TOKEN` is also empty in the job env), so it never read the changes. Nothing in this PR can affect it — re-running the job is the fix if a Copilot review is wanted.

`Markdown link check (relative links)` passed. The remaining checks (Quality Gate, Workflow Integration Tests, Browser E2E, CodeQL) are still running; I'm watching and will push fixes for anything this diff actually causes.

---

PR #4490, opened 2026-07-26
Resync the root lockfile, and drop an unused test import

georgi:
Two small things, the second of which unblocks CI for every open PR.

## Resync the root lockfile

Every job that runs `npm ci` at the repo root has been failing on `main`:

```
npm error Missing: electron-builder@26.15.3 from lock file
npm error Invalid: lock file's js-yaml@5.2.1 does not satisfy js-yaml@5.2.2
npm error Invalid: lock file's sharp@0.35.0 does not satisfy sharp@0.35.3
```

#4472 bumped `electron/package.json` and regenerated `electron/package-lock.json`, but not the **root** one. `electron/` is a root npm workspace, so the root lock is what `npm ci` validates at the repo root, and it still pinned `electron-builder@26.8.1` against the manifest's `26.15.3`.

Why it went unnoticed: `npm install` is lenient and reconciles the mismatch in place, so local development works fine. `npm ci` is strict and refuses — so only CI saw it.

Regenerated with `npm install --package-lock-only`, meaning the only inputs are the manifests already on `main`; no manifest was edited and no version picked by hand. The diff is large (+1041/−884) because the old lock was missing electron-builder's entire transitive subtree — precisely what the `Missing: app-builder-lib@…` / `dmg-builder@…` / `builder-util@…` errors reported. Direct-dependency changes are only #4472's four: `electron-builder` 26.8.1→26.15.3, `@electron/rebuild` ^4.0.3→^4.0.4, `js-yaml` ^5.2.2, `sharp` ^0.35.1.

## Drop an unused test import

Follow-up to #4487, which merged before this review comment was addressed. `makeClipVersion` was imported into `mobile/src/screens/__tests__/TimelineViewerScreen.test.tsx` but never used — the fixture builds its clip versions as object literals.

Worth recording why it slipped through: `oxlint` does not flag it (unused imports are not in its `correctness` category) and `tsc --noEmit` does not either (`noUnusedLocals` is off), so neither pre-commit gate would have caught it. I checked the tree's other two `@nodetool-ai/timeline` imports rather than only the flagged one; both are used.

## Testing

- `npm ci --dry-run` → succeeds (was exit 1)
- `check:deps`, `check:coupled-deps`, `check:circular` → pass
- `npm run build:packages` → 58/58 successful
- mobile `tsc --noEmit` clean; `npx oxlint src` clean apart from one pre-existing `curly` warning in `ChatStore.ts`
- mobile `npx jest` → 45 suites / 710 tests pass

### DIFF
--- a/mobile/src/screens/__tests__/TimelineViewerScreen.test.tsx
+++ b/mobile/src/screens/__tests__/TimelineViewerScreen.test.tsx
@@ -3,8 +3,6 @@ import { StyleSheet } from 'react-native';
 import { act, fireEvent, render, screen, waitFor } from '@testing-library/react-native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
 
-import { makeClipVersion } from '@nodetool-ai/timeline';
-
 import TimelineViewerScreen from '../TimelineViewerScreen';
 import type { RootStackParamList } from '../../navigation/types';
 import { resetDocumentStores } from '../../documents/documentStore';
--- a/package-lock.json
+++ b/package-lock.json
@@ -128,12 +128,12 @@
         "bufferutil": "^4.1.0",
         "electron-log": "^5.4.4",
         "electron-updater": "^6.8.9",
-        "js-yaml": "^5.2.1",
+        "js-yaml": "^5.2.2",
         "msgpackr": "^1.12.1",
         "node-llama-cpp": "^3.0.0",
         "react": "^19.2.7",
         "react-dom": "^19.2.7",
-        "sharp": "^0.35.0",
+        "sharp": "^0.35.1",
         "sqlite-vec": "^0.1.9",
         "ws": "^8.21.0",
         "zod": "^4.4.3",
@@ -153,7 +153,7 @@
         "@typescript-eslint/parser": "^8.63.0",
         "@vitejs/plugin-react": "^6.0.2",
         "electron": "39.8.10",
-        "electron-builder": "26.8.1",
+        "electron-builder": "26.15.3",
         "esbuild": "^0.28.1",
         "eslint": "^9.39.4",
         "jest": "^29.7.0",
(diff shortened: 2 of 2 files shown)


georgi:
CI is red here, but not because of this change — every job that runs `npm ci` at the repo root is failing, and it reproduces on `main`. This PR deletes two lines from one mobile test file and touches no manifest.

**Cause:** the root `package-lock.json` is out of sync with `electron/package.json`. #4472 (`chore(deps): bump the npm_and_yarn group across 1 directory with 8 updates`) bumped the electron workspace and regenerated `electron/package-lock.json`, but not the **root** lock. `electron/` *is* a root npm workspace, so the root lock is the one `npm ci` validates at the repo root, and it still pins the old versions:

```
root package-lock.json:  "electron-builder": "26.8.1"
electron/package.json:   "electron-builder": "26.15.3"
```

Reproduced locally with `npm ci --dry-run` at `af6aa4e` (main + this 2-line test edit):

```
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json ... are in sync.
npm error Missing: electron-builder@26.15.3 from lock file
npm error Invalid: lock file's js-yaml@5.2.1 does not satisfy js-yaml@5.2.2
npm error Invalid: lock file's sharp@0.35.0 does not satisfy sharp@0.35.3
npm error Missing: app-builder-lib@26.15.3 from lock file
...
```

The three "Missing/Invalid" roots match #4472's three direct bumps exactly (electron-builder, js-yaml, sharp), with the rest being their transitives.

**Fix:** `npm install` at the repo root and commit the regenerated root `package-lock.json`. I have not done it here on purpose — it is a repo-wide dependency change that can shift transitive versions across the whole tree, and folding a large lockfile diff into an unused-import removal would bury it. It also blocks every open PR, not just this one, so it wants its own change rather than riding along with this one.

Happy to push that as a separate PR if you'd like.

---

georgi:
Pushed the lockfile resync (`63b5767`) here rather than as a separate PR, since it's what unblocks this one.

Regenerated with `npm install --package-lock-only`, so the only inputs are the manifests already on `main` — no manifest was edited and no version was chosen by hand. `npm install` is lenient and reconciles the mismatch in place, which is why local dev never noticed; `npm ci` is strict and refuses, which is why CI did.

Verified locally before pushing:

- `npm ci --dry-run` → succeeds (was exit 1)
- `check:deps`, `check:coupled-deps`, `check:circular` → pass
- `npm run build:packages` → 58/58 successful
- mobile `tsc --noEmit` clean, 45 suites / 710 tests pass

The diff is large (+1041/−884) but its shape is expected: the old lock was missing electron-builder's transitive subtree entirely, which is what the original `Missing: app-builder-lib@…`, `dmg-builder@…`, `builder-util@…` errors were reporting. The direct-dependency changes are only #4472's four: `electron-builder` 26.8.1→26.15.3, `@electron/rebuild` ^4.0.3→^4.0.4, `js-yaml` ^5.2.2, `sharp` ^0.35.1.

---

PR #4493, opened 2026-07-26
Docs: PRD, design, and implementation plan for workflow triggers

georgi:
Three design docs, no code. Follows the `multiplayer-design.md` / `-implementation-plan.md` precedent.

- [`docs/triggers-prd.md`](docs/triggers-prd.md) — problem, users, 21 use cases, requirements, phasing
- [`docs/triggers-design.md`](docs/triggers-design.md) — data model, dispatch loop, state semantics, scheduling, auth, alternatives
- [`docs/triggers-implementation-plan.md`](docs/triggers-implementation-plan.md) — 10 PRs with file targets and exit criteria, ~32 engineer-days

## Why

The five `nodetool.triggers.*` nodes do not work as automations:

- **`ManualTrigger` cannot receive events at all.** Its inbox has no registered upstream, so `run()` sees end-of-stream on the first read and exits emitting nothing. `pushInputValue()` writes to a node's *downstream* inboxes, not its own, so the mechanism named in its own doc comment (`triggers.ts:181`) cannot reach it.
- **`WebhookTrigger` binds `http.createServer` inside the workflow process.** Two concurrent runs collide on the port, and on Fly only the app port is routed. No body cap, no signature verification.
- **No generator node observes `runner.cancel()`.** Cancelling a webhook run leaves the port bound and the generator parked until the process exits.
- **`Wait` returns instantly for any delay over ~24.8 days.** `setTimeout` clamps above 2³¹−1 ms — a pitfall the kernel documents at `packages/kernel/src/trigger.ts:98` and works around, and that `WaitNode.process()` reintroduces.
- **No product surface.** Nothing in `web/src` references trigger nodes.

Behind them sits a layer that was ported from Python and never connected: `trigger.ts`, `trigger-manager.ts`, `trigger-wakeup.ts` (784 lines, 1143 lines of tests, 63 cases, zero non-test callers), `Workflow.hasTriggerNodes()` (zero callers), and six tables with models and migrations — `trigger_registrations`, `trigger_inputs`, `run_inbox_messages`, `run_leases`, `run_node_state`, `workers`. `resumeJob()` delegates to `reconnectJob()`, so resume-from-suspension does not exist either.

The code is careful and well tested. It guards races that cannot happen, because nothing calls it.

## What the docs propose

**A trigger is a row in a table, not a node in a graph.** One loop claims due rows and enqueues ordinary jobs, so the graph stays a pure function of its inputs and schedules survive restart without a watchdog. The existing job queue supplies concurrency limits, the secret store supplies credentials, the cost ledger supplies attribution.

Twenty-one use cases shaped it, weighted toward freelancers and small agencies. Two changed the shape:

- **A run may return its own `next_fire_at`.** One nullable output subsumes job polling with backoff, client reminders, renewal-date scheduling, and per-client cadence. Cron and interval become defaults for the first fire; after that the workflow schedules itself, where the user can see and edit the logic.
- **Registrations are data a workflow can create.** An onboarding workflow registers a per-client trigger with that client's config, secret scope, and tags. Multi-tenancy becomes user-built rather than a platform feature.

The rest of what the use cases surfaced was operational safety, not mechanism: consecutive-failure auto-disable, bounded retries, spend ceilings, HMAC verification, and an untrusted-payload boundary for triggered agent runs.

## Scope notes

- Explicit non-goal: resuming a run that died mid-flight. Polling and webhooks retry from the top correctly, and approval flows split into two workflows plus a stored draft. The durable-execution substrate already in the schema stays there for a Phase 2 that may never be needed.
- One correction carried through the docs: state writes back on failure too. Run status controls retry scheduling; the `state` output controls position. Conflating them makes a partial batch reprocess items it already posted.
- Net effect if implemented: ~2800 lines deleted, ~900 added.

## Testing

Docs only. `npm run lint` covers `packages/*/src`, `web/src`, `electron`, and `mobile/src`, so it does not touch these files, and there is nothing to typecheck. Checked against `docs/WRITING_STYLE.md` for the forbidden-expression list.

## Open questions for review

1. Reaper threshold for stuck `claimed` rows — a 25-minute render must not be reaped. Derive from the job's own timeout?
2. Should `state` write-back require an explicit opt-in, so a workflow that happens to have an output named `state` does not silently become stateful?
3. Budget accounting: reuse the `application_budgets` reserve-then-reconcile ledger keyed on `registration_id`, or a simpler per-period sum over cost records?

---
_Generated by [Claude Code](https://claude.ai/code/session_01DV1BxggBoaHBj3KCMvafhk)_


georgi:
CI is red on six checks, all from the same pre-existing cause on `main`, not from this PR.

Every failure is the `npm ci` install step, before any check reaches code or docs:

```
npm error Missing: bytestreamjs@2.0.1 from lock file
npm error Missing: commander@9.5.0 from lock file
npm error Missing: jsonfile@6.2.1 from lock file
npm error Missing: universalify@2.0.1 from lock file
```

`package-lock.json` is out of sync with `package.json`. `bytestreamjs` appears zero times in the lockfile, and the lockfile was last modified in 10daf8d, which was already on `main` before this branch was cut.

This PR adds three markdown files and nothing else:

```
docs/triggers-design.md              | 338 +++
docs/triggers-implementation-plan.md | 267 +++
docs/triggers-prd.md                 | 196 +++
3 files changed, 801 insertions(+)
```

Affected: `Quality Gate / build`, `/ lint`, `/ deps`, `/ quality`, `Workflow Integration Tests`, `Workflow Runner Browser E2E`. `E2E Workflow Runner` is also failing on `main` at e5921935.

I'm not fixing it here. Regenerating the lockfile touches every dependency resolution in the repo and affects every open PR, so it belongs in its own change rather than bundled into a docs PR. I'll re-run these checks once `main` is green.

The relative links in the three docs resolve, so the markdown link check should pass on its own.

---

georgi:
Fixed in f05368f. My earlier diagnosis was incomplete — the drift is newer and more specific than I said.

CI builds a merge of this branch into `main`, and `main` moved after this branch was cut. #4472 bumped `electron/package.json` without regenerating the lockfile:

| Package | `electron/package.json` on main | `package-lock.json` on main |
|---|---|---|
| `electron-builder` | `26.15.3` | `26.8.1` |
| `js-yaml` | `^5.2.2` | `4.3.0` |
| `sharp` | `^0.35.1` | `0.34.5` / `0.35.0` |

`electron-builder@26.15.3` pulls transitives the old lockfile never recorded, which is where `bytestreamjs@2.0.1` and `pkijs@3.4.0` came from. `main` is broken on its own; every open PR inherits it.

Merged `main` in and regenerated with `npm install --package-lock-only`. `electron-builder` and `app-builder-lib` 26.15.3 now resolve under `electron/node_modules`, their transitives are recorded, and `sharp` resolves to 0.35.3 for electron and 0.35.0 for the node packages, each satisfying its declared range.

Verification, checking the test is sensitive rather than just green:

- `npm ci --dry-run` against **main's** lockfile → fails with the exact CI errors (`Missing: electron-builder@26.15.3`, `Invalid: lock file's sharp@0.35.0 does not satisfy sharp@0.35.3`, `Missing: bytestreamjs@2.0.1`)
- `npm ci --dry-run` against **this** lockfile → exit 0
- `npm run build:packages` → 58/58
- `web` and `electron` `tsc --noEmit` → clean
- `npm run lint` → clean

`mobile` typecheck is untested here: it is not a root workspace and `mobile/node_modules` is empty in this environment. CI installs it separately with `cd mobile && npm ci`, and nothing in this change touches `mobile/package-lock.json`.

Since the fix lives in a docs PR, it is worth a maintainer deciding whether to land it on `main` directly instead, so the other open PRs stop inheriting the failure.

---

PR #4494, opened 2026-07-26
Fix AtlasCloud "Cannot resolve image asset" on empty and non-URL image refs

georgi:
## The bug

Running **Seedance 2.0 — Image to Video** failed after ~2s with:

> Cannot resolve image asset for AtlasCloud — no usable uri or inline data

An asset input the user leaves empty does not arrive as `null`. The UI sends a
fully-shaped ref with every slot blank:

```json
{ "type": "image", "uri": "", "asset_id": null, "data": null, "metadata": null }
```

That object is truthy, so the factory's `v === undefined || v === null` skip
didn't catch it, and `resolveAssetForAtlas` ran out of resolution paths and
threw. Seedance's optional `last_image` triggers it on every run where it is
left unset.

## Changes (`packages/atlascloud-nodes/src/atlascloud-factory.ts`)

- **Blank-but-shaped refs resolve to `null`** — an absent optional input, so the
  field is simply omitted from the request.
- **Required asset inputs fail with a message naming the input**
  (`Seedance 2.0 — Image to Video: the image input "image" is empty — connect or
  upload an image`) instead of a resolver-internal one.
- **`data:` URIs pass through unchanged**, whether they arrive as a bare string,
  in `uri`, or in `data`.
- **Fall back to the canonical `loadMediaRefBytes`** for shapes the bespoke
  branches miss: `file://` URIs, absolute local paths, raw-RGBA in-flight images,
  and `asset_id` → `/api/storage/<id>.<ext>` storage candidates. This mirrors the
  same fix already made in `kie-nodes`.
- **Raw-RGBA images are PNG-encoded.** They previously matched the
  `data instanceof Uint8Array` branch and were shipped as raw pixel bytes
  labeled `image/x-raw-rgba` — a silent bad upload rather than an error.

### SSRF guard preserved

`loadMediaRefBytes` downloads http(s) unguarded, so the new fallback is skipped
for http(s) URIs. A *safe* public URL already passed through at the top of the
resolver, meaning anything still reaching that point is private/loopback/metadata
and stays behind `isSafeHttpUrl`. The existing SSRF test suite (loopback, RFC1918,
`169.254.169.254`, inet_aton numeric forms, IPv4-mapped IPv6) still passes
unchanged.

## Tests

Added to `packages/atlascloud-nodes/tests/atlascloud-factory.test.ts`:

- blank-but-shaped refs resolve to `null` without touching storage or
  `resolveAssetBytes`
- `data:` URI pass-through in all three positions
- `file://` and absolute-path refs resolve through the canonical resolver
- raw-RGBA input comes out as a real PNG (magic-byte assertion)
- end-to-end: empty optional `last_image` is omitted; empty required `image`
  throws the named error

`npm run test --workspace=packages/atlascloud-nodes` — 78 passed.
`npm run lint` and `npm run typecheck` pass (mobile typecheck fails only because
Expo deps aren't installed in this container; unrelated to this change).

---
_Generated by [Claude Code](https://claude.ai/code/session_01WzyppGjrnah7McMEUMZJv3)_

### DIFF
--- a/packages/atlascloud-nodes/src/atlascloud-factory.ts
+++ b/packages/atlascloud-nodes/src/atlascloud-factory.ts
@@ -16,9 +16,11 @@ import {
   registerDeclaredProperty
 } from "@nodetool-ai/node-sdk";
 import type { NodeClass, PropOptions } from "@nodetool-ai/node-sdk";
-import { mapPromptAssetsToInputs } from "@nodetool-ai/runtime";
+import { loadMediaRefBytes, mapPromptAssetsToInputs } from "@nodetool-ai/runtime";
 import type {
   AssetMediaKind,
+  MediaRefValue,
+  ProcessingContext,
   PromptAssetInputField,
   PromptAssetTextField
 } from "@nodetool-ai/runtime";
@@ -76,6 +78,13 @@ export interface AtlasManifestEntry {
 
 const ASSET_TYPES = new Set<AtlasFieldType>(["image", "video", "audio"]);
 const LIST_ASSET_RE = /^list\[(image|video|audio)\]$/;
+
+/**
+ * In-flight raw-RGBA images (protocol `RAW_RGBA_MIME`) carry straight-alpha
+ * pixels in `data`, not an encoded file. `loadMediaRefBytes` PNG-encodes them,
+ * so the outgoing data URI must say `image/png` rather than the ref's mime.
+ */
+const RAW_RGBA_MIME = "image/x-raw-rgba";
 
 type AssetRef = {
   uri?: string;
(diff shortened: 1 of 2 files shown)


georgi:
CI is red on every check, but none of it is this diff: all jobs die in `npm ci` before a single test runs, because `package-lock.json` is out of sync with `package.json` on `main`.

```
npm error Invalid: lock file's sharp@0.35.0 does not satisfy sharp@0.35.3
npm error Invalid: lock file's js-yaml@5.2.1 does not satisfy js-yaml@5.2.2
npm error Missing: electron-builder@26.15.3 from lock file
… (~60 more)
```

Same failure on `main` at e592193 — see [E2E Workflow Runner run 30196265278](https://github.com/nodetool-ai/nodetool/actions/runs/30196265278), which predates this PR. Reproduced locally too: `npm ci --dry-run` fails identically on a tree whose only diff from `main` is this PR's two files. The recent Dependabot bumps (sharp, js-yaml, electron-builder) updated `package.json` without a matching lockfile refresh.

This PR touches only `packages/atlascloud-nodes`; its own suite passes locally (78 tests), as do `npm run lint` and `npm run typecheck` for web/electron/packages. I'll re-check once `main` has a regenerated lockfile and merge it in so CI re-runs against a green base.

---

PR #4501, opened 2026-07-26
A trigger that keeps failing now stops itself

georgi:
Follow-up to #4499, closing four of the six open items it listed.

## A broken trigger fired forever

`trigger_registrations` had no counters, so a workflow that threw on every run kept firing on schedule all night. The PRD calls this a feature requirement, not a polish item, and its success metric is "zero registrations in the wild that have failed more than 5 times consecutively and are still enabled".

The table gains `consecutive_failures`, `run_count`, `disabled_reason`, `expires_at` and `max_runs` (additive migration, mirrored in `schema/`, `schema-pg/` and the bootstrap DDL). The dispatcher settles every delivered run against them: five consecutive failures disarms with `disabled_reason = "failures"`, a passed `expires_at` or a spent `max_runs` disarms before the run starts rather than after spending one more.

`enabled` stays the single armed bit. `disabled_reason` is what distinguishes a trigger the dispatcher gave up on from one a person switched off — the two were the same state before, which is why an auto-stop was invisible. Re-arming through `jobs.triggerStart` clears the reason *and* the failure count; without that, a registration disabled after five failures would disable itself again one run into its second life.

The rules live in `triggers/settle.ts` as functions over a registration row, not inside the dispatch path, so their tests are a table rather than a dispatch harness.

## The auto-stop was invisible in the workflow list

`jobs.triggersRunning` filtered to `enabled === 1`. A trigger that stopped on its own therefore disappeared from the response, and the bolt badge on its workflow card disappeared with it — the failure surfaced as an absence. It now also returns disarmed-by-the-dispatcher rows, with `enabled` telling the two apart.

Both surfaces explain the stop and how to undo it: the editor popover reads "Stopped" with *"Disabled after 5 consecutive failures. Turn it back on to retry."*, the toolbar button's accessible name becomes "Triggers: inactive, disabled automatically", and the mobile screen sorts stopped triggers above merely-failing ones, since the second is still running and the first has silently stopped automating. The sentence itself comes from one shared `describeTriggerDisabledReason`, so the editor, the phone and the log describe the same stop the same way.

## Triggered runs know whether they are the first (F5)

Every dispatch now passes `is_first_run`, `last_fired_at` and `now` as run params. `run_count` advances on success only, so a first run that fails sees `is_first_run` again on its retry — which is what a backfill-seeding graph wants, and is the semantic the design doc specifies. A workflow that declares no matching inputs is unaffected.

## file_watch wrote its fire time twice

The dispatcher stamped `last_fired_at` for every kind except `schedule`, but the file watcher had already stamped it at delivery — so a `file_watch` event cost a second `save()` and overwrote the sweep's timestamp with a slightly later one. The rule is now stated positively: a kind whose in-process ingestion adapter stamps the column owns it (`schedule`, `file_watch`); the dispatcher stamps the kinds that arrive from outside (`webhook`, `manual`).

## One trigger-kind map instead of two

`TRIGGER_KIND_BY_NODE_TYPE` moves to `@nodetool-ai/protocol`, which `web/` and `packages/websocket` both already depend on. That left `web/src/utils/triggerNodeTypes.ts` as a pure re-export plus `graphHasTriggerNodes`, which had no caller anywhere — both are deleted, and the tests move to `packages/protocol/tests/triggers.test.ts`.

## mobile/src/api.ts

Dropped the three `/api/jobs/triggers/*` paths, their operations, and the `TriggerWorkflow*` schemas. They described a REST contract keyed by workflow id that the tRPC procedures replaced and the server no longer serves; only `components["schemas"]` is consumed anywhere, so nothing referenced them. The file header now states the rule that kept them around — there is no regen script, so a moved endpoint has to be deleted here in the same change.

## Verification

`npm run typecheck`, `npm run lint` (design-token lint included) and `npm run test` pass: web 1010 suites / 12053 tests, electron 63 / 655, mobile 64 / 898. Backend suites: websocket 118 / 1589, models 43 / 746, protocol 33 / 617.

`npm run check` also runs the package test suites, where `@nodetool-ai/image-nodes` fails 52 tests with `No WebGPU adapter available (Node/Dawn)`. That is the sandbox having no GPU, unrelated to this change; every other package task passes.

## Still open from #4499

- **No Triggers tab.** The popover remains the only surface, and discoverability is unchanged — the bolt button still appears only once a trigger node is on the canvas.
- **O3/O4 — `max_usd` and per-registration cost attribution.** Not attempted: the budget check needs cost records stamped with `registration_id`, which means threading the id through the enqueue path into the cost ledger. `expires_at` and `max_runs` from F8 are enforced here, but nothing writes them yet — there is no editor field and no update procedure, so today they are settable only by a direct write. Enforcing them now means the later setter is a UI change, not a schema change.
- **First-run *question*.** F5's params landed; §8's dialog did not. None of the four shipped kinds polls a backlog — `file_watch` catch-up replays only from a stored cursor — so the "process existing items or only new ones" prompt has nothing to seed until a polling kind exists.
- **Tags** (F10) still have no schema.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WhSaPauTUMqw5E8NLuEp5q)_


georgi:
CI is red on `Quality Gate / test-packages`, and I don't think it's this branch. The `quality` leg is just the aggregator reporting that one, so there is a single real failure:

```
FAIL  packages/video-nodes/tests/model3d-render.test.ts
Error: connect ECONNREFUSED 127.0.0.1:36711
```

That test launches headless Chromium and drives it over CDP. Nothing in this branch touches `video-nodes`, 3D rendering, or any local server — the diff is trigger registrations, the dispatcher, and the two trigger UIs. `main` at `9c8d6e8` (the commit merged in here) went green on the same suite, and the same run's cleanup step reported orphan `chrome` processes, which fits a loaded runner rather than a code change.

It also looks like a known flake with a partial fix already in place: `connectCdp` in `render3d-headless.ts` polls `/json/list` for 10s specifically because, per its own comment, *"it turned up as an `ECONNREFUSED` on CI while several other Chrome-driving suites were running"*. Worth noting the error here is a **bare** `ECONNREFUSED`, not `connectCdp`'s wrapped `"could not attach to headless Chrome on port … within 10000ms"`, so it is arriving from a different point in the launch/attach sequence than the one that retry loop covers — raising that budget would not help. I'd rather not guess at a fix for a runner-only failure I can't reproduce than weaken the assertion that is the only thing catching a real render break.

Locally the suite passes: `packages/video-nodes` 18 files / 171 tests, and `model3d-render.test.ts` on its own 3× in a row, 7/7 each time.

Everything this branch does own is green locally after the `main` merge — typecheck, lint, and `npm run test` (web 1013 suites / 12075 tests, electron 63 / 655, mobile 66 / 913), plus websocket 119 / 1595, models 43 / 751, protocol 33 / 617.

Re-running the failed job needs Actions write, which the token here doesn't have (`403 Resource not accessible by integration` on `rerun-failed-jobs`). Could someone re-run `test-packages`? If it fails a second time on the same test I'll dig into the launch path properly instead of assuming contention.

---

PR #4520, opened 2026-07-27
security: close unauthenticated reads and scope collections to their owner

georgi:
Security audit of the API (`packages/websocket`, `packages/auth`) and `web/`. Three confirmed access-control vulnerabilities, plus hardening. Findings I did **not** fix are listed at the bottom.

## 1. Private workflows readable without authentication (high)

The auth-exemption allowlist in the server's `onRequest` hook exempted `/api/workflows` (list), `/api/workflows/:id`, `/api/workflows/names`, `/api/workflows/tools`, and `:id/dsl-export`. The REST handlers resolve identity from the server-set `x-user-id` header and fall back to user `"1"` when it is absent (`getUserId`, `http-api.ts:366`), so those paths served the local user's library to anyone who could reach the port — defeating the `Remote access requires authentication` 401 that local mode otherwise applies to a server bound beyond loopback.

Verified against a running server with `NODETOOL_TRUST_LOCALHOST=false` and a seeded private workflow:

```
$ curl -s http://127.0.0.1:7788/api/workflows/wf-secret
{"id":"wf-secret","access":"private","name":"Prod deploy pipeline",
 "graph":{"nodes":[{"id":"1","type":"nodetool.constant.String",
 "data":{"value":"sk-live-SUPERSECRET"}}],"edges":[]}, ...}
```

The full graph is disclosed, so anything a user parks in a node property — prompts, internal endpoints, keys pasted into a constant — goes with it.

## 2. OAuth credential endpoints readable and mutable without authentication (high)

`pathname.startsWith("/api/oauth/")` exempted the entire prefix, but only the two browser redirect targets need it — the identity provider sends the user back with no `Authorization` header, and those handlers authenticate through the single-use PKCE `state` minted at `/start`. Every other path operates on stored credentials for the caller's user id:

```
$ curl -s http://127.0.0.1:7788/api/oauth/hf/tokens
{"tokens":[{"id":"oc1","provider":"huggingface","account_id":"acct-123",
 "username":"victim-hf-user","token_type":"Bearer","scope":"read-repos", ...}]}
```

Unauthenticated, that also reached `/hf/whoami` and `/github/user` (linked third-party profile and email), `/hf/refresh` (forced token refresh), and `/openai/disconnect` (credential deletion). Token *values* were never returned — they stay encrypted at rest and the handlers only expose metadata.

## 3. Vector collections were not scoped to a user (high, multi-user only)

Collections are a flat global namespace in every `VectorProvider` (sqlite-vec, Pinecone, Supabase, Chroma) — the interface has no concept of a user, and neither the tRPC router nor the REST index endpoint looked at `ctx.userId`. In multi-user (Supabase) mode any authenticated user could list, rename, delete, and index into any other user's collection, and see its document count.

Ownership is now recorded in collection metadata (`owner_user_id`, stamped on create) and enforced at the API boundary. Teaching four providers about users is a much larger change; this covers the surface an untrusted caller can actually reach. In-process consumers — RAG nodes, agent tools, the CLI — talk to the provider directly and are unaffected, since they already run with their invoker's privileges.

Two judgment calls worth review:

- **Pre-existing collections stay shared.** They carry no owner, so everyone can still use them, and an update does not claim an unowned collection for the first editor. Retroactively assigning them would lock existing deployments out of their own data, and claiming-on-edit would silently lock out everyone who had been sharing one.
- **Someone else's collection answers NOT_FOUND, not FORBIDDEN**, so the error can't be used to enumerate other users' collection names.

Client metadata is stripped of server-owned keys before merging, so a caller can't rewrite `owner_user_id` to seize or hand off a collection.

## 4. Missing security response headers (medium)

The web app's CSP lives in a `<meta>` tag, and browsers ignore `frame-ancestors` there — it is only honoured in a header — so the app had no clickjacking protection, and no `nosniff` or `Referrer-Policy` anywhere.

## 5. Unsandboxed document-preview iframe (medium)

`PropertyDropzone` rendered `.html`/`.txt` asset bytes in an iframe with no `sandbox` attribute, from the app's own origin. It was the only HTML iframe in the app missing one. Currently mitigated by `storage-api.ts` serving `.html` as `text/plain`; this adds the second lock.

## Other hardening in the collections surface

- Collection names are validated (no path separators, control characters, surrounding whitespace, or >128 chars). Deliberately permissive so existing names keep working.
- A duplicate name returns `ALREADY_EXISTS` instead of leaking the store's `UNIQUE constraint failed: vec_collections.name` text, and a rename onto a taken name is refused up front.
- The index endpoint caps uploads at `NODETOOL_MAX_UPLOAD_BYTES` — it reads the whole file into a string and chunks it in memory, so Fastify's 100 MB body limit was the only bound.
- Provider errors are logged and answered with a generic message rather than echoed; they carry SQL text, file paths, and upstream URLs.

## Changes

- `packages/websocket/src/lib/public-routes.ts` (new) — both auth-exemption predicates, extracted so they are testable without booting the server (`server.ts` opens the database and binds a port at import). The prefix matches also gained a trailing separator, so a workflow id like `publications` no longer matches `/api/workflows/public`.
- `packages/websocket/src/lib/collection-access.ts` (new) — ownership and name rules for collections.
- `packages/websocket/src/trpc/routers/collections.ts`, `src/collection-api.ts` — enforce ownership, validate names, cap uploads, stop echoing provider errors.
- `packages/websocket/src/server.ts` — use the narrowed predicates; add `X-Content-Type-Options: nosniff`, `X-Frame-Options: SAMEORIGIN`, `Content-Security-Policy: frame-ancestors 'self'`, and `Referrer-Policy: strict-origin-when-cross-origin`. `/api/storage/*` is exempt from `nosniff` and the frame headers to preserve its documented cross-origin embedding contract (MCP App iframes, Electron renderer) and the `application/octet-stream` fallback that ``/`<video>` sniff.
- `web/src/components/properties/PropertyDropzone.tsx` — `sandbox=""` on the text/HTML preview iframe. The PDF iframe is left alone; Chrome's PDF viewer does not render inside an empty sandbox.
- Tests: `public-routes.test.ts` and `collection-access.test.ts` (new), plus ownership/validation cases added to the collections router and REST suites.

## Verification

Auth-exemption probes re-run against the rebuilt server:

| path | before | after |
|---|---|---|
| `/api/workflows`, `/api/workflows/:id`, `/names`, `/tools`, `:id/dsl-export` | 200 | 401 |
| `/api/oauth/{hf,github,openai}/tokens`, `/hf/whoami` | 200 | 401 |
| `/health`, `/api/config`, `/api/nodes/metadata` | 200 | 200 |
| `/api/workflows/public`, `/api/workflows/examples` | 200 | 200 |
| `/api/oauth/{hf,github}/callback` | 200 | 200 |

Collection isolation was checked end-to-end against the **real sqlite-vec provider, not mocks** — the scheme depends on metadata surviving the round trip:

```
PASS  owner persisted in store: alice
PASS  owner present in listCollections: alice
PASS  bob's listing is empty: 0
PASS  bob update / rename / delete blocked: NOT_FOUND
PASS  owner unchanged after metadata rewrite attempt: alice
PASS  duplicate name reports CONFLICT
PASS  duplicate error hides SQL text: false
```

`npm run lint` passes (remaining warnings pre-existing). `npx vitest run --root packages/websocket` — 124 files, 1653 tests passing. `web` PropertyDropzone suite passing. `npm run typecheck` passes for `packages` and `web`; `mobile` fails on missing modules because its dependency tree is not installed in this environment — pre-existing and unrelated.

## Reported, not fixed

- **`POST /api/kie/webhook` has no signature or shared-secret check.** Anyone who learns or guesses a `taskId` can resolve or reject an in-flight KIE generation with an attacker-controlled payload. The trigger webhook route (`triggers/webhook-route.ts`) does this correctly with a sha256 hash and `timingSafeEqual`. A fix means threading a secret through the `callBackUrl` built in `kie-base.ts` / `kie-provider.ts`, which I cannot test against the real KIE callback, so I left it rather than ship an untested change to a third-party integration.
- **`/api/storage/*` keys are not user-scoped.** Any authenticated user can `GET` or `PUT` any key, so in multi-user mode one tenant can overwrite another's asset bytes. Fixing this changes the asset URL scheme, so it needs its own migration.
- **`GET /api/files/local` accepts any absolute path**, guarded only by a small denylist of sensitive home directories — while the tRPC `files.list` equivalent is sandboxed to `$HOME`. It requires auth and is disabled when `NODETOOL_ENV=production`, but the two file-read surfaces disagree on how much of the disk is reachable.
- **Collection scoping covers the HTTP surface only.** A workflow node or agent tool that queries collections by name still reaches any collection in the store, because it bypasses this layer. Closing that means per-user scoping inside the providers.

## CI note

The CodeQL check reports one `js/missing-rate-limiting` alert on the auth hook in `server.ts`. It is a pre-existing false positive re-fingerprinted by this PR's line shifts — `@fastify/rate-limit` is registered globally *before* that hook and CodeQL has no model for it. Details in [this comment](https://github.com/nodetool-ai/nodetool/pull/4520#issuecomment-5088638578).

https://claude.ai/code/session_01HznMLQE3frSF6xnRjk9gQB

### DIFF
--- a/packages/websocket/src/collection-api.ts
+++ b/packages/websocket/src/collection-api.ts
@@ -6,13 +6,18 @@
  * moved to the tRPC `collections` router.
  */
 
+import { createLogger } from "@nodetool-ai/config";
+import { getMaxUploadBytes } from "@nodetool-ai/storage";
 import {
   getDefaultVectorProvider,
   CollectionNotFoundError,
   splitDocument
 } from "@nodetool-ai/vectorstore";
-import type { HttpApiOptions } from "./http-api.js";
+import { getUserId, type HttpApiOptions } from "./http-api.js";
 import { notifyResourceChange } from "./resource-events.js";
+import { canAccessCollection } from "./lib/collection-access.js";
+
+const log = createLogger("nodetool.websocket.collection-api");
 
 function jsonResponse(data: unknown, status = 200): Response {
   return new Response(JSON.stringify(data), {
@@ -39,7 +44,7 @@ function normalizePath(pathname: string): string {
 export async function handleCollectionRequest(
   request: Request,
   pathname: string,
-  _options: HttpApiOptions
+  options: HttpApiOptions
 ): Promise<Response | null> {
   pathname = normalizePath(pathname);
 
@@ -55,15 +60,41 @@ export async function handleCollectionRequest(
     return errorResponse(400, "Expected multipart/form-data");
   }
 
+  const collectionName = decodeURIComponent(indexMatch[1]);
+  const userId = getUserId(request, options.userIdHeader ?? "x-user-id");
+
   try {
     const provider = getDefaultVectorProvider();
-    const collectionName = decodeURIComponent(indexMatch[1]);
     const collection = await provider.getCollection({ name: collectionName });
+
+    // Same ownership rule the tRPC router enforces, and the same 404-not-403
+    // response for someone else's collection so this endpoint can't be used to
+    // probe for names. See lib/collection-access.ts.
+    if (
+      !canAccessCollection(
+        collection.metadata as Record<string, string | number | boolean>,
+        userId
+      )
+    ) {
+      return errorResponse(404, "Collection not found");
+    }
 
     const formData = await request.formData();
     const file = formData.get("file");
     if (!file || !(file instanceof File)) {
       return errorResponse(400, "No file provided");
+    }
+
+    // The whole file is read into a string and chunked in memory, so cap it at
+    // the same limit storage uploads use rather than relying on Fastify's
+    // 100 MB body limit.
+    const max = getMaxUploadBytes();
+    if (file.size > max) {
+      return errorResponse(
+        413,
+        `Upload exceeds maximum size: ${file.size} > ${max} bytes ` +
+          `(set NODETOOL_MAX_UPLOAD_BYTES to raise the limit)`
+      );
     }
 
     const text = await file.text();
@@ -96,7 +127,12 @@ export async function handleCollectionRequest(
     if (err instanceof CollectionNotFoundError) {
       return errorResponse(404, "Collection not found");
     }
-    const msg = err instanceof Error ? err.message : String(err);
-    return errorResponse(500, `Vector store error: ${msg}`);
+    // Log the detail, return a generic message: provider errors carry SQL
+    // text, file paths, and upstream URLs that shouldn't reach the client.
+    log.error("Collection index failed", {
+      collection: collectionName,
+      error: err instanceof Error ? err.message : String(err)
+    });
+    return errorResponse(500, "Vector store error");
   }
 }
--- a/packages/websocket/src/lib/collection-access.ts
+++ b/packages/websocket/src/lib/collection-access.ts
@@ -1,0 +1,108 @@
+/**
+ * Ownership and input rules for vector collections exposed over HTTP.
+ *
+ * The `VectorProvider` interface has no concept of a user — collections are a
+ * flat, global namespace shared by every caller. That is fine for the local
+ * single-user case, but in multi-user (Supabase) mode it means any
+ * authenticated user can list, read, rename, delete, and index into any other
+ * user's collection. Teaching every provider (sqlite-vec, Pinecone, Supabase,
+ * Chroma) about users is a much larger change, so ownership is recorded in
+ * collection metadata and enforced here, at the API boundary.
+ *
+ * In-process consumers — RAG nodes, agent tools, the CLI — talk to the
+ * provider directly and are deliberately unaffected: they already run with the
+ * privileges of whatever invoked them.
+ *
+ * Collections that predate this (no owner recorded) stay readable and writable
+ * by everyone. Retroactively assigning them to a user would lock existing
+ * deployments out of their own data, so they are treated as shared. Only
+ * collections created through the API from here on are isolated.
+ */
+
+/** Metadata key holding the id of the user who created the collection. */
+export const OWNER_METADATA_KEY = "owner_user_id";
+
+/**
+ * Metadata keys the server owns. A client that could set these through
+ * `collections.update` would be able to hand itself another user's collection,
+ * so they are stripped from client input and re-applied from server state.
+ */
+export const RESERVED_COLLECTION_METADATA_KEYS: readonly string[] = [
+  OWNER_METADATA_KEY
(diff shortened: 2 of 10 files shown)


georgi:
## CodeQL failure: pre-existing false positive, not fixing

The failing check is one alert — `js/missing-rate-limiting` ("This route handler performs authorization, but is not rate-limited"), high severity, on `packages/websocket/src/server.ts`, attributed to the auth `onRequest` hook.

**It is not introduced by this PR.** The diff on `server.ts` (`git diff origin/main...HEAD`) touches exactly three things: the import block, the `onSend` header hook, and one line in the public-path allowlist (`pathname.startsWith("/api/oauth/")` → `isPublicOAuthRequest(pathname)`). The auth hook's body and the rate-limiter registration are untouched. What changed is the hook's *line offset* — removing the two predicate helpers near the top of the file (−26 lines) and adding the import (+4) and the `onSend` block (+16) shifted it, which broke CodeQL's line-hash `partialFingerprints` and got a baseline alert re-reported as new.

**It is also not substantively true.** `@fastify/rate-limit` is registered with `global: true` at `server.ts:697`, *before* the auth hook, precisely so floods are rejected with 429 before any token verification runs — the existing comment there says as much. It is on by default (`getHttpRateLimitConfig()` in `lib/http-rate-limit.ts` reads `enabled: !parseBoolEnv("NODETOOL_RATE_LIMIT_DISABLED")`, 1000 req/min per IP), and a running server confirms it: `Per-IP HTTP rate limiting enabled {"max":1000,"timeWindowMs":60000}`. CodeQL's rate-limiting query models `express-rate-limit` and friends; it has no model for a globally-registered `@fastify/rate-limit` plugin, so it cannot see the limiter.

I'm not pushing a change for this. The options are to dismiss the alert in the code-scanning UI (as "false positive"), which is a maintainer action I shouldn't take unilaterally, or to add an inline suppression comment — which would bury a policy decision in the source. Neither belongs in a security PR that is otherwise about removing auth bypasses.

Worth noting the one real caveat while we're here, independent of this PR: `isRateLimitExempt` exempts loopback, so the limiter does not apply to localhost. That is deliberate (it mirrors the auth hook's localhost posture so Electron/local dev is never throttled) and is the right call for the default deployment, but it does mean a reverse-proxied setup that is *not* on the trusted-proxy list gets no per-IP limiting, since every request arrives from loopback. That predates this PR too.

---

PR #4576, opened 2026-07-29
perf(websocket): trim authoritative run framing

heavy-d:
improve latency


georgi (dismissed):
The optimization to skip the eager `job_update: running` frame for `requireTerminalResult: true` jobs introduces a protocol regression.

**The issue:** In `packages/kernel/src/runner.ts`, the kernel's "running" `job_update` is emitted at lines 584-589, *after* `graph.validate()` and correlation analysis succeed (lines 557-581). If those checks throw (`GraphValidationError`), execution jumps straight to the `catch` block and emits only a `failed` update — no `running` message is ever sent.

Previously, the WebSocket runner's own eager "running" frame covered this gap for every client. With this change, for `requireTerminalResult: true` clients (which is the **SDK default** per `packages/protocol/src/api-schemas/sdk-lifecycle-v1.ts:281`), a graph-validation failure means the client gets no running acknowledgement at all — just a terminal failure. This is a real behavioral change to protocol framing guarantees, not just a "duplicate frame" trim.

**What's needed to unblock:**
1. Either emit the running frame before validation in the kernel, or add a fallback in the websocket runner for the early-failure case.
2. A test covering the early-validation-failure path for `requireTerminalResult: true` jobs.
3. A PR description that explains the actual change and its scope.

**Note on CI:** The `test-packages` failure (`application-budget-gate.test.ts:342`) is pre-existing on main — same test, same assertion. Not caused by this PR.

---

PR #4577, opened 2026-07-29
Fix comfy.* bridge request shapes to match the Python worker

georgi:
Found while verifying the worker protocol from both sides — the TS bridge in `packages/runtime` against `nodetool-core`'s `src/nodetool/worker/`.

Everything outside `comfy.*` checks out: framing, version negotiation, `discover`/`worker.status`, `execute`/`execute.stream`/`cancel`, and all `provider.*` and `models.*` shapes agree field-for-field. The `comfy.*` family does not — the bridge sends field names the worker never reads, so the worker's own argument validation rejects the requests.

`nodetool-core`'s `docs/comfy-proxy.md` is the authoritative wire reference (this repo's `python-bridge-types.ts` says as much), and `comfy_handler.py` matches it. This side was the one that drifted.

## Request-shape fixes

| Message | Bridge sent | Worker reads | Result before this PR |
|---|---|---|---|
| `comfy.execute` | `prompt` | `workflow` | `comfy.execute requires a non-empty 'workflow'` |
| `comfy.upload` | `blob` | `data` | `comfy.upload requires binary 'data'` |
| `comfy.models.download` | flat `{folder, url\|repo_id, filename}` | nested `source: {type, ...}` | `Unknown model source type: None` |

`comfy.execute` is the only one of these with a production caller (`ComfyWorkerWorkflowNode` in `packages/integration-nodes`), so **every ComfyUI worker run was failing**. The other two are latent — the bridge exposes them but nothing calls them yet.

Two more on the same family:

- `comfy.models.list` has no `folder` filter on the wire; the worker always returns the whole volume. The bridge now sends an empty payload and narrows client-side, so the documented filter actually applies instead of being silently dropped.
- `comfy.object_info` returns `{object_info: ...}`. Unwrapped, so the method returns the catalog its contract promises.

## Response-type fixes

These only produced `undefined` at the read site rather than throwing, because the interfaces carry index signatures:

- `ComfyEvent`: `node_output` carries `outputs` (not `output`); `preview` carries `format` + `image` (not `mime_type` + `blob`). Added `queue_position` for `queued`.
- `ComfyModelInfo`: entries carry `size_bytes`, not `size`.
- `ComfyModelDownloadUpdate`: added the `"exists"` status the worker emits on the already-downloaded fast path.
- `ComfyStatusInfo`: added the `system_stats` / `queue_remaining` / `error` fields `comfy.status` returns.
- `ComfyExecuteResult`: added `status`, which distinguishes a completed run from a worker-side cancellation.

Also exposed `include_temp` as `ComfyExecuteOptions.includeTemp` — the worker supports it, but there was no way to ask for it.

## Why the tests didn't catch this

The fake worker in `python-websocket-bridge.test-helpers.ts` answers any payload it is sent, so all three request-shape bugs passed a green suite while failing against a real worker. It now applies the same argument validation `comfy_handler.py` does and answers a matching `error` frame when a request is malformed, so a shape regression fails the suite. Its `node_output` and `comfy.models.list` responses were also emitting the wrong field names; those now mirror the worker.

## Verification

- `npx vitest run packages/runtime packages/protocol packages/integration-nodes` → 3401 passed (180 files)
- `npx tsc --build tsconfig.build.json` → clean
- `npx oxlint packages/runtime/src` → clean
- `uv run pytest tests/worker -q` in `nodetool-core` → 127 passed (unchanged; no Python edits here)

No protocol version bump: this is a client-side correction to v3, not a wire change, so `BRIDGE_PROTOCOL_VERSION` and `MIN_BRIDGE_PROTOCOL_VERSION` stay put.

https://claude.ai/code/session_01VVSMgubQKhwLLoERBBJuhx

---
_Generated by [Claude Code](https://claude.ai/code/session_01VVSMgubQKhwLLoERBBJuhx)_

### DIFF
--- a/packages/runtime/src/python-bridge-base.ts
+++ b/packages/runtime/src/python-bridge-base.ts
@@ -1072,14 +1072,17 @@ export abstract class PythonBridgeBase
    * would hang and the pending maps would leak.
    */
   comfyExecute(
-    prompt: Record<string, unknown>,
+    workflow: Record<string, unknown>,
     options: ComfyExecuteOptions = {},
     onEvent?: (event: ComfyEvent) => void,
     requestId: string = randomUUID()
   ): Promise<ComfyExecuteResult> {
-    const data: Record<string, unknown> = { prompt };
+    // The worker's field is `workflow`, not `prompt` — see the comfy.execute
+    // request schema in nodetool-core's docs/comfy-proxy.md.
+    const data: Record<string, unknown> = { workflow };
     if (options.blobs) data.blobs = options.blobs;
     if (options.previews) data.previews = true;
+    if (options.includeTemp) data.include_temp = true;
     if (typeof options.timeout === "number") data.timeout = options.timeout;
 
     return new Promise<ComfyExecuteResult>((resolve, reject) => {
@@ -1150,9 +1153,10 @@ export abstract class PythonBridgeBase
     bytes: Uint8Array,
     options: Record<string, unknown> = {}
   ): Promise<Record<string, unknown>> {
+    // The worker reads the payload from `data`, not `blob`.
     return this._providerCall("comfy.upload", {
       filename,
-      blob: bytes,
+      data: bytes,
       ...options
     });
   }
@@ -1165,7 +1169,12 @@ export abstract class PythonBridgeBase
   }
 
   async comfyObjectInfo(): Promise<Record<string, unknown>> {
-    return this._providerCall("comfy.object_info", {});
+    // The worker wraps the catalog as `{object_info: {...}}`; unwrap so callers
+    // get the catalog itself, as this method's contract promises.
+    const result = await this._providerCall("comfy.object_info", {});
+    return (
+      (result.object_info as Record<string, unknown> | undefined) ?? result
+    );
   }
 
   async comfySystemStats(): Promise<Record<string, unknown>> {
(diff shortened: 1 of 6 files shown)


georgi:
CI is red on `test-packages` (and therefore on `quality`, which is just the aggregator leg), but the failure is pre-existing on `main` and unrelated to this diff.

The failing test is `packages/websocket/tests/application-budget-gate.test.ts:342` — "closes the ledger row out at what the run actually cost":

```
AssertionError: expected { …(10) } to match object { Object (status, actualUsd) }
  {
    "actualUsd": 0.42,
-   "status": "completed",
+   "status": "failed",
  }
```

The same failure — same file, same line, same assertion, same `1 failed | 1939 passed (1940)` tally — is on the most recent `main` run, [30410868598](https://github.com/nodetool-ai/nodetool/actions/runs/30410868598) at `05995cc5`, about six hours before this branch was pushed.

It also can't plausibly be mine: `packages/websocket` contains no reference to `comfy` at all, and this PR only touches `packages/runtime`'s comfy bridge surface and its tests. Running the test in isolation on this branch passes 18/18, which points at cross-test state rather than a code defect — the run log shows job id `job-settle` reported `completed`, `completed`, then `failed` within the same file, so the ledger row being asserted on is likely one a neighbouring test left behind.

Not fixing it here — it's a separate flake in a package this PR doesn't touch, and folding a billing-ledger fix into a protocol PR would muddy both. Happy to open a separate issue or PR for it if that's useful. I'll re-run CI here once `main` goes green.

The checks this PR is actually responsible for are green: `packages/runtime`, `packages/protocol`, and `packages/integration-nodes` pass 3401 tests across 180 files, with `tsc --build` and `oxlint` clean.

---

georgi:
Reviewed — all field-name fixes verified against the PR claims. The test helper's new worker-contract validation is a good addition that would have caught the original bug. CI `test-packages` failure is pre-existing on main (same `application-budget-gate.test.ts:342` failure in `packages/websocket`, unrelated to this PR's `packages/runtime` changes). Merging.

---

PR #4585, opened 2026-07-29
docs: document the ComfyUI integration

georgi:
Validated the ComfyUI integration end to end and gave it a docs page. Docs-only change — no source files touched.

## Validation

Everything on the TypeScript side works and is covered:

| Suite | Result |
|---|---|
| `packages/runtime` — `comfy-executor`, `python-bridge-comfy`, `python-bridge-base-coverage`, `swappable-python-bridge` | 100 passed |
| `packages/integration-nodes` — `comfy.test.ts` | 10 passed |
| `web` — `comfyDynamicSchema`, `dynamicSlots` | 39 passed |
| `npm run lint` / `npm run typecheck` | pass |

Traced the whole path: the browser schema parser (`web/src/utils/comfyDynamicSchema.ts`) → the loader UI → the two nodes (`packages/integration-nodes/src/nodes/comfy.ts`) → the direct executor (`packages/runtime/src/comfy-executor.ts`) and the `comfy.*` bridge family (`packages/runtime/src/python-bridge-base.ts`), including the `SwappableBridge` passthrough. The request shapes match the fake worker's contract checks added in #4577, and `supportsComfy()` correctly gates on protocol v3 **plus** `worker.status.comfy.enabled`.

## Gaps found

Recorded in the new page's "Known limitations" section rather than fixed, since the fixes depend on the Python worker's actual conventions:

- **The worker node has no workflow loader.** `DYNAMIC_COMFY_NODE_TYPE` is a single string (`lib.comfy.RunWorkflow`), so `lib.comfy.RunWorkflowOnWorker` falls back to the default node renderer. It declares `supportsDynamicInputs`/`supportsDynamicOutputs`, but there is no UI to populate them — handles have to be added by hand.
- **Output-slot naming diverges.** The direct node emits `<comfyNodeId>:<kind>`, matching what the schema parser declares. The worker node names slots after the keys in the worker's `result.blobs` (the test fixture uses `9_image`), so wiring the loader to it as-is would produce slots that never receive a value.
- **The worker node returns outputs buffered**, while the direct node streams one frame per file.
- **`include_temp` is unreachable from a node.** `ComfyExecuteOptions.includeTemp` exists on the bridge but no node property surfaces it.

## Changes

- **`docs/comfyui.md`** (new) — both nodes side by side, properties, the API-format requirement and the three load paths (paste / `.json` / ComfyUI-exported `.png`), the curated `Load*`/`Save*` class maps and their prefix fallbacks, the `<comfyNodeId>:<field>` handle convention, HTTP upload vs. bridge-blob input handling, the full `comfy.*` bridge table (including the `workflow`-not-`prompt` and nested-`source` shapes that are easy to get wrong), known limitations, a troubleshooting table keyed on errors the code actually raises, and a source map.
- Linked from `docs/_includes/sidebar.html` (Models & Providers), `docs/llms.txt`, `docs/worker-deployment.md`, and `docs/python-bridge-protocol.md`.
- `packages/integration-nodes/README.md` — promoted the ComfyUI nodes out of the "Other" catch-all into their own entry.

https://claude.ai/code/session_01WRKgD9Um6KPuTQ9QsribeA

---
_Generated by [Claude Code](https://claude.ai/code/session_01WRKgD9Um6KPuTQ9QsribeA)_

### DIFF
--- a/docs/_includes/sidebar.html
+++ b/docs/_includes/sidebar.html
@@ -97,6 +97,7 @@
           <li><a href="{{ '/models' | relative_url }}" class="sidebar-link" data-nav-path="/models">Local vs Cloud</a></li>
           <li><a href="{{ '/providers' | relative_url }}" class="sidebar-link" data-nav-path="/providers">Provider Reference</a></li>
           <li><a href="{{ '/huggingface' | relative_url }}" class="sidebar-link" data-nav-path="/huggingface">HuggingFace</a></li>
+          <li><a href="{{ '/comfyui' | relative_url }}" class="sidebar-link" data-nav-path="/comfyui">ComfyUI</a></li>
           <li><a href="{{ '/models-manager' | relative_url }}" class="sidebar-link" data-nav-path="/models-manager">Models Manager</a></li>
         </ul>
       </details>
--- a/docs/comfyui.md
+++ b/docs/comfyui.md
@@ -1,0 +1,256 @@
+---
+layout: page
+title: "ComfyUI"
+description: "Run ComfyUI workflows inside NodeTool — directly against a ComfyUI server, or proxied over a NodeTool GPU worker."
+---
+
+NodeTool runs ComfyUI workflows as nodes in a NodeTool graph. You export a
+workflow from ComfyUI in **API (prompt) format**, load it into a node, and its
+`Load*` nodes become typed inputs while its `Save*` nodes become typed outputs.
+Everything around it — asset handling, LLM steps, mini apps — stays NodeTool's.
+
+Two nodes cover two topologies:
+
+| | `lib.comfy.RunWorkflow` | `lib.comfy.RunWorkflowOnWorker` |
+|---|---|---|
+| Title | Run ComfyUI Workflow | Run ComfyUI Workflow (Worker) |
+| Talks to | Any ComfyUI HTTP/WebSocket endpoint | A NodeTool worker that fronts ComfyUI |
+| Transport | ComfyUI's own `/prompt`, `/ws`, `/view`, `/history` | The worker bridge's `comfy.*` messages |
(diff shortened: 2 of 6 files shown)


georgi:
CI is red on `Quality Gate / test-packages` (and therefore the aggregate `quality` leg), but the failure is pre-existing on `main` and unrelated to this PR.

**The failure**

```
FAIL  packages/websocket/tests/application-budget-gate.test.ts
  > application invocation settlement
  > closes the ledger row out at what the run actually cost

AssertionError: expected { …(10) } to match object { status, actualUsd }
    "actualUsd": 0.42,
-   "status": "completed",
+   "status": "failed",
```

The cost figure settles correctly (`actualUsd: 0.42`); only the status does. `streamJobMessages` closes the ledger row out as `failed` even though the active job's status and the resolved result both say `completed`. 158 of 159 test files pass.

**Why it isn't from this PR**

- The diff is six Markdown/HTML files. `git diff --stat origin/main...HEAD` touches nothing under `packages/` except `integration-nodes/README.md`, so `packages/websocket` is byte-identical to the base branch.
- `main`'s own latest run fails on the same two jobs — [run 30439558379](https://github.com/nodetool-ai/nodetool/actions/runs/30439558379) at `c7fba4c`, `test-packages` + `quality`.
- It reproduces locally on an unmodified checkout, so it's deterministic rather than flaky.

Not fixing it here — it's a real question about the settlement path in `packages/websocket`, and folding it into a docs PR would bury it. Everything else on this PR is green: `lint`, `typecheck`, `build`, `bundle`, `deps`, `parity`, `examples`, CodeQL, both E2E suites, and the markdown link check.

I'll re-check when the base branch goes green and merge it in so CI re-runs against a fixed base.

---

PR #4586, opened 2026-07-29
fix(gemini): allow built-in tools alongside function calling

georgi:
## Problem

Requests to Gemini fail with a 400 whenever a built-in tool and function declarations are sent together:

> Please enable `tool_config.include_server_side_tool_invocations` to use Built-in tools with Function calling.

`GeminiProvider` maps `web_search` → `googleSearch` and `code_interpreter` → `codeExecution`, appending them to the same `tools` array that already carries `functionDeclarations`. Gemini requires an explicit opt-in for that combination, and we never sent it. Any chat turn with web search or the code interpreter enabled next to the `ui_*` editing tools (e.g. `ui_storyboard_add_shot`) was rejected before it reached the model.

This is a request-shape bug in our provider, not a server permission or environment setting — nothing about the deployment needed changing.

## Fix

Set `toolConfig.includeServerSideToolInvocations` when a built-in tool and at least one function declaration are both present. It is omitted otherwise, so requests that use only function tools, or only a built-in, keep their existing shape.

The `tools` / `toolConfig` assembly was duplicated verbatim between `generateMessage` and `generateMessages`, which is how the streaming path would have drifted from the fix. Both now call one `applyTools` helper.

`toolConfig.functionCallingConfig` becomes optional in the request type, since the flag can now be the only field set.

## Tests

`packages/runtime/tests/providers/gemini-provider.test.ts`:

- Updated the existing web-search + `toolChoice` case, which asserted the exact payload Gemini rejects.
- New: the flag is set for `googleSearch` and for `codeExecution` when function tools ride along.
- New: `toolConfig` stays absent for function tools alone and for a built-in alone.

Full runtime provider suite: 1675 passing.

## Also: an unrelated red CI leg

`Quality Gate / test-packages` was already failing on `main` (reproduced at 148d840, this PR's parent), so it is fixed here to get CI green rather than left blocking.

`application-budget-gate.test.ts` builds a fake `ProcessingContext` that never grew the `addMessageListener` that `createRelayActivityWaiter` calls. The waiter threw `context.addMessageListener is not a function` on every run; `streamJobMessages` catches any throw and forces `status = "failed"` before settling the ledger, so the settlement test saw a failed row where it expected a completed one.

Only the fixture was stale — a real `ProcessingContext` implements the method, so the production settlement path was never affected. The sibling fixture in `unified-websocket-runner-runjob-coverage.test.ts` already has this shape; this copies it.

`packages/websocket` suite: 159 files / 1940 tests passing, up from 1 failed / 1939.

https://claude.ai/code/session_01Eopbh9WLJE2C25jKmwqmeM

### DIFF
--- a/packages/runtime/src/providers/gemini-provider.ts
+++ b/packages/runtime/src/providers/gemini-provider.ts
@@ -83,7 +83,12 @@ interface GeminiRequest {
     | { codeExecution: Record<string, never> }
   >;
   toolConfig?: {
-    functionCallingConfig: { mode: "ANY"; allowedFunctionNames?: string[] };
+    functionCallingConfig?: { mode: "ANY"; allowedFunctionNames?: string[] };
+    /**
+     * Required by Gemini when a built-in tool (googleSearch, codeExecution) is
+     * sent alongside functionDeclarations — the API 400s without it.
+     */
+    includeServerSideToolInvocations?: boolean;
   };
   generationConfig?: Record<string, unknown>;
 }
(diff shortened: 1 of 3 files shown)


georgi:
CI is red on `Quality Gate / test-packages`, but the failure is on the base branch too — not from this PR.

Failing test:

```
packages/websocket/tests/application-budget-gate.test.ts
  > application invocation settlement
  > closes the ledger row out at what the run actually cost

- "status": "completed"
+ "status": "failed"
```

I checked out 148d840 — this PR's parent, and the current tip of `main` — and reproduced the identical failure with none of my changes applied. The run also logs `job message streaming failed context.addMessageListener is not a function`, which points at the invocation-settlement path in `unified-websocket-runner.ts` rather than anything in `packages/runtime`.

This PR touches only `packages/runtime/src/providers/gemini-provider.ts` and its test file, so it can't reach the application budget ledger. `Quality Gate / quality` is just the aggregator leg reporting the same failure.

Every other check passed: lint, typecheck, build, bundle, deps, parity, examples, test-app, Workflow Integration Tests, Workflow Runner Browser E2E, route loading, and CodeQL.

Leaving the pre-existing failure alone as out of scope. I'll re-run once `main` is green.

---

PR #4587, opened 2026-07-29
feat(pricing): sync GenSpend price catalog nightly

georgi:
Pulls realtime generative-media prices from the [GenSpend](https://genspend.io) catalog into a shipped JSON catalog, refreshed by a nightly job. **283 priced model ids across 9 providers, 225 of them from an exact provider-native id.**

GenSpend shipped a typed `providerModelId`, `capabilities`, typed variant facets, `aliases[]`, strong ETags and `?envelope=1` in response to this PR; the integration uses all six. **The join is on the native id — name matching no longer does the joining work.**

## What lands

- **`scripts/sync-genspend-pricing.mjs`** + **`scripts/genspend/`** — fetches the catalog, resolves each offering to the model ids NodeTool ships, and writes `packages/model-pricing/src/generated/genspend-pricing.json`. `npm run sync:genspend` to rewrite, `npm run sync:genspend:check` to see whether it is stale.
- **`.github/workflows/genspend-pricing.yml`** — nightly at 03:20 UTC (plus `workflow_dispatch`), opens a PR when a price moved. No provider API keys needed. A number that gates a run's budget gets reviewed, not auto-merged.
- **`getModelUnitPrice`** gains GenSpend as a third source, after FAL and kie. Those two come from the providers themselves so they stay ahead; GenSpend covers everything else.
- **Attribution** — the cost estimate panel credits its price sources and shows the date the prices last moved.
- **`packages/model-pricing/README.md`** — lookup order, match routes, guards, fetch details.

## Joining GenSpend models to NodeTool model ids

GenSpend keys models by its own slug (`seedance-2`); a run is priced by the provider-native id on the node (`bytedance/seedance-2.0/text-to-video`). Five routes bridge them, recorded per entry, most trusted first:

| Route | Entries | What it is |
| --- | --- | --- |
| `alias` | 1 | Pinned by hand in `scripts/genspend/aliases.json`; `null` blocks a model |
| `variant` | 87 | A `variants[]` row naming a model NodeTool ships — carries that endpoint's **own** price |
| `provider-id` | 122 | `providerModelId`, the id the provider's API invokes and bills with |
| `receipt` | 15 | The `sourceUrl` is a model page carrying the native id |
| `catalog` | 58 | Name match — spreads a resolved price across a model's sibling task endpoints (see below) |

Every claimed id is verified against the provider's own NodeTool listing (`getAvailableImageModels` / `VideoModels` / `TTSModels` — the model picker's source) before it becomes a key. That is what keeps an id for a model NodeTool doesn't ship, like `resemble-ai/chatterbox`, from becoming a key nothing can look up, and it covers the providers that publish no ids at all.

### Why a name-matching tier still exists

Not because ids are missing — because NodeTool's ids are **more granular than one per offering**. GenSpend gives one callable id per provider offering; NodeTool ships an aggregator's model as several per-task node ids. AtlasCloud's HappyHorse 1.1 is three nodes at one price:

```
alibaba/happyhorse-1.1/text-to-video       ← providerModelId names this one
alibaba/happyhorse-1.1/image-to-video      ← catalog
alibaba/happyhorse-1.1/reference-to-video  ← catalog
```

Measured across the shipped catalog, of the 58 `catalog` entries:

- **54 are sibling endpoints of a model whose native id resolved exactly.** Without them those nodes would read as unknown-cost in an estimate that already knows the model's price.
- **4 are the only cases where no native id matched**, and there the id is *more specific* than anything NodeTool ships — `bytedance/seedream-v5.0-lite/edit-sequential` and `google/nano-banana-2-lite/edit-developer` are real AtlasCloud endpoints our AtlasCloud manifest doesn't carry. Verification declines to key on an id no node can select, and the name match prices the two nodes we do ship.

So the remaining inference is "this price also applies to the same model's other task endpoints", capability-checked and labelled `catalog` — not "this name looks like that model".

Also worth calling out: **variant rows are a correctness fix, not just coverage.** fal's flux-pro family bills $0.04 / $0.05 / $0.06 across endpoints that previously all shared the model's headline number.

### Coverage

From 447 offerings over 135 tracked models:

| Provider | Offerings priced | Model ids |
| --- | --- | --- |
| fal_ai | 73/73 | 138 |
| atlascloud | 22/47 | 50 |
| kie | 24/29 | 36 |
| replicate | 21/21 | 21 |
| together | 12/15 | 12 |
| gemini | 9/13 | 9 |
| openai | 8/8 | 8 |
| minimax | 5/6 | 5 |
| elevenlabs | 4/5 | 4 |

Unresolved offerings are printed by the sync and written to its `--report` file (a CI artifact), so a gap is visible rather than silently absent. Most are models the provider doesn't ship in NodeTool at all — AtlasCloud has no FLUX, kie has no Suno. xAI is walked every run but priced by nothing, because its model listing needs a live API call; the catalog's `providers` field lists only the nine that actually carry a price, so it can't overstate what the catalog answers.

## Guards against a wrong number

A wrong price here gates a run's budget, so the matcher refuses more than it guesses:

- **Published capabilities are respected.** The receipted task flags (`t2i`/`i2i`/`t2v`/`i2v`/`r2v`/`v2v`) drop an endpoint whose task the model cannot do — kie's `qwen-image` reports `i2i: false`, so its price no longer lands on an image-edit endpoint, a match flagged as questionable when the name matcher first produced it. Only an explicit `false` blocks; `null` is unknown, not refuted.
- **Same modality, and speech is not music.** "Gemini 3.1 Flash Image" and "Gemini 3.1 Flash TTS" normalize alike once the task word is stripped; the modality guard keeps them apart, and `capabilities.kind === "music"` keeps a music model off a provider's TTS listing.
- **Exact key equality**, never prefix or fuzzy: `seedance-2` cannot price `seedance-2-mini`, `kling-3-pro` cannot price `kling-3-turbo`. Published `aliases[]` are matched too, which is how Nano Banana Pro and Gemini 3 Pro Image find each other.
- **Generation endpoints only.** Upscalers, lip-sync and voice-clone bill on their own basis and never inherit a model's generation price.
- **Ambiguity is dropped, not resolved.** A name hitting more than eight ids is a family name; it is reported instead of priced.
- **A video-input variant never prices a text-to-video endpoint.** `videoInput: true` is a billing axis; taking that row for a t2v endpoint would undercharge, so the headline price stands.
- **Tier ambiguity resolves upward.** Where a provider selects tier by parameter, Pro and Standard share one callable id and one NodeTool node could run either — two models on one id keep the dearer price, while the same model listed twice keeps the cheaper, which is what a run pays.
- **No unit conversion.** `priceUsd` is in the provider's own unit and only comparable inside one `unitClass`; each entry keeps `unit_class` verbatim and maps it to a `billing_unit` label matching the FAL catalog's vocabulary.

Neither of the last two hazards occurs in today's data; both are guarded because the data is allowed to change.

## Fetching

`?envelope=1` plus `If-None-Match`; a 304 ends the run as a no-op. Two transport details, both handled:

- **The strong ETag arrives weak.** A compressing intermediary (Brotli, on Node's default `Accept-Encoding`) rewrites `"abc"` → `W/"abc"`, and the origin then compares strongly and misses. The strong form is stored and sent.
- **The ETag covers the envelope's `generatedAt`**, which turns over with the 60 s edge cache, so a nightly run usually gets a fresh body anyway. `etag` and `catalogGeneratedAt` are frozen alongside the prices — otherwise the shipped file would change nightly and open a PR with no price movement. Verified: three consecutive syncs leave the file byte-identical.

Hashing the ETag over catalog state rather than the response body would make conditional requests pay off at nightly cadence — the one remaining upstream nice-to-have.

## Review notes

- **Purely additive to pricing behaviour.** FAL and kie keep first and second place in `getModelUnitPrice`, so every price that resolved before this PR resolves to the same number. GenSpend only fills what was previously `null` — which is what the editor shows as an unknown-cost node and what the budget gate excludes from its total.
- **The generated catalog is not hand-editable.** `npm run sync:genspend:check` fails if it drifts from a live fetch, so a hand edit is caught rather than shipped.
- **What to actually review** is the diff of `packages/model-pricing/src/generated/genspend-pricing.json`: a price moving by an order of magnitude, a `unit_class` that no longer matches its `billing_unit`, a model losing its price, or a new `catalog`-route entry naming a model it shouldn't.

## Verification

- `npx vitest run --root packages/model-pricing` — **50 pass** (normalization, all five routes, every guard, the capability tri-state, the music/speech split, aliases, per-variant pricing, both tie-break rules, the metadata freeze, the priced-providers list, key sorting)
- `npx jest src/components/costs src/hooks/__tests__/useWorkflowCostEstimate.test.tsx` — 29 pass, including a panel test for the attribution line
- `npm run typecheck` — web and electron clean (mobile fails only on uninstalled Expo deps in the dev container)
- `npm run lint` — clean
- `npm run build:packages`, `npx vite build`, `npm run backend:smoke` — pass, so the JSON module resolves in the browser bundle and the packaged backend alike
- `npm run sync:genspend:check` — catalog matches a live fetch
- **CI green on `7ffd136`**, all 24 checks. The earlier `test-packages` failure was `video-nodes`' headless-Chrome test (`model3d-render.test.ts`, `ECONNREFUSED` on the CDP port) failing to launch under load; it passed on a fresh run with no change to that package, and this branch touches no `video-nodes` file.


georgi:
CI is red on `398571d`, and the failure is not from this diff — it needs a re-run, which my token isn't permitted to trigger (`rerun-failed-jobs` → 403).

**What failed:** `Quality Gate / test-packages`, one test:

```
FAIL packages/video-nodes/tests/model3d-render.test.ts
  Error: connect ECONNREFUSED 127.0.0.1:44113          (after 25.9s)
```

`quality` then failed only because it aggregates that leg. Everything else on the commit is green — build, lint, deps, typecheck, bundle, parity, examples, test-app, CodeQL, both browser E2E suites, route load, component visuals.

**Why it isn't mine:**

- This branch touches no `video-nodes` file. The diff is `packages/model-pricing/**`, `scripts/genspend/**`, `scripts/sync-genspend-pricing.mjs`, `web/src/components/costs/**`, `web/jest.config.ts`, `CLAUDE.md`, `package.json`, and the workflow — nothing `video-nodes` builds or imports.
- The same job was green on the previous commit of this same branch (`c205b8e`), with identical `video-nodes` code.
- `npm run test --workspace=packages/video-nodes` passes locally, 171/171.

**Shape of the flake, for whoever owns that test:** the headless case is designed to skip when no Chrome binary is found, so CI without a browser still passes. Here Chrome *was* found, so the test ran, but the CDP port never became connectable inside the retry budget — the sibling `model3d-cdp-connect.test.ts` covers exactly that retry loop and its timeout. So the gap is a launch that is detected as available but doesn't come up in ~26 s under load, rather than a missing-browser case. Treating an exhausted CDP budget the same way as a missing binary, or giving the launch a longer budget, would close it; I've left that alone since it's outside this PR and could mask a real regression in that node.

A re-run of the failed jobs should clear it. If it reproduces on the re-run, I'll take another look at whether anything here could plausibly reach that test.

---

georgi:
Reviewed the diff — code looks good: purely additive (no deleted exports or files), no junk files, GenSpend is wired as a third fallback after FAL and kie so existing pricing behavior is unchanged.

**CI status:** The `test-packages` failure is a flaky test in `@nodetool-ai/video-nodes` (`model3d-render.test.ts` — `ECONNREFUSED 127.0.0.1:44113` on headless Chrome CDP). This PR doesn't touch `packages/video-nodes/` at all, and all 49 `model-pricing` tests (including the new 44-test `genspend-sync.test.ts`) passed. The same `test-packages` leg passes on the 5 other open PRs.

The `Playwright visual snapshots` check is also still in progress.

Recommend re-running CI once snapshots complete. Mergeable once CI is fully green.

---

PR #4595, opened 2026-07-29
Polyfill crypto.randomUUID for insecure (plain-http) contexts

Neosettler:
crypto.randomUUID is a secure-context-only API, so it is undefined when the web app is served over http from a non-localhost origin (e.g. a server IP). The ~30 direct call sites then throw, which breaks workflow/tab creation: the UI works from http://localhost but not http://<server-ip>.

Add an early polyfill that defines crypto.randomUUID via crypto.getRandomValues (not gated to secure contexts) only when the native method is missing, so secure contexts keep the browser implementation. Fixes all call sites at once with no HTTPS requirement.


georgi:
@claude review

georgi:
I found no blocking correctness issue. The polyfill generates a valid RFC 4122 UUID v4, preserves the native implementation where available, and is imported early enough for normal web-app modules. The PR is small and appropriately scoped.

Findings

P2: Add a focused regression test

The PR fixes an environment-specific startup failure but adds no test proving that:

1. the fallback is installed when crypto.randomUUID is absent,
2. generated values have version 4 and RFC variant bits,
3. an existing native method is not overwritten.

Without that test, deleting or moving the early import could silently reintroduce the problem. The repository already has crypto stubs and a separate mobile UUID polyfill, suggesting this is an established compatibility surface rather than a one-off workaround.

A Vitest test could temporarily replace globalThis.crypto, dynamically import the module, and assert something like:

expect(uuid).toMatch(
  /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/
);

This is not severe enough to block the fix, but I would request it if NodeTool generally requires regression coverage.

Minor design comments

The claim that this “fixes all call sites at once” is true for code running in the main web document, not necessarily separate Web Workers or other entry points. Several crypto.randomUUID uses exist outside web/src, including browser runtime packages and worker-related code.

That does not appear to invalidate this PR, since the reported UI creation failure happens in the main renderer, and at least some browser runtime code already has its own fallback. For example, workflow-runner’s browser job ID generation falls back to a timestamp and random string when randomUUID is missing.  I would nevertheless narrow the PR wording from “all call sites” to “web renderer call sites.”

The current workflow statuses are action_required, not conventional test failures, likely because workflows from this contributor require approval. Therefore CI has not actually validated the latest head yet.

Verdict

Approve after CI is authorized, optionally requesting a regression test. I would not block the practical fix over architectural purity. Centralizing the compatibility behavior at the renderer entry point is preferable to changing roughly 30 UUID call sites individually.

PR #4599, opened 2026-07-29
docs: reliability architecture — one Run Host, one protocol codec, cross-surface conformance

georgi:
Adds `docs/RELIABILITY_ARCHITECTURE.md`, a design proposal for the execution layer. No code changes.

## The finding

The recent failure classes (CLI vs WebSocket divergence, packaged Electron breakage, reconnect / backpressure / half-open sockets, fakes accepting shapes the real Python worker rejects, lifecycle races, provider drift) share one cause: the repo has **more than one implementation of "run a workflow"**, each with its own graph hydration, context construction, cancellation, error mapping, and job-status persistence.

Measured on `main`:

- `new WorkflowRunner(` — 13 occurrences across 11 non-test files, 3 of them inside `packages/websocket/src/unified-websocket-runner.ts` (7,709 lines)
- `new ProcessingContext` — 16 non-test sites
- `hydrateGraphNodeFlags` — 7 non-test call sites

Unit tests are scoped inside one copy each, so no test in the repo can fail when two copies disagree. That is why coverage stayed high through every one of these incidents.

## The proposal

1. **`packages/execution` / `RunHost`** — the only code allowed to hydrate a graph, build a `ProcessingContext`, construct a `WorkflowRunner`, supervise a run, or persist job status. Surfaces (WebSocket, HTTP, CLI, Electron, mobile, app-runtime, MCP, trigger dispatch) become transport adapters that may not assert or branch on node type. Enforced by an import lint rule, not convention.
2. **One canonical `RunEvent` stream**, cursor-addressable so `attach(runId, cursor)` makes reconnect a Run Host concern instead of a WebSocket concern.
3. **One protocol codec per boundary**, with a conformance suite the *real* Python worker must pass and against which fakes are validated at construction.
4. **A Reliability Harness** that runs the same journey on every surface, normalizes each into a `RunTranscript`, and diffs them.

`RunEnv` (clock, rng, idgen, storage, secrets, provider transport, python bridge, faults, capabilities) is the single seam for both determinism and fault injection, so no chaos flag ever enters a production path.

## What the doc covers

- §1 why unit tests structurally miss these bugs (test-double-as-spec, `src/`-only module resolution, hidden interleavings, local assertions vs global invariants)
- §3–5 golden journeys: yes, with three conditions (run on artifacts not sources, contracts not scripts, engineered determinism); 18 journeys chosen by the seam each covers, not by feature popularity
- §6 + §10 the invariant set — per-run (transcript), resource (probes), system (never-violated), always-on rather than opt-in
- §7–8 one execution implementation, plus six mechanisms that stop drift from regenerating, and a 6-phase migration where each step is gated by the harness proving the transcript did not change
- §9 fault-injection matrix keyed to `RunEnv` seams, including a seeded deterministic scheduler hook to make lifecycle races reproducible
- §11 four-tier release gate; the release-blocking tier validates packaged artifacts and includes a reviewed transcript diff against the previous release
- §12 harness architecture and the `SurfaceDriver` contract (drivers translate, only the Judge asserts)
- §13 what not to do; §14 the three items to build if only three get built

Generalizes the instinct already present in `docs/KERNEL_PARITY_GAPS.md` and `npm run backend:smoke` from TS-vs-Python parity to surface-vs-surface parity, mechanized so it does not depend on anyone remembering.

## Notes

- `docs/_config.yml` excludes the file from the published site — engineering proposal, not user docs.
- No front matter, no site nav entry, so `docs-ci` link and leak checks are unaffected.
- Prose follows `docs/WRITING_STYLE.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MApZpk8V2wrXcSNKefhsTD)_

### DIFF
--- a/docs/RELIABILITY_ARCHITECTURE.md
+++ b/docs/RELIABILITY_ARCHITECTURE.md
@@ -1,0 +1,897 @@
+# Reliability Architecture
+
+One execution path, one protocol codec, and a cross-surface conformance harness
+that validates releases with golden workflow journeys instead of isolated
+package tests.
+
+**Navigation**: [Architecture](architecture.md) | [Execution Strategies](execution-strategies.md) | [Kernel Parity Gaps](KERNEL_PARITY_GAPS.md) | [Development Standards](DEVELOPMENT_STANDARDS.md)
+
+Status: proposal. Written as a plan of record for the next several years of
+execution-layer work, not as a description of what exists today.
+
+## 0. Thesis
+
+The recent failures were not caused by missing tests. They were caused by
+NodeTool having **more than one implementation of "run a workflow"**, each with
+its own graph hydration, context construction, lifecycle, cancellation, error
+mapping, and job-status persistence. Every new surface added another one.
+Unit tests cover each copy in isolation and therefore cannot see the thing that
+is actually broken: the divergence between the copies.
+
+There are at least six live copies today:
+
+| Site | Lines | What it re-implements |
+|---|---|---|
+| `packages/websocket/src/unified-websocket-runner.ts` | 7,709 | context, hydration, cancel, job status, asset autosave, error sanitization, chat/agent glue |
+| `packages/websocket/src/headless-job-runner.ts` | 278 | context, hydration, job row, terminal status |
+| `packages/workflow-runner/src/run.ts` | 225 | context, hydration, abort wiring |
+| `packages/cli/src/nodetool.ts` (`workflows run`) | ~200 of 2k | context, hydration, secret resolution, output printing |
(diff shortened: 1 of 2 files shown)


georgi:
This PR has a merge conflict: `docs/RELIABILITY_ARCHITECTURE.md` was already merged to main via PR #4601 (`docs: reliability architecture proposal — execution facade, golden journeys, fault injection`). The two versions differ in size (this PR's is 897 lines vs main's 509 lines), so there may be additional content here worth preserving.

Options:
1. Close this PR if #4601 supersedes it
2. Rebase onto main and merge in the additional content from this version

---

PR #4614, opened 2026-07-30
feat: Code Node AI authoring — describe the result, get a typed working node

georgi:
Implements `docs/plans/code-node-ai-authoring.md`: describe the result you want and get a working, typed Code Node, instead of writing JavaScript or hunting the catalog for a chain of type-compatible nodes.

Nothing is removed. A generated node is indistinguishable from a hand-written one — generation fills the fields the node already has (`code`, `dynamic_inputs`, `dynamic_outputs`, `dynamic_properties`) and gets out of the way. No new persistence, no node state, no contract layer.

**All 9 tasks landed.** Ships behind a flag (`NODETOOL_CODE_GENERATION=1`), off by default.

- [x] 1 — Sandbox capability manifest
- [x] 2 — Transport schemas
- [x] 3 — `CodePlanner` + `SubmitCodeTool`
- [x] 4 — Serving path
- [x] 5 — `code_model` preference
- [x] 6 — Dialog and apply
- [x] 7 — Graph entry points
- [x] 8 — Preview
- [x] 9 — Eval suite and rollout flag

## Live eval results

The `code-gen` suite was run against a real model (`claude_agent_sdk`), not just the scripted harness:

| Model | First-pass | Post-repair | Mean score | Submit rounds |
|---|---|---|---|---|
| sonnet | 8/8 (100%) | 8/8 (100%) | 1.00 | 1.0 |
| haiku | 8/8 (100%) | 8/8 (100%) | 1.00 | 1.0 |

`--min-success 0.9` exits 0, so the CI gate works against a live provider.

Two caveats, stated plainly. **This is one provider, not two** — two models on the same transport is weaker evidence than the plan's "two configured providers", because it doesn't exercise a second tool-calling implementation. And **cost reads `-`**: `CodePlanner` aborts the provider loop from inside the accepting tool, and the SDK only reports usage on a terminal `result` message a cancelled query never emits (documented in `packages/agents/CLAUDE.md`). The runs weren't free; the usage is unobservable.

## Three bugs found along the way

**The Code Node advertised libraries that don't exist.** The authoring instructions promised `_` (lodash), `dayjs`, `cheerio`, `csvParse` and `validator` as guest globals. None of them do — the QuickJS guest has no module loader, so a host bridge is the only way a library reaches user code. Code written against that advice failed at runtime.

The same phantom names were in `SANDBOX_GLOBALS`, which drives input-handle inference, so it cut both ways: a real undefined variable named `_` never became an input handle, while the actual bridges (`format`, `data`, `progress`) were missing from the set and could sprout bogus ones.

⚠️ **Behavior change worth noting**: a Code Node whose code references a variable named `data` will no longer get an input handle for it, because `data` is now a real sandbox bridge. Correct, but visible on existing nodes.

**The generation instruction was leaking into traces.** `CodePlanner` passed it as the span's `objective`, which lands in `agent.objective` and ships to Traceloop/OTLP when configured — contrary to the plan's rule that telemetry keeps usage metadata, not instructions, samples or code. Spans now carry the request's shape, and the omission is pinned by a test rather than left to review.

**Generated ports could be typed with names nothing connects to.** Found by running the eval against a live model: asked for a count, it declared the output type as `integer`. NodeTool's numeric types are `int`/`float`/`number`, and `isConnectable` compares these names — so the handle would have been created with a type no `int` input accepts. The node looks correct and the wire silently won't attach.

The transport schema can't catch this: `type` is deliberately an open string, since asset and node-defined type names are unbounded. The feedback round can. The tool now rejects the specific failure models actually produce — JSON Schema and TypeScript spellings of types NodeTool has under another name — and names the replacement, so the model corrects and resubmits. Names that are merely unrecognized still pass. The prompt states the vocabulary too, so the round is rarely needed; after the change neither model emitted an alias.

## Design notes

**The manifest removes the conditions for drift** rather than just fixing the prose. Names derive from `EXPOSED_BRIDGE_NAMES`/`GUEST_HELPER_NAMES`, numbers from the exported limit constants, and clamp ceilings from calling `resolveSandboxLimits` instead of restating the table. Bridge docs are a mapped type over the name tuple, so marshaling a new global without documenting it is a compile error. Web can't import the manifest (agents isn't a web dependency and pulls the QuickJS WASM runtime), so its list stays local and is pinned from the agents side — the drift test reads the web files off disk and fails in either direction.

**Generation is the `GraphPlanner` loop with a smaller payload** — a forced submit tool, validation inside the tool, failures returned as tool results so the model resubmits, capped at 3 rounds. Only the accepted `submit_code` arguments are read; prose and code fences are never scraped. The tool parses with acorn and checks every declared output is assigned on every statically visible return path — a branch that omits one is workflow branching, not a codegen bug, so the message points at `nodetool.control.If`/`Switch`.

**Transport: tRPC mutation, not the agent socket.** The plan preferred the agent WebSocket path, but it's session-shaped end to end: every command is addressed to a `create_session` session, `sendMessageStreaming` throws unless the renderer advertises the full `ui_*` tool manifest, and the `llm` session writes Thread/Message history per turn — exactly what this call must not do. `GraphPlanner` lives there as a tool inside a chat turn, not a servable endpoint. Using it would have meant inventing a session-less command family on a session protocol, so this took the plan's documented fallback.

**Apply is one `updateNodeData`** carrying code, title, typed slots and defaults together. `customEquality` already shallow-compares those fields, so it registers as one undo entry — asserted against the real temporal store, since splitting the write would strand a half-generated node behind an undo.

**Cancel restores the graph exactly.** `onClose` fires for both Apply and Cancel, so the host can't tell them apart from the callback — it reads the node instead, checking for the non-empty `code` only `applyCodeGenSubmission` writes. A test snapshots node ids and edge shapes either side of a cancel.

**Samples are opt-in, and oversize payloads block rather than truncate.** By default the model sees names and types only. Ticking the box reveals the actual serialized payload — not a summary — and Generate stays disabled until it's acknowledged. Over `MAX_SAMPLE_VALUES_BYTES` it names both numbers and refuses: silently trimming would send a different payload than the one shown, quietly breaking the consent the checkbox exists to provide.

**The flag is server-driven runtime config, not a user setting.** Generation needs the mutation, provider secrets, and a tool-capable model — all server-side — so whether it can work is a property of the deployment. A user toggle would let someone switch on a feature their server can't serve. All four surfaces are *absent*, not disabled, when off.

## Testing

`protocol` 772 · `agents` 1780 · `websocket` 1987 · `cli` 444 · `config` 146 · `web` 1061 suites / 12,481 tests — all green, plus the full Quality Gate on CI. Typecheck clean; `eslint src` in web unchanged at its 44 pre-existing errors. The packaged-backend `bundle` leg passes, which matters because adding `acorn` as a new runtime dependency of `agents` is exactly what breaks flat `_modules/` staging.

## For reviewers

- `acorn` is declared in `packages/agents` for the first time, resolving against the root-hoisted 8.16.0 (`web/` already imported it undeclared). CI's `deps` and `bundle` legs pass with it, but it's worth a glance since nothing declared it before.
- **Still not done**: the two Playwright E2E journeys the plan lists (source-output and destination-handle), which need a live server driving a real browser. The evals above cover generation quality; they do not cover the UI path end to end. The flag is off by default partly for that reason.
- One telemetry residual left deliberately: `BaseProvider` puts up to 2000 chars of the model's **prose** response on non-streaming `llm.chat` spans. Generated code never reaches a span (it arrives as tool-call args; only `tool_calls_count` is traced), and the serving path discards `llm_call` payloads after reading token counters. Suppressing the prose too would be a cross-cutting `BaseProvider` change that degrades the analyzer-agent tracing documented in `packages/agents/CLAUDE.md`. Happy to add a per-call opt-out if wanted.
- Minor repo observations, both unrelated to this diff: `packages/agents`' `lint` is `tsc --noEmit` while `build:packages` runs `tsc --build`, which checks more — I hit a case where lint passed and the build failed on the same file. And `@nodetool-ai/video-nodes`' `model3d-render` test intermittently fails CI with `connect ECONNREFUSED` on its headless-Chrome CDP port ([analysis](https://github.com/nodetool-ai/nodetool/pull/4614#issuecomment-5137263766)); it blocked this PR twice before passing. Happy to harden either in a follow-up.


georgi:
## `test-packages` is red on an unrelated headless-Chrome test

Flagging this rather than leaving the red check unexplained: the failure is not in this diff, and I don't think it should be fixed here.

**What fails** — the same single test on two consecutive SHAs (`6706a9f`, `c4261bf`):

```
FAIL  tests/model3d-render.test.ts > RenderToImage headless render
      > renders a GLB to a PNG of the requested size   (26128 ms)
Error: connect ECONNREFUSED 127.0.0.1:39271
```

`@nodetool-ai/video-nodes`, 1 failed / 171 passed. It launches headless Chrome and drives it over CDP; the run spent 26s retrying a debug port that never accepted a connection. The job's cleanup then terminates orphan `chrome` / `chrome_crashpad_handler` processes.

**Why it isn't this branch:**

- The diff contains no 3D, Chrome, CDP, video, or rendering code.
- `video-nodes` depends on none of the packages this PR changes except `protocol`, where the change is purely additive (one new schema file plus one barrel line).
- The suite passes locally — 172 tests, and the Chrome/CDP-driven files pass 3/3 on repeat.
- The failure mode is a refused TCP connect to a locally spawned browser, not an assertion.

I was wrong about this twice before landing on the cause, so to be explicit about the residual uncertainty: `main` is green, including this PR's base commit, so I can't prove it's pre-existing — only that nothing in the diff plausibly reaches it. Most likely a loaded runner failing to bring Chrome up in time; both failures fell within about 25 minutes of each other.

**Not fixing it here.** The fix would be to make someone else's Chrome-dependent test resilient (retry the launch, or skip when no browser is reachable) — unrelated to Code Node authoring and better as its own change. Happy to do it in a follow-up if you'd prefer that over re-running the job.

**Unrelated diagnostic note:** `test-packages` runs everything through Turbo concurrently, and when one package's process dies you get an exit code with no failing test name — the vitest output is interleaved away. That turned a one-line failure into a multi-round investigation.

---

georgi:
Resolved — `test-packages` is **green** on `c41ee95`, and the full Quality Gate now passes (all 25 checks). The `model3d-render` headless-Chrome failure was intermittent, as the analysis above concluded: same code, same runner image, passing this time.

Leaving the earlier comment up since the diagnosis still stands if it resurfaces — and the offer holds to make that test retry its Chrome launch (or skip when no browser is reachable) in a follow-up, so it stops blocking unrelated PRs.

---

PR #4616, opened 2026-07-31
marketing: card art from real workflow output, every template indexable

georgi:
Card art gated the marketing pages. `generate-template-entries.mjs` emits `indexable: false` for any example with no matching `<name>.jpg`, keeping its `/templates` page out of the sitemap and the smoke walk — so **122 examples had a page nobody could find**, and only 53 of 219 had art at all.

**219/219 indexable now, up from 36.**

## The art comes from what the workflows produced

`scripts/generate-example-thumbnails.mjs` takes two sources, preferring the honest one:

```bash
npm run gen:example-thumbnails                                        # rendered cards
node scripts/generate-example-thumbnails.mjs --outputs /tmp/sweep     # real output
```

**`--outputs <dir>`** — a directory of `<Example Name>.(png|jpg|webp|mp4|mov)` collected from a `workflows run` sweep. The image, or the first frame of the clip, becomes the card **full bleed** under a gradient scrim carrying the title. 27 cards are real generated output from this branch's verification sweeps.

**Otherwise** — a card rendered from the workflow itself: title, description, and the node pipeline as chips. Read from the graph, so it stays true as the graph changes rather than drifting into decoration.

Built on **sharp**, already a dependency, compositing SVG. No new image toolchain, and it runs in CI.

## Why generated rather than executed, for the remaining 122

They are older shipped examples needing paid APIs and input media that aren't available here, and this branch's sweeps already spent real money. The script makes execution strictly better whenever someone *does* have those inputs: run a sweep, point `--outputs` at it, and those cards upgrade from rendered to real with no other change.

## Verification

- `templateEntries.generated.ts` regenerates byte-identical on a fresh checkout — the CI staleness gate is satisfied
- 219 entries, 219 indexable
- No shipped card art modified: the 53 pre-existing cards are untouched

## Two things worth knowing

**`--force` overwrites everything, including hand-made art.** It did exactly that mid-work; I caught it in `git status` and restored the 36 affected shipped cards with `git checkout`. The flag behaves as documented, but on a repo with hand-authored art it is a live footgun. A provenance marker in the file would be the real guard — not added here.

**JPG cannot carry alpha.** `Cut a Product Out of Its Background` fills its frame with what looks like the input, because a transparent cutout flattens to white. It is the genuine output. If cutout examples matter for the gallery, PNG cards would fix it.


georgi (commented):
Reviewed the CI status. The Quality Gate checks (typecheck, lint, tests, bundle, parity, reliability, examples) all pass. Two marketing-specific checks failed:

1. **"Typecheck, lint & build"** — `providerCatalog.generated.ts` is stale. The error: `[provider-catalog] marketing/src/data/providerCatalog.generated.ts is stale. Run: npm run seo:provider-catalog`. This is likely because the PR is based on an older main commit (`006a8b6`) and main has since advanced. Merging main into this branch and running `npm run seo:provider-catalog` in `marketing/` should fix it.

2. **"Playwright smoke tests"** — `Timed out waiting 180000ms from config.webServer`. The marketing dev server didn't start in time. This looks like a CI infrastructure timeout, not a code issue.

This is a 369-file PR (mostly image assets). The diff is too large for GitHub's diff API (>300 files). I confirmed no cross-PR conflicts with #4617 and #4618 (which I just merged) — those touched script editor and tool registry code, not marketing or assets.

**Action needed**: merge main into this branch and regenerate the provider catalog to get CI green.

---

PR #4633, opened 2026-08-01
Workflow Supervisor phase A: the escalation mechanism

georgi:
Implements **PR 1 and PR 2** of [the supervisor implementation plan](docs/workflow-supervisor-implementation-plan.md) — the kernel mechanism, no LLM anywhere in the diff.

A failing node invocation raises an **escalation**; a handler returns a **verdict** (retry · substitute · skip · end_stream · fail). Without a supervisor no escalation is ever constructed, and a supervised *clean* run emits the same message stream as an unsupervised one — asserted, not assumed.

## Why the two PRs landed together

PR 1's allowed-set enforcement is only as strong as PR 2's invocation-scoped cost tracking. Shipping PR 1 alone would leave a window where `retry` is offered after a billed call — the one thing §5.3 exists to prevent. The plan doc records the merge.

## What's here

**protocol** — `Escalation`, `Verdict`, `Intervention`, and the `supervisor_escalation` / `supervisor_decision` message variants, Zod-first like the rest of the union; `retry_safe` on `NodeDescriptor`.

**runtime**
- Resolved-secret registry on `ProcessingContext`. Redaction needs a subject, and the resolver previously answered one key and forgot it.
- Invocation-scoped cost/asset accounting via `AsyncLocalStorage`. A stack on the shared context cross-attributes whenever concurrent actors complete out of order — there's a test for exactly that reversal.
- `RecoverableNodeError`, so a node hands over the value that needs repairing instead of losing it.

**kernel**
- `SupervisorHandle`, `FailClosedHandle`, `BoundedHandle` — decision/retry ceilings, per-decision timeout derived from the run signal, serialized queue, sticky verdicts keyed on `(nodeId, failureSignature)`. Every boundary resolves as `fail`.
- Allowed-action computation, enforced by the actor **independently of any schema**: a handle returning `retry` for an unsafe node gets `fail`.
- Redaction at escalation construction — `Escalation` is public (message, `RunResult`, websocket), so no unredacted instance exists anywhere in the system.
- Runtime substitute validation against declared output types, with an async hook for reference resolution.
- The recovery loop across the buffered, controlled, streaming-output, and `run()` paths. It wraps **invocation only**: a routing failure fails the node, because a retry after partial delivery would duplicate it.
- `skip` emits `lineage_done` **and** `lineage_scope_closed` on every outgoing edge, so a downstream aggregate collapses instead of waiting on a scope the skipped invocation never opened.

**node-sdk** — `retrySafe` metadata opt-in, declared across the pure-compute and read-only node categories. Unknown stays unsafe: cost tracking cannot see external writes, so a node nobody classified loses a safe retry rather than gaining an unsafe one.

## Tests

54 new kernel tests, 6 runtime tests. Each design rule that could fail silently has one: reverse-order cost attribution, a hostile handle rejected without schema help, a planted secret absent from every emitted message and from `RunResult`, a skipped fan-out item letting its aggregate collapse, a skipped iterator's scopes getting closed, `end_stream` keeping what it produced, a `run()` node's verdict set being exactly `{end_stream, fail}`, seven identical failures costing one decision.

Full kernel suite 988 passing, runtime 2638 passing, node-sdk 691 passing; `build:packages`, lint and prettier clean.

## Deliberate gaps

- **Reference-typed outputs can't be substituted yet.** Resolving an `ImageRef` uri against run storage needs a resolver the kernel doesn't hold. Rather than accept an unresolvable ref, `substitute` is withheld when any declared output is reference-typed; the validator already takes the `resolveRef` hook PR 3 will wire.
- **No surface is wired.** `ExecutionSessionOptions.supervisor`, the CLI flags, and the UI belong to PR 4+ and need the agent from PR 3. Today only kernel tests can construct a handle.
- **Provider retry audit** (design §10.3) is findings-only: [docs/workflow-supervisor-provider-retry-audit.md](docs/workflow-supervisor-provider-retry-audit.md). Gemini and Ollama have no backoff at all and would flood the supervisor with escalations a two-second wait would fix. Those fixes belong in runtime, not here — but they should land before PR 3 makes escalations expensive.

## Doc changes

Design §5.1 now says the handle is wired through the runner into each actor rather than onto `ProcessingContext` — node code has no business raising or answering escalations, and the actor already holds what the escalation constructor needs. `Escalation` gains `emitted`, and `SupervisorHandle.decide` returns a `DecisionOutcome` (verdict + `decidedBy` + cost) rather than a bare verdict, so the intervention record has one source.

---
_Generated by [Claude Code](https://claude.ai/code/session_011yMUYSSrceG3dTcX3XmsGJ)_

### DIFF
--- a/docs/workflow-supervisor-design.md
+++ b/docs/workflow-supervisor-design.md
@@ -68,6 +68,7 @@ export interface Escalation {
   spentCostUsd: number;            // provider cost recorded by this invocation
   createdAssets: boolean;
   retrySafe: boolean;              // node metadata opt-in; unknown ⇒ false ⇒ no retry
+  emitted: boolean;                // streaming: the invocation already sent something downstream (§5.4)
 }
 
 export type Verdict =
@@ -102,7 +103,7 @@ Status stays the existing enum. "Completed — supervised" is `status: "complete
 
 ### 5.1 Where it sits
 
-`ProcessingContext` gains one optional member, `escalate?: (e: Escalation) => Promise<Verdict>`, wired by `WorkflowRunner` from a new option:
+`WorkflowRunner` takes the handle as an option and hands it to each `NodeActor`, alongside the callback that records interventions. The handle stays in the kernel rather than on `ProcessingContext`: node code has no business raising or answering escalations, and the actor already holds everything the escalation constructor needs. What the context *does* gain is the resolved-secret registry the constructor masks against (§6.1) and per-invocation cost/asset accounting (§5.3).
 
 ```ts
 export interface WorkflowRunnerOptions {
@@ -111,8 +112,15 @@ export interface WorkflowRunnerOptions {
 }
 
 export interface SupervisorHandle {
-  decide(e: Escalation, signal: AbortSignal): Promise<Verdict>;
+  decide(e: Escalation, signal: AbortSignal): Promise<DecisionOutcome>;
   close(): void;
+}
+
+/** The verdict, who made it, and what it cost — the intervention record minus its escalation. */
+export interface DecisionOutcome {
+  verdict: Verdict;
+  decidedBy: "agent" | "sticky" | "bounds" | "default";
+  costUsd?: number;
 }
 ```
 
--- a/docs/workflow-supervisor-implementation-plan.md
+++ b/docs/workflow-supervisor-implementation-plan.md
@@ -19,7 +19,11 @@ PRs marked ∥ can proceed in parallel once their listed dependency lands.
 
 ---
 
-## Phase A — the mechanism (kernel, no LLM)
+## Phase A — the mechanism (kernel, no LLM) — **shipped**
+
+PR 1 and PR 2 landed together: PR 1's allowed-set enforcement is only as good as
+PR 2's invocation-scoped cost tracking, so shipping the first without the second
+would have meant a window where `retry` was offered after a billed call.
 
 ### PR 1 — Escalation types and the actor hook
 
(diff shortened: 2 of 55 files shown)


georgi:
Pushed `07888ea` addressing the review. One of the comments turned out to be a real bug rather than a nit, and two CI legs are red on the base branch.

## The review comment that mattered

Both hydration paths carried `?? node.retry_safe` after the registry lookup, copying the pattern of the neighbouring streaming/control flags. That is wrong for this flag specifically: whether re-running a node duplicates a payment or a publish is a property of its *implementation*, not of a file. A saved graph asserting `retry_safe: true` for a node whose class never declared it would have granted the supervisor a retry on a side-effecting node — precisely the asymmetric failure design §5.3 exists to prevent.

Fixed in `Graph.loadFromDict` and node-sdk's `hydrateGraphNodeFlags`: the flag now comes from the registry or not at all, so an unresolved type gets no retry rather than the saved claim. `withExplicitNodeFlags` keeps the declared value — there the descriptor's author is code, not a saved file. Two tests pin both directions.

Also fixed: `decidedBy` now reports `"kernel"` when a verdict is rejected for falling outside the allowed set (crediting the handle hid the one case worth finding); `node:async_hooks` is loaded lazily through a shared module matching the runtime's existing pattern; the two literal NUL bytes in the sticky-cache keys are gone; the escalation error is narrowed properly; and the duplicated `retry_safe` key in the graph-resolver test is removed.

## Two CI legs are failing on `main`, not on this branch

I verified both by reverting `packages/{runtime,protocol,kernel,node-sdk}` to `origin/main` in a clean tree and re-running:

- **`test-packages`** → `@nodetool-ai/elevenlabs-nodes`: a `vi.mock` hoisting error (`Cannot access 'MockWebSocket' before initialization`) in `tests/realtime-{stt,tts}.test.ts`. This PR touches neither that package nor `python-websocket-bridge.ts`. Reproduces with my changes fully reverted.
- **`Workflow Runner Browser E2E`** → all 10 specs time out waiting for `workflowRunnerReady`, with `[browser pageerror] process is not defined`. Also reproduces with my changes fully reverted.

Worth noting on the second one: the `node:async_hooks` review comment predicted exactly this symptom, so I checked rather than assumed. The lazy-load fix alone does **not** make the E2E pass locally — the `process is not defined` error is upstream of it and present without this PR. I kept the lazy import anyway; it matches the package's established pattern and removes a duplicated fallback store.

The `quality` aggregator failures are rollups of these two legs (plus one earlier run cancelled mid-flight by a push).

Green locally on the current head: kernel 993, runtime 2644, node-sdk 691, core-nodes 237, base-nodes 2592; `build:packages` 60/60; lint and prettier clean.

---

PR #4640, opened 2026-08-02
Add workflow-tool evals with interactive escalation

georgi:
Every tool-loop case today hands the model an objective that is fully specified: everything it needs is in the prompt, so guessing is never required and never penalized. This adds a suite where it is — `nodetool eval workflow-escalation`.

Each case withholds something only the user can decide and gives the model an `ask_user` tool wired to a **scripted user**. The question is matched against the case's reply script, the matching reply comes back as the tool result, and every exchange is recorded — so a case scores both *that* the model escalated, *what* it asked about, and whether the graph it went on to build matches the answer it got back.

## Changes

- **`packages/agents/src/evals/escalation.ts`** — the scripted user. Replies match by RegExp or keyword list against the question (plus any `options`/`context`); an off-script question gets a deliberately useless fallback answer. `checkEscalationExpectations` is pure: ask counts, topics asked (`mustAsk`), no-off-script-asks, and `askBefore` — the confirm-before-you-act constraint that `ui_delete_node` must not precede the first `ask_user`.
- **`tool-loop-eval.ts`** — an optional per-case `escalation` config. When present the channel's tool joins the surface tools and its turns join the observation, so any tool-loop surface can opt in, not just the graph editor. Cases without it are unchanged.
- **`packages/agents/src/evals/escalation-cases.ts`** — five graph-editor cases over the existing headless bridge and node catalog (now exported as `TOOL_LOOP_NODE_CATALOG`):
  - `ask-for-missing-names` — input/output names withheld; must ask, then use `article`/`digest`.
  - `confirm-before-delete` — a dead branch in a seeded graph; must confirm before `ui_delete_node`, and must not touch the main chain.
  - `ask-which-step` — two node types fit equally; must ask, then add the one named and not the ones ruled out.
  - `escalate-missing-capability` — the objective needs an image node the catalog lacks; must escalate rather than invent a node type, then build the fallback the user describes.
  - `no-escalation-needed` — every value is pinned and `ask_user` is on the table; reaching for it is the failure. Without this case the suite would reward a model that asks about everything.
- **`packages/cli/src/commands/eval.ts`** — registered as a tool-loop suite, so it inherits `--list`, `--cases`, `--json`, `--out` and the `--min-success` CI gate.
- Docs: root `CLAUDE.md` suite list and a section in `packages/agents/CLAUDE.md`.

## Testing

`packages/agents/tests/escalation-tool-loop.test.ts` — 20 tests, scripted providers only, no network. Covers the channel (matching, fallback, per-run transcripts), the pure checker, the runner with and without an escalation config, and a **golden transcript per shipped case** asserting a 1.00 score, so no case can be silently unsatisfiable. Negative transcripts pin the failures too: deleting without asking, asking off-script, asking when nothing is open.

- `npm run test --workspace=packages/agents` — 1827 passed
- `npm run test --workspace=packages/cli` — 444 passed
- `npm run lint` — clean (pre-existing warnings only)
- `npm run build:packages` — clean; `npm run typecheck` clean for web/electron (mobile fails on absent `mobile/node_modules` in this environment, unrelated)
- `nodetool eval workflow-escalation --list` lists all five cases

No live-provider run is included — that costs inference and needs keys.

https://claude.ai/code/session_01Duxc2qAsBPtVSPXNCNNSaS

---
_Generated by [Claude Code](https://claude.ai/code/session_01Duxc2qAsBPtVSPXNCNNSaS)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -477,18 +477,27 @@ npm run dev:nodetool -- eval task-planner -p anthropic -m claude-sonnet-5
 npm run dev:nodetool -- eval script-planner -p openai -m gpt-5.4-mini
 ```
 
-Alongside `graph-planner` (one-shot DSL) there are seven **tool-loop** suites
+Alongside `graph-planner` (one-shot DSL) there are eight **tool-loop** suites
 that drive a real provider through the frontend `ui_*` tool contract against a
 headless bridge — no browser — and score the multi-turn tool-calling flow
-structurally: `tool-loop` (graph editor), `script-tools`, `sketch-tools`,
-`timeline-tools`, `storyboard-tools`, `model3d-tools`, `app-tools`. Same flags,
-metrics, and `--min-success` CI gate as `graph-planner`. Details:
+structurally: `tool-loop` (graph editor), `workflow-escalation`, `script-tools`,
+`sketch-tools`, `timeline-tools`, `storyboard-tools`, `model3d-tools`,
+`app-tools`. Same flags, metrics, and `--min-success` CI gate as
+`graph-planner`. Details:
 [packages/agents/CLAUDE.md](packages/agents/CLAUDE.md).
+
+`workflow-escalation` runs the graph tools over objectives that are missing
+something only the user can decide — a name, permission to delete, a choice
+between two node types — plus an `ask_user` tool wired to a scripted user. Each
+case scores both the question the model asked and whether the graph it went on
+to build matches the answer, and one case pins every value so that asking at all
+is the failure.
 
 ```bash
 npm run dev:nodetool -- eval timeline-tools --list
 npm run dev:nodetool -- eval script-tools -p anthropic -m claude-sonnet-5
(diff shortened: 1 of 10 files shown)


georgi (commented):
Reviewed the full diff — code looks good. Clean separation between the escalation channel, expectations checker, and the generic runner integration. Things I verified:

- The `lastIndex = 0` reset on global regex matchers prevents the match/miss alternation bug.
- `noInventedTypes` uses `Object.prototype.hasOwnProperty.call` correctly, so a node type like `"toString"` doesn't falsely pass.
- Golden transcripts pin every shipped case at 1.00, so no case can be unsatisfiable.
- `no-escalation-needed` guards the opposite failure mode (reward-hacking by asking about everything).
- The `TOOL_LOOP_NODE_CATALOG` export from `tool-loop-cases.ts` is a clean rename with no functional change to existing cases.
- No cross-PR conflicts with the three PRs merged in this session (#4636, #4637, #4638) — zero file overlap.

CI has 20/22 checks green with "examples" and "Load every route without errors" still running. Ready to merge once CI is fully green.

---

PR #4666, opened 2026-08-03
Close six gates that reported success without checking anything

georgi:
Hunt for signals that look like evidence and aren't: something a reader would reasonably take as "this works", with nothing behind it that establishes that.

Method was mutation, not reading: break the thing on purpose, confirm something goes red, revert. Every finding below has a red/green pair. Candidates that survived reading but died under mutation are listed at the bottom — one of them was mine.

---

## 1. Shipped app bundles had no registry gate

The only finding with user-visible breakage already in the tree.

`scripts/validate-examples.mjs` skipped `*.app.json`, with a comment deferring to `scripts/build-example-apps.mjs`, "which runs `nodetool app debug --no-run` over each one". Two things were wrong with that: `app debug` validates widget bindings and never consults the node registry (zero references to `validateGraph` in `packages/execution/src/app-debug/`), and no CI workflow runs `build-example-apps` at all.

**RED — the named delegate, on a bundle seeded with a node type that cannot exist:**
```
$ python3 -c "...replace nodetool.text.Prompt -> nodetool.totally.BogusDoesNotExist..."
$ nodetool app debug bogus.app.json --no-run
GATE_EXIT=0
✅ App wiring is valid (static check only — no run executed).
  app: Concept Studio · 36 widget(s)
```

The consequence had already shipped. 12 nodes across 5 installable bundles set `provider: "huggingface_fal_ai"`, which no package registers — the same string `packages/node-sdk/tests/graph-validation.test.ts:544` uses as its canonical *unregistered-provider* fixture. The source workflows these bundles were built from use `fal_ai`; the bundles were stale snapshots of graphs already fixed upstream. At runtime `getProvider` is an exact `Map.get` that throws — no alias layer, no fallback, and none of the five apps expose the `model` property for a user to override.

**Fix:** bundles now expand to their embedded graphs and run through the same validator the plain examples get; the 12 provider ids are corrected to `fal_ai`.

```
RED   (gate fixed, data still broken): 238/248 graphs clean, 10 failures, exit 1
GREEN (both fixed):                    248/248 graphs clean,             exit 0
```
Coverage went from 236 graphs to 248 — the 12 bundle graphs were previously ungated. The 10 failures landed on exactly the 5 suspected bundles.

## 2. `turbo --affected` pruned root-file-only PRs to zero test tasks

`--affected` scopes the graph by package ownership *before* input hashes are consulted, so turbo attributes a root-level file to the root workspace, which no package depends on. A PR touching only `Dockerfile`, `scripts/**`, or `web/src/**` ran no backend test and the leg reported green. This defeated the `$TURBO_ROOT$` entries in the `test` task's `inputs` added for exactly this purpose — only `globalDependencies` short-circuits the pruning.

Suites this silenced, each reading a fixture outside its own package: `websocket/dockerfile-examples` (root `Dockerfile`), `model-pricing/genspend-sync` (`scripts/genspend/*`), `agents/sandbox-manifest-drift` (`web/src/**` — i.e. exactly the drift it exists to detect).

```
RED   (original turbo.json, Dockerfile-only commit): test tasks = 0
GREEN (fix applied, same commit):                    test tasks = 58
```
Verified for all four paths (`Dockerfile`, `web/src/utils/codeOutputInference.ts`, `scripts/genspend/aliases.json`, `scripts/example-apps/apps.mjs`) — all now 58. Positive control: `tsconfig.base.json`, already in `globalDependencies`, yields the full graph both before and after.

## 3. `quality-guard.yml` reported every check as Passed

`if npm run typecheck 2>&1 | tee typecheck.log; then` tests **tee's** exit status, which is 0 whenever the log write succeeds. GitHub's default shell is `bash -e`, which does not set `pipefail`. `FAIL` never became 1 and `has-errors` was always `false`.

```
BEFORE (bash -e):              AFTER (shell: bash -> bash -eo pipefail):
  TypeScript: Passed             TypeScript: Failed
  has-errors=false               has-errors=true
  exit=0                         exit=1
```
The workflow has zero callers today, so this gates nothing — but the post-check block in the same file does it correctly, which marks this as an unnoticed copy-paste divergence rather than a deliberate choice.

## 4. `passWithNoTests` hid a vanished filter in base-nodes

Two blocking CI commands select tests by filename substring (the `parity` leg, and `test:integration`). With `passWithNoTests: true`, renaming a targeted file makes the filter match nothing and exit 0. base-nodes has 31 suites and does not need the flag; the genuinely-empty node packages keep it.

```
RED   nonexistent filter -> exit 0
GREEN nonexistent filter -> exit 1;  real `parity` filter still passes (exit 0)
```

## 5. `reliability/harness` ran in no workflow

25 suites, 144 tests, wired into nothing: `test:packages` filters to `./packages/*`, and root `test` covers only web/electron/mobile. Its own `scripts/reliability-ring1.mjs` justifies excluding two journeys from the Ring 1 loop because their happy paths are "covered by the harness's own tests (part of `npm run test`)" — `npm run test` never reached them, so those happy paths were covered by nothing.

```
RED   reliability harness test task present: False (58 tasks)
GREEN reliability harness test task present: True  (59 tasks)
Suite passes clean: 25 files, 144 passed | 5 skipped, exit 0
```

## 6. Docs asserting APIs that do not exist

`SimpleAgent` and `AgentExecutor` have **zero** occurrences in any `.ts`/`.tsx` under `packages/`, `web/`, `electron/`, and `packages/agents/src/agent-executor.ts` is not a file — yet four docs named them, including `.claude/skills/nodetool-agent-config/SKILL.md`, which instructs agents to pick between them. Corrected to the classes that exist (`TaskPlanner`, `TaskExecutor`, `ParallelTaskExecutor`, `StepExecutor`).

Chat's `--agent`/`--no-agent` flags and the `/agent` command were documented as functional with a mode list. The source marks both flags `[deprecated] No-op`, and `/agent` is not in `COMMANDS`. Every chat example in CLAUDE.md passed the dead flag, so a reader thinks they are selecting a planning mode when nothing happens. Also filled in five real providers the list omitted and the four real slash commands it lacked.

---

## Checked and found sound (negative results)

- **`packages/integration-nodes/tests/regression-api-data.test.ts:109`** — an un-awaited `expect(node.process()).rejects.toThrow(...)`. Reads as a dead assertion in a named regression test. It is not: removing the credential check from the source and rebuilding turns it **red** (exit 1). Vitest attributes floating rejections to the test. My first run appeared to confirm the bug — that was a stale-`dist` artifact, since the test imports the package, not `src/`. Rebuilding reversed the result.
- **`examples/workflows/**` under `--affected`** — genuinely covered. The always-on `parity` leg runs plain vitest with no turbo, so the suites reading those files are not silenced. Only the three cross-boundary suites in §2 were.
- **Snapshot tests** — none exist repo-wide; zero `toMatchSnapshot` calls, zero `.snap` files.
- **`.only`** — none in test code; `forbidOnly` is set for CI in the Playwright config.
- **`continue-on-error` / `|| true` in workflows** — every instance is on a diagnostic, cleanup, or explicitly-advisory step. No gate silently disarmed.
- **Smoke scripts** (`docker-smoke.mjs`, `smoke-backend-bundle.mjs`, `verify-backend-bundle.mjs`) — these assert properly and exit non-zero. `verify-backend-bundle.mjs` even treats an empty reference set as an error, which is the standard the rest of this PR applies elsewhere.

Two of my own verification commands were themselves false signals, which is the point of the method: `echo "EXIT=$?"` after a pipe (captures the pipe, not the command — used `${PIPESTATUS[0]}` after), and a `pgrep -f validate-examples.mjs` liveness check that matched **its own waiter shell's command line**, reporting "still running" for a job that had finished. Both are the exact shape being hunted.

## Not fixed, deliberately

Two larger candidates surfaced and are left alone rather than patched speculatively: `nodetool validate` checks a model's *provider* but never its *id*, so a retired model id in a shipped example passes the gate and fails at runtime; and a family of image nodes with unit-ambiguous parameters that silently no-op (pixel-vs-fraction offsets, enum values falling through to a default branch). Both are real, neither is a one-line fix, and both deserve their own change.

## Verification

`npm run lint` exit 0. `typecheck:web` and `typecheck:electron` exit 0. `npm run typecheck` fails only on `mobile/`, whose deps are not installed in this sandbox (it is deliberately not a root workspace); no change here touches mobile.

https://claude.ai/code/session_01JYhGMYc5daRdA9mbVR8Ue3

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYhGMYc5daRdA9mbVR8Ue3)_

### DIFF
--- a/.claude/skills/nodetool-agent-config/SKILL.md
+++ b/.claude/skills/nodetool-agent-config/SKILL.md
@@ -11,10 +11,10 @@ You help users create and configure NodeTool agents — autonomous AI systems th
 Objective → Agent → (Skill resolution → Planning → Execution) → Result
 ```
 
-Three agent classes (from `@nodetool-ai/agents`):
+Agent classes (from `@nodetool-ai/agents`):
 - **Agent**: Multi-step with planning, tool use, iterative execution
-- **SimpleAgent**: Single-step with structured output schema
-- **AgentExecutor**: Lightweight value extraction
+- **TaskPlanner**: Decompose an objective into a task DAG
+- **ParallelTaskExecutor**: Execute independent tasks of a plan concurrently
 
 # YAML Config Template
 
--- a/.github/workflows/quality-guard.yml
+++ b/.github/workflows/quality-guard.yml
@@ -56,6 +56,11 @@ jobs:
       - name: Run quality checks
         id: check
         continue-on-error: true
+        # `shell: bash` runs with `-eo pipefail`. Without it GitHub's default is
+        # `bash -e`, where `if npm run typecheck | tee log; then` tests *tee's*
+        # status — always 0 — so every check below reported Passed and
+        # `has-errors` was always false, whatever the commands did.
+        shell: bash
         run: |
           echo "## Pre-Flight Quality Check" >> $GITHUB_STEP_SUMMARY
           FAIL=0
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -115,7 +115,7 @@ protocol → config → security → auth → storage
 - **Styling**: MUI v7 + `sx` prop for one-off, `styled()` for reusable. Theme values only, no hardcoded colors/spacing. Prefer `FlexRow`/`FlexColumn` over `Box sx={{ display: "flex" }}` when the shorthand props (`gap`, `align`, `justify`) reduce verbosity; use `Box` directly when you have significant additional `sx` overrides anyway.
 - **Node graph**: ReactFlow 12. Nodes extend `BaseNode` from `@nodetool-ai/node-sdk`.
 - **LLM providers**: All in `packages/runtime/src/providers/` — Anthropic, OpenAI, Gemini, Ollama, Mistral, Groq, Claude Agent SDK
-- **Agent system**: `packages/agents/` — full planning agent (TaskPlanner → DAG of Steps), SimpleAgent (single-step), AgentExecutor (value extraction)
+- **Agent system**: `packages/agents/` — full planning agent (TaskPlanner → DAG of Steps), TaskExecutor/ParallelTaskExecutor (walk the DAG), StepExecutor (tool-calling loop for one step)
 - **Workflow execution**: Actor-model in `packages/kernel/` — DAG-based, message-passing between node actors
 - **Python bridge**: `PythonStdioBridge` in `packages/runtime/` — spawns `python -m nodetool.worker --stdio`, communicates via length-prefixed msgpack over stdin/stdout. Lazy-connected on first workflow with Python nodes.
 - **Serialization**: MsgPack for WebSocket messages, JSON for REST API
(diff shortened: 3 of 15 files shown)


georgi:
## CI status: one failure, and it isn't from this diff

**`examples` — green, and it proves the flagship fix.** The gate now expands app bundles and validates their embedded graphs:

```
Validating 223 example workflow(s) and 11 app bundle(s)...
  ok  .../apps/concept-studio.app.json › Pokemon Maker
  ok  .../apps/film-studio.app.json › Script to Screen
  ok  .../apps/product-launch-kit.app.json › Product Mockup Generator
  ...
248/248 graphs validate cleanly.
```

Those 12 bundle graphs were previously validated by nothing. On this branch before the data fix, the same gate reported `238/248, 10 failures, exit 1`.

**Also green:** `parity` (the `passWithNoTests` removal), `reliability` (the newly-wired `reliability/harness`), plus `typecheck`, `lint`, `build`, `deps`, `docker`, `bundle`, `app-build`, `test-app`, CodeQL, and both browser E2E jobs.

**`test-packages` — failed on a pre-existing Chrome flake.**

```
FAIL  packages/video-nodes/tests/model3d-render.test.ts
Error: connect ECONNREFUSED 127.0.0.1:36963      (25865ms)
Test Files  1 failed | 19 passed (20)
Tests       1 failed | 189 passed (190)
```

Four reasons this is not caused by the diff:

1. **No code path connects it.** This PR touches no `video-nodes` source and none of its dependencies. The only way it can reach that task is by changing its cache key.
2. **It passes locally**, on this branch: 20 files, 190 tests, exit 0 — including the headless render (878ms).
3. **The repo already knows about this exact failure.** `packages/video-nodes/src/nodes/model3d/render3d-headless.ts:117` documents it verbatim: the `ECONNREFUSED` "turned up as an `ECONNREFUSED` on CI while several other Chrome-driving suites were running." `connectCdp` exists specifically to retry it.
4. **It escaped that retry.** `connectCdp` wraps failures as `"RenderToImage: could not attach to headless Chrome on port N within Nms: ..."`. The CI error is the *bare* `connect ECONNREFUSED`, so it did not come from the guarded connect — it points at `launch()` or a Chrome that died mid-render, which the retry does not cover.

Why it surfaced here: editing `turbo.json` changes the global hash, so this PR ran the **entire** graph cold — `Cached: 50/105`, with several Chrome-driving suites executing concurrently — rather than the usual mostly-cached run. That load spike is a one-time cost of this PR touching `turbo.json`; later PRs won't pay it.

**Why I'm not pushing a fix.** Wrapping `launch()` in the same retry discipline is the obvious candidate, and I can't demonstrate it works: I can't reproduce CI's parallel-Chrome contention locally, so I'd have no red-before/green-after evidence — exactly the speculative build-system change this PR argues against. I'd rather leave it as a named unknown than ship a guess.

For the record, I chased a wrong cause first and discarded it: I suspected `dist/render3d-page.js` was missing from the downloaded `packages-dist` artifact, and reproduced a failure with the identical shape (1 failed / 6 passed → exit 1) by deleting that file. But that produces an asset-resolution error, not a CDP connection refusal. Matching symptom, wrong mechanism.

**Ask:** please re-run the `test-packages` job. `rerun-failed-jobs` returns `403 Resource not accessible by integration` for me. If it fails again on the same test I'll treat it as a real blocker and investigate the launch path properly; if it passes, the flake hypothesis holds and the branch is green.

---

PR #4672, opened 2026-08-03
fix: hydrate propertyTypes on the four surfaces still missing it

georgi:
Follow-up to #4671, which fixed `nodetool debug`. The same omission was in four more surfaces.

## The bug

`ExecutionSession` only fills `propertyTypes` when handed a `resolveNodeType`:

```ts
ExecutionSession.create({ graph, registry })                    // flags only
ExecutionSession.create({ graph, registry, resolveNodeType })   // + propertyTypes
```

Correlation analysis reads list-ness from that map and nothing else. On the first form every `list[...]` handle reads as non-list, so a stream arriving on one collapses to empty scope and the node runs once, on the last value. The run completes — it just quietly produces less than the same graph does under `workflows run`.

In #4671 that made `Directed Film to Timeline` look like it generated N keyframes and animated one:

```
workflows run   keyframe=2  animate=2
debug           keyframe=2  animate=1
```

The divergence was the harness, not the graph.

## Fixed here

| Surface | Was |
|---|---|
| `cli/src/evals/graph-runner.ts` | eval scores a graph the runner executes differently |
| `agents/src/workflow-agent.ts` | agent reports a result the graph doesn't produce |
| `websocket/src/headless-job-runner.ts` | headless jobs produce less than the editor |
| `websocket/src/lib/app-run-server.ts` | backs the app-debug and app-build services |

Checked per call site rather than inferred: none pre-hydrates. `normalizeGraph` fixes shape, not types, and `propertyTypes` appears nowhere in `packages/agents`.

**Not affected:** the six `unified-websocket-runner` call sites hydrate themselves via `Graph.loadFromDict` with their own resolver and pass no `registry`. I'd previously asserted this from a code comment; this time I read each one.

## The guard

A per-site fix doesn't hold a class of bug that has recurred five times. `packages/execution/tests/execution-session-hydration-audit.test.ts` walks every `ExecutionSession.create` in `packages/`, and fails on any that passes a `registry` without a `resolveNodeType`. Pre-hydrating callers pass no registry, so they're not matched.

Confirmed it catches a regression rather than passing vacuously — removing one fix:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+   "packages/websocket/src/headless-job-runner.ts"
```

It also asserts it can see more than 5 call sites, so it can't silently pass by matching nothing.

## Review this part hardest

Hydration resolves `outputs` too, so these surfaces now key results the way the runner does — by the node's **top-level `name`**, not `data.name`. Verified against `workflows run` with an Output node carrying `name: "TopLevelName"` and `data.name: "d"`:

```
output key: TopLevelName
```

`agent-graph-mode.test.ts` asserted parity against a baseline hydrated with `hydrateGraphNodeFlags` — flags only, which is not what any real run produces. Its baseline now hydrates like production, so the test checks the parity it claims to; the expected keys move from `d`/`b` to `Double`/`Boom` accordingly.

This is a user-visible change for anything consuming agent or headless-job outputs by key. It's the correct direction — these surfaces should agree with the runner — but it is a behavioural change, not just a bug fix.

## Verification

- execution **130 pass** · cli **462 pass** · agents **1996 pass**
- lint clean on all four packages · `tsc --noEmit` clean on all four
- audit test proven to fail without the fix, by filename

Docs: `correlation-design.md` gains a section on why a surface that skips hydration disagrees with the runner, with the before/after numbers — the breadcrumb the squash of #4671 dropped.

### DIFF
--- a/docs/correlation-design.md
+++ b/docs/correlation-design.md
@@ -724,6 +724,42 @@ Removed from production behavior:
 
 3. **Python worker rollout.** Python bridge compatibility remains a rollout concern for Python-backed nodes that need full output correlation metadata, lineage, `lineage_done`, and `lineage_scope_closed`.
 
+## Hydration: a surface that skips it disagrees with the runner
+
+Correlation analysis reads list-ness from `propertyTypes`, and nothing else.
+A saved workflow never carries that map — it is filled during hydration, from
+node metadata, and only on the path that gets a resolver:
+
+```ts
+ExecutionSession.create({ graph, registry })                       // flags only
+ExecutionSession.create({ graph, registry, resolveNodeType })      // + propertyTypes
+```
+
+On the first form every `list[...]` handle reads as non-list. A stream arriving
+on one collapses to empty scope and the node runs once, on the last value, with
+`received multiple values at empty scope` in the log. The run completes; it just
+produces less than the same graph does elsewhere.
+
+`nodetool debug` shipped on the first form. On one file:
+
+```
+workflows run   keyframe=2  animate=2
+debug           keyframe=2  animate=1
+```
+
+which made `Directed Film to Timeline` look like it generated N keyframes and
+animated one. It does that under `debug` and not under the runner, so the
+divergence was the harness, not the graph — after a debugging session spent
+looking for a kernel bug that was not there.
+
+Four other surfaces had the same omission: the eval graph runner, the workflow
+agent, the headless job runner, and the app run server. Callers that hydrate
+themselves (`Graph.loadFromDict` with their own resolver — the websocket
+runner) pass no `registry` and are unaffected.
+
+`packages/execution/tests/execution-session-hydration-audit.test.ts` fails, by
+filename, on any call site that passes a registry without a resolver.
+
 ## Known gaps and skipped tests
 
 These gaps are covered by `it.skip` regression tests that are kept (not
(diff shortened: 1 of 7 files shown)


georgi:
### Scope narrowed after CI: three surfaces, not four

CI caught what my local run missed — I linted `websocket` but never ran its tests. `app-debug-route.test.ts` and `app-build-route.test.ts` fail with the hydration applied, for exactly the output re-keying this PR flagged as the thing to review hardest.

**Mini-app widgets bind outputs by node id.** The fixture binds `op:main/out:out1` against a node `out1` whose `properties.name` is `text`. That node-id key is what the un-hydrated path produces; hydrating re-keys the output by the name, the binding stops resolving, and the harness reports *"bound to op:main/out:out1 but never received a value"* on a run that completed fine.

Answering which key an app document *should* bind by changing it underneath every existing app is not this PR's call. `app-run-server.ts` is reverted and recorded in the audit's `KNOWN_UNHYDRATED` map with the reason — debt, not exemption. A second assertion fails if an entry there stops being un-hydrated, so the list can't rot into a silent widening.

**Still fixed:** eval graph runner, workflow agent, headless job runner.

### Correcting my own earlier claim

I said these surfaces "now key results by the node's top-level `name`", from a probe where the Output node carried *both* a top-level `name` and a `data.name`. That shape is unusual. A node carrying only `data.name` — what 224 of 243 gallery Output nodes use — emits that key unchanged:

```
node: { type: "nodetool.output.Output", data: { name: "out1" } }
→ output_name = "out1"     (unchanged by hydration)
```

The re-keying only bites where the two disagree. That's why three surfaces were safe and this one wasn't, and my blanket statement was wider than the evidence.

### Verification

execution **131** · cli **462** · agents **1996** · websocket **2034** — all pass. Lint clean ×4.

Audit proven on both assertions: removing a fix names the file, and marking a known entry as hydrated flags it stale.

PR #4718, opened 2026-08-05
fix(reliability): anchor the lifecycle check on the final job terminal; give ws-transport-faults its own ring handling

georgi:
Three of the four red Ring 1 journeys. Both are harness/CI defects — no product code changes.

## 1. `mid-run-cancel-node` + `mid-run-cancel-streaming`

`cancelJob` answers the command with a terminal `job_update{status:"cancelled"}` immediately (`packages/websocket/src/unified-websocket-runner.ts:3827-3832`) and deliberately does not set `active.finished` — the comment at `:3834` says to let the drain deliver what's left. The node terminals follow. Proof they arrive: CI reports `lifecycle.running-after-job-terminal` but never `unmatched-running`, so the pairing closes by end of record.

`checkLifecycle` anchored on the FIRST terminal `job_update` (`lifecycle.ts:33`, guard `jobTerminalIndex === undefined`), which on cancel is the eager ack. It judged a converged run against a frame the protocol emits before convergence. Now it anchors on the last.

Duplicate terminals remain `terminal-uniqueness.ts`'s job — that separation is the stated design (`lifecycle.ts:11-14`).

**Why this doesn't blind the check.** A node with no terminal at all is still caught twice: by the anchor-independent end-of-record `unmatched-running` sweep (`:72-83`), and by being open at the last terminal. The only shape that stops being flagged is "node closes between two terminal `job_update`s" — which is either the documented eager-ack protocol or an undocumented duplicate terminal, and `terminal-uniqueness.multiple` catches the latter because untagged duplicates are counted. No violation class becomes unreachable.

Note the anchor must include `redundant`-tagged frames: the driver tags the *second* cancelled frame (`ws-server.ts:290-301`), so "last non-redundant" would re-select the eager ack and keep the bug.

## 2. `ws-transport-faults`

`reliability-ring1.mjs` ran it bare. A bare invocation applies every declared fault at once (`cli.ts:147-158` falls back to the full matrix), and the ws faults share one process-wide proxy slot (`faults/ws-proxy.ts:14-19`), so last-write-wins leaves a black-hole fault active — the client never sees a terminal and the driver's 4000ms timeout fires. That timeout is the designed behavior, not a slow path: an unfaulted run finishes in ~8ms.

The journey declares `["ws-server"]` only, and a per-fault ring loop cannot work either — `cli.ts:181` always injects the kernel oracle, which by design diverges from a black-holed client. Exclusion plus the existing per-fault vitest suite (`tests/journeys/ws-transport-faults.test.ts`, already in the Quality Gate) is the only correct shape. Added to `JOURNEYS_WITH_OWN_RING_HANDLING` alongside the two journeys already there for this reason.

## Red first

Two pushes on this branch:
- **Push 1** — tests only. `test-packages` fails with `lifecycle.running-after-job-terminal` on the converged-cancel record.
- **Push 2** — the fix. Green.

Run URLs in a comment below, so the flip is inspectable rather than asserted.

## Expected Ring 1 after this

`ok: mid-run-cancel-node`, `ok: mid-run-cancel-streaming`, and the `ws-transport-faults` block gone.

**`provider-failure-mid-stream` still fails** — separate root cause (`WsServerDriver` never calls `initTestDb`, so the autosave dedupe read throws and aborts the drain). Fixed in the next PR. Do not read that line as a failure of this one.

## Blast radius

`checkLifecycle` consumers: Ring 0 (kernel-only, single terminal — unaffected), Ring 1, `harness gate`. Every existing fixture in `tests/journeys/` and `tests/invariants/` has exactly one terminal `job_update`, so first == last and nothing flips. `tests/drivers/cli.test.ts` and `browser.test.ts` load the cancel journeys but assert status, not invariants.

### DIFF
--- a/reliability/harness/src/core/invariants/lifecycle.ts
+++ b/reliability/harness/src/core/invariants/lifecycle.ts
@@ -6,12 +6,20 @@
  *   that invocation — a node can run more than once in a record (loops,
  *   re-entrant control events), so pairing is tracked as a per-node stack,
  *   not a single flag.
- * - No node is left `running` after the `job_update` terminal.
+ * - No node is left `running` after the FINAL `job_update` terminal. A surface
+ *   may answer a cancel with an eager terminal ack and keep draining — the
+ *   ws-server relay does exactly that (`unified-websocket-runner.ts`'s
+ *   `cancelJob` announces the stop immediately and deliberately leaves
+ *   `active.finished` unset so the node terminals still arrive) — so the
+ *   authoritative terminal is the last one, not the first. Anchoring on the
+ *   first would call a run dirty for converging after an ack the protocol
+ *   sends before convergence.
  *
  * `terminal-uniqueness.ts` covers the job-level "exactly one terminal
  * `job_update`" and precedence rules — kept separate so each §6 bullet has
  * its own attributable module, per the task's "may fold into lifecycle if
- * cleaner, but keep the checks individually attributable" note.
+ * cleaner, but keep the checks individually attributable" note. Duplicate
+ * terminals stay its business: this module only asks where the run ended.
  */
 import type { RunRecord } from "../record.js";
 import type { Violation } from "./types.js";
@@ -29,11 +37,10 @@ export function checkLifecycle(record: RunRecord): Violation[] {
 
     if (type === "job_update") {
       const status = asString(message, "status");
-      if (
-        jobTerminalIndex === undefined &&
-        status !== undefined &&
-        JOB_TERMINAL_STATUSES.has(status)
-      ) {
+      // Last terminal wins (see the header note). Redundant-tagged frames
+      // count: the ws driver tags the SECOND cancelled frame, so skipping
+      // tagged ones would re-select the eager ack.
+      if (status !== undefined && JOB_TERMINAL_STATUSES.has(status)) {
         jobTerminalIndex = index;
       }
       return;
@@ -104,7 +111,7 @@ export function checkLifecycle(record: RunRecord): Violation[] {
       if (openCount <= 0) continue;
       violations.push({
         invariant: "lifecycle.running-after-job-terminal",
-        message: `node "${nodeId}" is still "running" at the job_update terminal (frame ${jobTerminalIndex})`,
+        message: `node "${nodeId}" is still "running" at the final job_update terminal (frame ${jobTerminalIndex})`,
         nodeId,
         frameIndex: jobTerminalIndex,
         details: { openCount }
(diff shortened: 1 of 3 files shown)


georgi (commented):
This PR is intentionally incomplete — it's "Push 1" of a two-push strategy, adding failing tests that demonstrate the lifecycle anchoring bug. The implementation fix to `lifecycle.ts` (anchoring on the last terminal `job_update` instead of the first) hasn't been pushed yet.

CI failures are expected:
- **test-packages**: The new test at `lifecycle.test.ts:105` fails because `checkLifecycle` still anchors on the first terminal (line 33: `jobTerminalIndex === undefined` guard).
- **quality**: Failed because it depends on test-packages.

The test code itself looks correct — it properly models the eager cancel ack scenario and the unclosed-node safety net. Will re-review once Push 2 lands with the fix.

Note: The second claim in the title ("give ws-transport-faults its own ring handling") appears to already be in place on main — `ws-transport-faults` is already in `JOURNEYS_WITH_OWN_RING_HANDLING` in `scripts/reliability-ring1.mjs` (line 69), and that file is not part of this diff.

---

georgi:
Closing — obsolete, and the lifecycle half would have been actively harmful.

**#4713 fixed this in the product while this PR was in flight**, by deleting the eager `job_update cancelled` frame. Its comment is the correct diagnosis:

> Deliberately NO eager `job_update cancelled` frame here: the kernel's own terminal frame relays through the drain loop AFTER the node-level terminal updates, and an out-of-band frame ahead of them tells the client the job is over while nodes still read "running" — the exact lifecycle violation the reliability harness's mid-run-cancel journeys pin (`lifecycle.running-after-job-terminal`), **and what left canvas nodes stuck spinning after a Stop.**

So `running-after-job-terminal` was never a false positive. It was reporting a real, user-visible defect — nodes spinning forever on the canvas after Stop — and the fix belonged in the emitter, not the checker. Relaxing the anchor to "last terminal" would have blinded the invariant to precisely the regression #4713 repaired.

The `ws-transport-faults` half is superseded too, and by a better fix: main excludes the journey from the bare loop *and* adds a per-fault ring block for the two recoverable faults (`ws-delay`, `ws-fragment`), keeping them in Ring 1. This PR dropped the journey from Ring 1 entirely.

`provider-failure-mid-stream` is fixed as well — the `Asset.paginate` replay-dedupe read is now best-effort, which is the right layer and makes the `initTestDb` harness patch unnecessary.

Ring 1 on main: **All 16 Ring 1 reliability run(s) passed.**

PR #4738, opened 2026-08-05
Add Alibaba Cloud (Model Studio / DashScope) LLM provider

georgi:
Adds an `alibaba` provider for Alibaba Cloud Model Studio (Qwen models) via the DashScope OpenAI-compatible endpoint.

## What's in it

**Runtime** (`packages/runtime`)
- `AlibabaProvider extends OpenAICompatProvider` — base URL `https://dashscope-intl.aliyuncs.com/compatible-mode/v1`, secret `DASHSCOPE_API_KEY`, live model listing from `GET /models` (graceful `[]` on failure), tool support on, GMI-style empty TTS/ASR/video/image/embedding lists so inherited OpenAI media models don't leak under the alibaba id
- Registered in the provider index, credential-check probe for the Verify Key button, Stryker mutate entry
- `PROVIDER_IDS.ALIBABA` in `@nodetool-ai/protocol`

**CLI** — `alibaba` in `KNOWN_PROVIDERS`, default model `qwen3-max`

**Settings** — `DASHSCOPE_API_KEY` declared in the websocket settings registry (Models & Providers page), pointing at the Model Studio console

**Web + Electron** — provider card in the settings catalog (existing `alibabacloud-color.svg` icon), model-menu icon maps, display name/URL formatting, required-secret gating in `ModelMenuStore`, onboarding card in `BootMessage`

**Docs + marketing** — capability matrix and prose in `docs/providers.md`, provider tables in `docs/models-and-providers.md` / `docs/architecture.md`, OpenAI-compatible developer guides, CLI provider list in `CLAUDE.md`, skills, marketing provider entry + logo domain

## Tests
- New `alibaba-provider.test.ts` (mirrors the GMI suite, incl. compat-fetch chat/stream cases) and an alibaba row in the OpenAI-subclass hardening table
- CLI provider/smoke/stub expectations updated
- Scoped runs pass: runtime provider suites (35 tests), registry/credential-check (36), websocket settings (26), full CLI suite (557)

https://claude.ai/code/session_01S3URpY3f2xoVx68M3Kawof

---
_Generated by [Claude Code](https://claude.ai/code/session_01S3URpY3f2xoVx68M3Kawof)_


georgi (commented):
I found three provider-compatibility issues that should be addressed before relying on the new Alibaba integration across the model picker and CLI.

packages/runtime/src/providers/alibaba-provider.ts
```diff
@@ -0,0 +1,112 @@
...
+  type ASRModel,
+  type EmbeddingModel,
+  type ImageModel,
+  type LanguageModel,
+  type TTSModel,
+  type VideoModel
+} from "./types.js";
+
+const ALIBABA_BASE_URL =
+  "https://dashscope-intl.aliyuncs.com/compatible-mode/v1";
```
georgi, line 18: [P1] Make the DashScope endpoint match the key's region

This hardcodes the legacy Singapore endpoint for both model listing and chat, while Model Studio API keys are region-specific. A key created in Frankfurt, Virginia, Tokyo, Beijing, or Hong Kong will be rejected here, and the settings UI exposes no endpoint/region control, so a valid key simply appears invalid. Alibaba's current endpoint table says the base URL varies by region and recommends workspace-specific domains: https://www.alibabacloud.com/help/en/model-studio/compatibility-of-openai-with-dashscope. Please make the base URL configurable (for example via `DASHSCOPE_BASE_URL`) and use the same resolved value here and in `credential-check.ts`, or explicitly scope the key flow to Singapore.

packages/runtime/src/providers/alibaba-provider.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    this._alibabaFetch = fetchFn;
+  }
+
+  override getContainerEnv(): Record<string, string> {
+    return { DASHSCOPE_API_KEY: this.apiKey };
+  }
+
+  override async hasToolSupport(_model: string): Promise<boolean> {
```
georgi: [P2] Do not advertise tool support for every listed model

`/models` is not a function-calling capability list, but this unconditional result is propagated to the model API and causes agents to send `tools` for every returned ID. Alibaba publishes a narrower function-calling model list (https://www.alibabacloud.com/help/en/model-studio/qwen-function-calling); chat-compatible families such as Qwen-Math/QVQ are not included, and Alibaba documents a 400 `The tool call is not supported` response for unsupported models. Please gate this per model (or filter the fetched catalog) and add at least one negative capability test.

packages/cli/src/providers.ts
```diff
@@ -105,6 +106,7 @@ export const DEFAULT_MODELS: Record<string, string> = {
   minimax: "MiniMax-M2",
   cerebras: "llama-3.3-70b",
   gmi: "meta-llama/Llama-3.3-70B-Instruct",
+  alibaba: "qwen3-max",
```
georgi, line 109: [P1] Avoid the inherited token-limit parameter for this default

`OpenAICompatProvider.buildChatRequest()` always sends `max_completion_tokens`, including its default value, but Alibaba documents that parameter for Qwen Max only from Qwen3.7-Max onward; `qwen3-max` is not in the supported list and older models use `max_tokens`: https://www.alibabacloud.com/help/en/model-studio/qwen-api-via-openai-chat-completions. Consequently the out-of-box CLI path constructs an unsupported request parameter. Either choose a default documented to accept `max_completion_tokens`, or override Alibaba request assembly so the token-limit field is selected per model.

georgi:
All three review points are addressed in the latest push:

- **Region-scoped endpoint** — new `DASHSCOPE_BASE_URL` setting (also read from env), resolved by one shared `resolveAlibabaBaseURL()` used by the provider, its `/models` listing, `getContainerEnv`, and the credential-check probe, so key verification hits the same endpoint chat does. Default stays the international/Singapore endpoint; documented in `docs/providers.md`.
- **Token-limit parameter** — `OpenAICompatProvider` gained a `transformChatRequest` hook (identity by default); the Alibaba provider uses it to send `max_tokens` instead of `max_completion_tokens`, which every Qwen chat model accepts, so the `qwen3-max` CLI default no longer constructs an unsupported parameter. Covered by a request-body test.
- **Tool-support gating** — `hasToolSupport` now denies the families outside Alibaba's function-calling list (Qwen-Math, QVQ, Qwen-VL-OCR, Qwen-MT, audio/TTS, embeddings) by prefix, with negative capability tests; everything else stays enabled so newly listed chat models keep working.

The push also adds the provider contract cassette the `test-packages` CI leg flagged. Runtime (2699), websocket (2201), and CLI (557) suites pass locally.

---

PR #4760, opened 2026-08-06
feat(credits): credit ledger, subscription plans, and a metered nodetool provider

georgi:
## What

Credit management and subscription plans for the Studio product (follow-up to #4758), with metering modeled as a **provider**, not a deployment switch: NodeTool's own managed models are a real provider (`nodetool`) whose spend is metered against the user's credit balance, while BYOK providers coexist unmetered on the same server.

## Design

- **Balance = grants − spend.** `nodetool_credit_ledger` stores grants only (plan accruals, top-ups, adjustments). Spend is read from the `nodetool_predictions` rows every provider call already writes, at 1 credit = $0.01 — no double bookkeeping.
- **Plans without a cron.** `nodetool_user_subscriptions` holds one row per user (Free 300/mo, Creator 3,000/mo $12, Pro 10,000/mo $40). Monthly grants accrue lazily on the first status read of the month, keyed `plan:<user>:<plan>:<YYYY-MM>` so the primary key makes double accrual impossible.
- **The `nodetool` provider is what gets metered.** Each curated model (`NODETOOL_MODELS` in protocol) names a delegate provider+model; `NodetoolProvider` runs the delegate on *platform-owned* keys (`NODETOOL_PLATFORM_FAL_KEY`, `NODETOOL_PLATFORM_ANTHROPIC_KEY`), absorbs the delegate's cost at the delegate's price, and lists only funded models — the provider appears in pickers exactly when the server can serve it. `model-pricing` translates `nodetool/...` ids to the delegate before catalog lookup, so estimates are real numbers.
- **Enforcement follows the provider.** Workflow runs are gated only on the `nodetool` slice of their cost estimate (`estimateNodetoolSpend` → `admitCreditRun`, beside the existing application-budget gate); the direct `generate_media`/`transcribe_audio` RPCs are gated only when called with `provider: "nodetool"`. BYOK is never gated. The gate fails open on its own errors, like the app-budget gate.
- **Payments are a stub on purpose.** `credits.topup` adds credits with no payment behind it and the UI says so; a payment provider integration replaces that mutation with a checkout session writing the same ledger rows from its webhook.

## Changes

- `packages/protocol`: `PROVIDER_IDS.NODETOOL`, curated catalog `nodetool-models.ts`, `api-schemas/credits.ts` (status reports `meteredProvider`), cloud-profile inclusion.
- `packages/runtime`: `providers/nodetool-provider.ts` (delegating provider, cost absorption, funded-model listing) + registration.
- `packages/model-pricing`: nodetool→delegate id translation in `getModelUnitPrice`.
- `packages/models`: `credits.ts` (catalog, subscription, accrual, balance, `checkCredits`), schema + pg mirror, migration `20260806_000000`, test-DB DDL.
- `packages/websocket`: `credit-gate.ts`, `trpc/routers/credits.ts`, provider-scoped gating in `unified-websocket-runner.ts`.
- `web/src/studio`: curated models point at the `nodetool` provider; server-backed credits chip; `/studio/account` page (balance, usage, plan cards, test top-up); provider display name.
- `docs/agentic-video-product.md` rewritten to match.

## Verification

- Tests: credits model (7), credit gate (3), NodetoolProvider (4 — funded filtering, provider stamping, missing-key and unknown-id errors), pricing translation (2), migration-count bump. Full `models` suite green (814); websocket suite green (2,211); web typecheck + eslint green.
- Live server smoke: with `NODETOOL_PLATFORM_FAL_KEY` set, `models.imageByProvider`/`videoByProvider` for `nodetool` return exactly the funded curated models stamped `provider: "nodetool"`; the anthropic-delegated director is absent without its key. `/studio/account` renders; Free→Creator switch moves the balance 300 → 3,300 instantly.

https://claude.ai/code/session_011GwuPTt5T1sLkmpwWUt2N2

### DIFF
--- a/packages/models/src/credits.ts
+++ b/packages/models/src/credits.ts
@@ -12,7 +12,8 @@
  * would write `topup` rows and flip `plan_id`; until then plan switches are
  * instant and top-ups are stubbed.
  */
-import { eq, sql } from "drizzle-orm";
+import { and, eq, sql } from "drizzle-orm";
+import { NODETOOL_PROVIDER_ID } from "@nodetool-ai/protocol";
 
 import { getDb } from "./db.js";
 import { creditLedger, userSubscriptions } from "./schema/credits.js";
@@ -222,10 +223,18 @@ export async function creditStatus(userId: string): Promise<CreditStatus> {
     .where(eq(creditLedger.user_id, userId));
   const grantedCredits = Number(grantRows[0]?.total ?? 0);
 
+  // Only the managed provider's spend counts against credits — BYOK
+  // predictions (fal_ai, anthropic, …) ride the user's own keys and must
+  // never drain the balance.
   const spendRows = await db
     .select({ total: sql<number>`COALESCE(SUM(${predictions.cost}), 0)` })
     .from(predictions)
-    .where(eq(predictions.user_id, userId));
+    .where(
+      and(
+        eq(predictions.user_id, userId),
+        eq(predictions.provider, NODETOOL_PROVIDER_ID)
+      )
+    );
   const spentUsd = Number(spendRows[0]?.total ?? 0);
   const spentCredits = Math.ceil(spentUsd / USD_PER_CREDIT);
 
--- a/packages/models/src/credits.ts
+++ b/packages/models/src/credits.ts
@@ -1,0 +1,272 @@
+/**
+ * User credits and subscription plans — the Studio product's billing layer.
+ *
+ * The ledger stores grants only. Spend is never written here: every provider
+ * call already lands in `nodetool_predictions` with its USD cost, so
+ *
+ *   balance = sum(ledger.delta) - ceil(prediction spend / USD_PER_CREDIT)
+ *
+ * Plans accrue lazily: the first balance read in a month inserts that month's
+ * grant, keyed `plan:<userId>:<periodKey>` so the primary key makes double
+ * accrual impossible. No cron, no payment state — a payment provider webhook
+ * would write `topup` rows and flip `plan_id`; until then plan switches are
+ * instant and top-ups are stubbed.
+ */
+import { eq, sql } from "drizzle-orm";
+
+import { getDb } from "./db.js";
+import { creditLedger, userSubscriptions } from "./schema/credits.js";
+import { predictions } from "./schema/predictions.js";
+import { createTimeOrderedUuid } from "./base-model.js";
+
+/** One credit is one US cent of provider spend. */
+export const USD_PER_CREDIT = 0.01;
+
+export interface CreditPlan {
+  id: string;
+  name: string;
+  /** Credits granted at the start of each calendar month (UTC). */
+  monthlyCredits: number;
+  /** Display price; billing itself is not implemented. */
+  priceUsdPerMonth: number;
+  blurb: string;
+}
+
+export const CREDIT_PLANS: readonly CreditPlan[] = [
+  {
+    id: "free",
+    name: "Free",
+    monthlyCredits: 300,
+    priceUsdPerMonth: 0,
+    blurb: "Try both creation paths with a monthly starter allowance."
+  },
+  {
+    id: "creator",
+    name: "Creator",
+    monthlyCredits: 3_000,
+    priceUsdPerMonth: 12,
+    blurb: "Enough for regular short-form work: stills, clips, and voice."
+  },
+  {
+    id: "pro",
+    name: "Pro",
+    monthlyCredits: 10_000,
+    priceUsdPerMonth: 40,
+    blurb: "Headroom for daily production and longer cuts."
+  }
+] as const;
+
+export const DEFAULT_PLAN_ID = "free";
+
+export const planById = (planId: string): CreditPlan | null =>
+  CREDIT_PLANS.find((p) => p.id === planId) ?? null;
+
+/** Calendar-month key, UTC — "2026-08". */
+export const periodKeyFor = (now: Date): string =>
+  `${now.getUTCFullYear()}-${String(now.getUTCMonth() + 1).padStart(2, "0")}`;
+
+export interface UserSubscription {
+  userId: string;
+  planId: string;
+  status: string;
+  updatedAt: string;
+}
+
+export interface CreditStatus {
+  userId: string;
+  plan: CreditPlan;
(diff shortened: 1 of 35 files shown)


packages/models/src/credits.ts
```diff
@@ -0,0 +1,272 @@
...
+      total: sql<number>`COALESCE(SUM(${creditLedger.delta}), 0)`
+    })
+    .from(creditLedger)
+    .where(eq(creditLedger.user_id, userId));
+  const grantedCredits = Number(grantRows[0]?.total ?? 0);
+
+  const spendRows = await db
+    .select({ total: sql<number>`COALESCE(SUM(${predictions.cost}), 0)` })
+    .from(predictions)
+    .where(eq(predictions.user_id, userId));
```
georgi: This aggregation needs to be scoped to `provider = "nodetool"`. As written, every prediction for the user, including `fal_ai`, `anthropic`, and other BYOK calls, is deducted from the managed credit balance. Existing users with historical BYOK spend can therefore start at zero, and future BYOK calls continue draining credits, contradicting the provider-scoped contract. Please add the provider predicate and a regression test with both managed and BYOK predictions.

packages/websocket/src/trpc/routers/credits.ts
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Prototype top-up: adds credits with no payment behind it. A payment
+   * provider integration replaces this mutation with a checkout session and
+   * writes the ledger row from the webhook instead.
+   */
+  topup: protectedProcedure
+    .input(topupInput)
+    .output(creditStatusOutput)
+    .mutation(async ({ ctx, input }) => {
+      await grantCredits(
```
georgi, line 80: This lets any authenticated caller mint up to 50,000 credits per request, repeatedly. Because those credits unlock calls made with platform-owned keys, the RPC makes managed provider spend effectively unlimited; labeling the UI button as a test does not protect the endpoint. Please remove or gate this behind an explicit development-only configuration, or make grants admin/payment-webhook-only before exposing the router.

packages/websocket/src/unified-websocket-runner.ts
```diff
@@ -7576,6 +7638,15 @@ export class UnifiedWebSocketRunner {
     }
 
     const userId = this.userId ?? "1";
+    // Only NodeTool's managed provider is metered; BYOK providers run on the
+    // user's own keys. Direct generation has no per-node estimate, so the
+    // gate refuses on an empty balance.
+    if (req.provider === "nodetool") {
+      const creditDecision = await admitSpend(userId, 0);
```
georgi: A zero-cost check only verifies that the balance is currently positive. This direct RPC does not write a `Prediction`, so a user who has received the free grant can call `generate_media` indefinitely through the platform key, including the variations fan-out, without ever reducing the balance. The same issue exists in `transcribe_audio`. Please reserve and record/settle the actual provider cost for these paths, or disable the managed provider here until that accounting exists.

packages/websocket/src/unified-websocket-runner.ts
```diff
@@ -3059,9 +3098,32 @@ export class UnifiedWebSocketRunner {
...
+      req.job_id ?? randomUUID(),
+      ApiErrorCode.BUDGET_EXCEEDED,
+      decision.reason
+    );
+  }
+
   async runJob(req: RunJobRequest): Promise<void> {
     req._accepted_at_ms ??= performance.now();
     if (!(await this.admitApplicationRun(req))) return;
+    if (!(await this.admitCreditRun(req))) return;
```
georgi, line 3151: This check happens before concurrency queueing and does not reserve credits. Multiple submissions can all pass against the same balance, then queued jobs execute later without another check after earlier jobs have spent it. A small balance can therefore authorize substantially more platform spend than it contains. Please atomically reserve estimated credits per accepted run and settle against actual cost, or at minimum recheck/reserve when a queued run is dequeued, following the application-budget pattern.

packages/runtime/src/providers/nodetool-provider.ts
```diff
@@ -0,0 +1,232 @@
...
+  /** Run a delegated call and absorb the delegate's cost delta as our own. */
+  private async absorbing<T>(
+    inner: BaseProvider,
+    call: () => Promise<T>
+  ): Promise<T> {
+    const before = inner.getTotalCost();
+    try {
+      return await call();
+    } finally {
+      this._absorbedCost += inner.getTotalCost() - before;
```
georgi, line 102: The cached inner provider can serve overlapping calls, so measuring a call via the shared cumulative `getTotalCost()` is not concurrency-safe. If A and B both read 0, A adds `x`, then B finishes at `x + y`, the outer provider absorbs `x` and then `x + y`, double-counting A. `generateMessages` has the same pattern. Please use request-scoped cost/usage, a fresh delegate per concurrent call, or serialize accounting around each shared delegate.

georgi (commented):
I found five accounting and enforcement issues that should be addressed before this is used with platform-owned provider keys. Inline comments include concrete failure modes and suggested fixes.

georgi:
All five findings addressed in ddfd5fb:

1. **BYOK spend draining credits** — the balance's spend SUM is now scoped to `provider = "nodetool"`, with a regression test mixing managed and BYOK prediction rows.
2. **Open credit mint** — `credits.topup` throws `FORBIDDEN` unless the operator sets `NODETOOL_ENABLE_TEST_TOPUP`; `credits.status` reports the flag and the UI hides the button when it's off.
3. **Unaccounted direct RPCs** — `generate_media` on the managed provider now admits against a unit-price estimate (variations included), reserves it for the call, and writes a prediction row at `max(delegate-tracked cost, unit estimate)` so the balance decrements; `transcribe_audio` records delegate-tracked cost the same way.
4. **Queue/concurrency race** — admitted runs reserve their `nodetool` estimate under the job id and release at terminal state or cancel-while-queued; `admitSpend` counts in-flight reservations, so N submissions can no longer all pass against the same balance. Reservations are process-local with a TTL backstop — the comment notes a multi-instance deployment should move them into the DB like `application_budgets`.
5. **Delegate cost double-count** — `NodetoolProvider` constructs a fresh delegate per call (delegates build clients lazily, so this is cheap), making the cost-absorption window race-free.

---

PR #4776, opened 2026-08-07
Extend the codeact eval suite to cover the full nodetool.* API surface

georgi:
Extends the `codeact` eval suite from four toy-toolbelt cases to full coverage of the `nodetool.*` object-model API surface, then uses live runs of the extended suite to fix what they exposed in the codeact prompts and the finish bridge.

## Eval suite

- **Harness**: `CodeActEvalCase` gains `createTools(recorder)` (per-case instrumented toolbelts), `maxIterations`, and `namespaces`; `RecordingTool` is exported for reuse.
- **19 new offline cases** across two deterministic in-memory worlds, with fakes named like real belt tools so the executor lights up the `nodetool.*` prelude:
  - `codeact-api-core.ts` — workflows, ad-hoc graph builder (executable five-node catalog), nodes, models, media incl. the critique judge, jobs (background + `wait`), assets, sub-agents, `nodetool.batch`.
  - `codeact-api-surfaces.ts` — collections/RAG, memory, web (with an unconfigured backend so fallback routing is real), documents, email, style, apps, timelines, sketches, scripts, storyboards.
- Ordering constraints live in the worlds, not the prompts, so result checks stay route-agnostic. A coverage test fails when any `nodetool.*` namespace loses its last case; scripted-provider tests pin each world's argument names to what the prelude actually sends.
- `scripts/dump-codeact-run.ts` replays one case against a real provider and writes every action's code + bridged calls to `nodetool-debug/` — the tuning loop's microscope.

## What the live runs taught, and what changed

Measured on `claude_agent_sdk`/sonnet (`IS_SANDBOX=1 … --max-iterations 40`), seven iterations. First full run: 17/23 accepted, 4.4 actions/case (smoke: 12.5). Final full runs: 20–22/23 per run with the misses rotating (sampling variance — every case passes under the final prompts in most runs), mean score 0.98–0.99, 2.3 actions/case.

- **One-call-per-action runs**: every early failure passed its result checks but blew action budgets — one tool call per action, a separate finish-only turn, protocol rounds (job polls, escalation resolves) spread one per action. The action contract now names the anti-pattern, requires `finish()` in the computing action, and frames protocol loops as loops. Escalation resolution went from 10 actions to 1.
- **Shape blindness**: per-action dumps showed perfect one-action pipelines dying on undocumented return envelopes (`calculate` → `{value}`, `workflows.list()` → `{workflows}`), then burning 2–6 actions re-probing. `finish()` is now documented as catch-and-repair-in-the-same-action, unfamiliar values get logged before extraction, and the workflows/models docs state the envelopes their real backing tools return.
- **Schema-valid garbage**: the chaining pressure produced `"[object Object]"`, statuses standing in for outputs, and JSON-serialized envelopes in string fields — all type-correct. The finish bridge now rejects coerced/serialized envelopes (naming the offending paths) so the in-action repair loop fires; the prompt adds types-not-truth and innermost-value warnings.
- **Objective ambiguity**: remaining misses were reasonable readings of loose case wording (adherence report as critique verdict, asset record as file body, delete-and-recreate as in-place update, filtered counts as totals); those objectives now name the value and route they measure.

## CI note

The `test-app` leg is red on a pre-existing base failure (`CodeBody.test.tsx` Ask-AI visibility, broken by main's `wip: CodeAct-only execution mode` commit) — reproduced on the unmodified base, unrelated to this diff. All other legs are green.

https://claude.ai/code/session_01Bt5fRAzzELYhSo4ChWs3BG


georgi:
CI note: the `test-app` leg is red on a pre-existing base failure, not this diff. `CodeBody.test.tsx` ("hides Ask AI on a connected Code node" + one sibling) fails at this branch's base commit (`c0f95e5`, current main head) with zero `web/` changes here — the recent `wip: CodeAct-only execution mode` commit changed CodeBody's Ask-AI visibility without updating its test. Reproduced locally: `cd web && npx jest src/components/node_types/__tests__/CodeBody.test.tsx` fails identically on the unmodified base. Will re-run CI when the base recovers.

---

PR #4787, opened 2026-08-08
fix(security): block IPv6-mapped and decimal-encoded IPs in SSRF guard

bunlongheng:
isSafeExternalUrl() only checked dotted-quad IPv4 literals for private/loopback ranges. Two bypass forms slip through: IPv4-mapped IPv6 (::ffff:192.168.1.1, which WHATWG URL serialises as ::ffff:c0a8:101) and decimal-encoded IPv4 (2130706433 = 127.0.0.1). Added normalization that unwraps ::ffff: prefixes in both dotted-quad and hex form, and converts all-digit hostnames to dotted-quad before the existing range check, so both encoding forms get caught by the same private-range logic. No async DNS - purely synchronous pattern matching.

### DIFF
--- a/packages/websocket/src/unified-websocket-runner.ts
+++ b/packages/websocket/src/unified-websocket-runner.ts
@@ -333,6 +333,17 @@ function isSafeExternalUrl(url: string): boolean {
   ) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


georgi:
@claude review

PR #4798, opened 2026-08-09
Derive the sandbox manifest's globals from the real QuickJS guest

georgi:
## The bug

`partitionNativeGlobals()` in `sandbox-manifest.ts` derived `nativeGlobals` / `blockedGlobals` from `Object.entries(buildSandbox().sandbox)` — the **host-side intent record**. That is neither what `@sebastianwessel/quickjs` actually installs in the guest, nor does it include QuickJS's own built-ins. Probing the live guest with `runInSandbox({code: 'return Object.getOwnPropertyNames(globalThis)'})` showed drift in both directions.

**3 phantom names** — documented in every prompt and both mirror sets, but `typeof === "undefined"` in the guest:

```
btoa: "undefined"   atob: "undefined"   structuredClone: "undefined"
```

A model reaching for `btoa(x)` got a ReferenceError and nothing flagged it — precisely the failure the manifest exists to prevent.

**37 real guest globals were undocumented**, including every typed array: `Uint8Array`, `ArrayBuffer`, `DataView`, `Int8Array`…`Float64Array`, `BigInt`, `Proxy`, `Reflect`, `ReferenceError`, `WeakRef`, `performance`, `queueMicrotask`, `Headers`/`Request`/`Response`, `Buffer`, `process`, `env`.

`Uint8Array` is the documented return type of `fromBase64`, `crypto.digest`, `crypto.getRandomValues`, `workspace.readBytes`, `data.unzip` and all ten `image.*` members. Before this change:

```
validateCodeNodeBody({ code: 'const a = new Uint8Array(4); const v = new DataView(a.buffer); …' })
→ [{ severity: "error", code: "code_undefined_name",
     message: 'The code reads "BigInt", "DataView", "Uint8Array", "performance",
               which are neither a sandbox API nor an input of this node…' }]
```

Downstream of the same set, web's `inferInputs` would invent a dynamic input slot named `Uint8Array`, and the `code-gen` eval scored a model down via `unknownApiReferences` for writing correct code.

The drift test **locked the bug in**: it pins both mirror sets to `documentedGlobals()` bidirectionally, so adding `Uint8Array` to `SANDBOX_GLOBALS` failed the test, and so did removing `btoa`.

## The fix

The guest global list is now a checked-in snapshot (`GUEST_GLOBALS_SNAPSHOT`, 89 names) regenerated by `packages/agents/scripts/probe-guest-globals.ts`, with a new drift test that boots the real sandbox and asserts the snapshot still equals the guest's `globalThis`.

`getSandboxManifest()` stays **synchronous**. Booting QuickJS is async, but the function is called synchronously from every prompt builder, the graph validator path and the evals — making it async was not worth the blast radius, so the snapshot plus a live-probe test carries the guarantee instead.

Names other runtimes have and this guest lacks (`Intl`, `AbortController`, `Blob`, `FormData`, plus the three phantoms) move into `blockedGlobals`, whose documented meaning is exactly that. A body reading one now gets an error naming the sandbox equivalent rather than silence:

> `btoa("x")` → *"which other JavaScript runtimes have but this sandbox does not — it throws ReferenceError. Use the sandbox equivalent: toBase64/fromBase64 for base64…"*

Both mirrors (`packages/node-sdk`, `web/`) are updated so the bidirectional assertions now lock in the correct set. Final counts: 69 native, 15 blocked, 106 documented names.

## Also: three gaps around the CodeAct prompt

`buildCodeActSystemPrompt` is built for **every agent step and every chat turn** and had no drift guard at all.

- **`CODEACT_INJECTED_GLOBALS`** declares the guest names the executors install (`tools`, `state`, `finish`, `searchTools`, and the object-model preludes' `nodetool` / `openWorkflow`) next to the code installing them, so the manifest check can reach the injected half of the surface. The new `codeact-prompt-drift.test.ts` fails when the contract names an API the sandbox lacks — verified by appending "`lodash.groupBy(rows)`" to the action contract and watching both variants fail with `expected ['lodash'] to deeply equal []`.
- **`manifest.notes` now render into the CodeAct summary.** The prompt never said `Intl` is absent, which is the entire reason the `format.*` bridge exists. The hand-copied "no module loader" sentence is deleted rather than kept alongside the note — a hand-copy is the drift this design exists to prevent.
- **The chat variant's hidden bridges are derived** from the `requiresContext` flag the manifest already carries, leaving `fetch` and `getSecret` as a commented exception with the reason. A test pins the derived set against what `createChatCodeActSession` actually enforces (no `context`, `maxFetchCalls: 0`), linking assertion to enforcement for the first time. The chat contract also claimed `fetch` was "disabled" when `maxFetchCalls: 0` means it exists and throws.

## Verification

Manifest and guest now agree in all three directions:

```
PRESENT but undocumented:        (none)
DOCUMENTED as native but absent: (none)
BLOCKED but actually present:    (none)

repro Uint8Array/DataView/BigInt/performance -> []          (was 1 error naming all four)
btoa now flagged                              -> code_undefined_name
guest runs the repro                          -> ok
guest btoa                                    -> ReferenceError: 'btoa' is not defined
```

- `npm run test --workspace=packages/agents` — 152 files, 2185 passed, 1 skipped
- `npm run test --workspace=packages/node-sdk` — 51 files, 756 passed, 2 skipped
- `cd web && npx jest src/utils` — 150 suites, 2202 passed
- `npm run lint` — clean
- `npm run typecheck` — web, electron and packages clean. The mobile leg fails only on missing `react-native`/`expo` modules: `mobile/node_modules` is empty in this environment and no file in this diff is under `mobile/`.

## Known gap left open

The `extraSections` (the graph-model and `nodetool.*` prompt sections) are still outside the drift check. Their examples bind local names (`wf`, `g`, `rows`) that would need a broad allowlist, which would weaken the check for the parts it does cover. Worth a follow-up that teaches the extractor about locally-bound names rather than allowlisting around it.

https://claude.ai/code/session_01XQwBvmV7Wjxn5RFR8E1Mst

---
_Generated by [Claude Code](https://claude.ai/code/session_01XQwBvmV7Wjxn5RFR8E1Mst)_

### DIFF
--- a/packages/agents/scripts/probe-guest-globals.ts
+++ b/packages/agents/scripts/probe-guest-globals.ts
@@ -1,0 +1,19 @@
+/**
+ * Print the QuickJS guest's own globals as the snapshot literal
+ * `GUEST_GLOBALS_SNAPSHOT` in `src/code-gen/sandbox-manifest.ts` expects.
+ * Run after upgrading `@sebastianwessel/quickjs` or changing the prelude:
+ *
+ *     npx tsx packages/agents/scripts/probe-guest-globals.ts
+ */
+import { runInSandbox } from "../src/js-sandbox.js";
+
+const run = await runInSandbox({
+  code: "return Object.getOwnPropertyNames(globalThis)"
+});
+if (run.error) throw new Error(run.error);
+
+const names = (run.result as string[])
+  .filter((name) => !name.startsWith("__"))
+  .sort();
+
+console.log(names.map((name) => `  ${JSON.stringify(name)}`).join(",\n"));
--- a/packages/agents/src/code-gen/sandbox-manifest.ts
+++ b/packages/agents/src/code-gen/sandbox-manifest.ts
@@ -1,16 +1,18 @@
 /**
  * Serializable description of the guest surface `nodetool.code.Code` runs on.
  *
- * Every name and number here is read out of `js-sandbox.ts` rather than
- * restated, so a prompt or help text derived from the manifest cannot advertise
- * an API the sandbox does not marshal. Only the human prose (signatures,
- * descriptions) is written here.
+ * Every bridge, helper and limit here is read out of `js-sandbox.ts` rather
+ * than restated, so a prompt or help text derived from the manifest cannot
+ * advertise an API the sandbox does not marshal. Only the human prose
+ * (signatures, descriptions) and `GUEST_GLOBALS_SNAPSHOT` — which no
+ * synchronous host call can answer — are written here.
  */
 import {
   buildSandbox,
   resolveSandboxLimits,
   DELETED_GUEST_GLOBALS,
   EXPOSED_BRIDGE_NAMES,
+  GUEST_HELPER_NAMES,
   DEFAULT_TIMEOUT_MS,
   DEFAULT_FORMAT_LOCALE,
   DEFAULT_SELECT_HTML_LIMIT,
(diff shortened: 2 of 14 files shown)


georgi (commented):
Code review: the changes look correct and well-structured. Ready to merge once the Playwright visual snapshots check completes.

**What this does:** Derives the sandbox manifest's global list from the actual QuickJS guest (via a checked-in snapshot and a live-probe drift test), rather than inferring it from the host-side `buildSandbox` record. Renames `CODEACT_RESERVED_NAMES` → `CODEACT_INJECTED_GLOBALS` and expands it to cover the graph-model and nodetool-api preludes. Adds drift tests for the CodeAct prompt and the chat variant's exclusion list.

**Verified:**
- `CODEACT_RESERVED_NAMES` has no consumers outside `packages/agents/` — the rename is safe.
- The `SANDBOX_GLOBALS` updates in `node-sdk` and `web/` are consistent with the new snapshot.
- New tests cover: snapshot drift, prompt drift, chat exclusion correctness, manifest note propagation.

**CI:** 25/26 checks green; Playwright visual snapshots still running — merge when green.

---

georgi:
Heads up before merging: that review covers `5f7a9b8`, but `38a5a71` landed after it and includes a runtime behaviour change, not just tests. Worth a second look at the smaller commit.

It fixes three defects a review pass found in the first one:

1. **`process.env` was not empty.** The note added in `5f7a9b8` told every model that `process` carries an empty `env`; it actually held `{"NODE_DEBUG":"true"}`, because `@sebastianwessel/quickjs` defaults the option and `runInSandbox` never passed one. Fixed by passing `env: {}` rather than rewording — `NODE_DEBUG` can flip debug paths in the library's `node:util` polyfill, and guest code has no business reading host-shaped environment variables. **This is the behaviour change**: `process.env` is now `{}` in the guest.

2. **`env` became un-inferable as a Code-node input.** Adding every real guest global to both mirror sets reserved `env` in `web/`'s input inference, which asks a different question than the validator. Dynamic inputs are exposed after the library's stubs and shadow them, so a Code node with an input named `env` works — reserving the name would have dropped that handle on re-inference and left the code silently reading `{}`. That is the same "the pinning test prevents the fix" trap this PR exists to remove, one layer up, so the drift test grew an `INTENTIONAL_OMISSIONS` escape hatch and `env` is inferable again.

3. **Manifest notes were filtered by substring.** The CodeAct prompt excluded the Code-node notes by matching their prose from another file, so rewording one would have leaked "Return an object whose keys are the node's outputs" into a prompt whose steps complete through `finish()`. Notes are now `{text, audience}` and the filter keys on the field.

CI on `38a5a71` is currently 21 green / 0 failed, with `examples`, `test-packages` and Playwright visual snapshots still running.

---

PR #4813, opened 2026-08-10
feat(seo): consolidate the comparison pages and add the node-based-AI entity page

georgi:
Executes the §0.10 roadmap added in #4808. Three shipped items, plus two corrections to the audit itself.

## Consolidation — `/vs/*` folded into `/alternatives/*`

Both routes ran `generateStaticParams` over the same `competitors` array, so 13 records produced 26 URLs competing for one query set. `/alternatives` won 11 of 12 head-to-head pairs (4,817 impressions to 1,117), so it becomes canonical and `/vs/:slug` redirects permanently.

The copy `/vs` owned moved onto the alternatives template rather than being deleted: the at-a-glance cards (honoring `competitorBulletTone`, which my first pass got wrong by hardcoding the minus icon) and the explainer, under "NodeTool vs X" H2s so the head-to-head queries keep on-page support. That also carries the **n8n** and **jan** angles across — the two pages that out-ranked their twins and the reason the audit said to port before redirecting.

Five fields that only described the retired page are gone from the type and all 13 records (`vsTitle`, `vsDescription`, `vsOgTitle`, `vsOgDescription`, `heroHeading`); `heroParagraph` now leads the at-a-glance section. `ComparisonMesh`'s `basePath` prop went with them — one caller, one value left.

## New page — `/node-based-ai`

The "node based ai" / "ai node editor" cluster is 120 query variants and 11,123 impressions — two thirds of all non-branded impressions — stuck at position 7.5 because every variant lands on the homepage, which has to serve every intent at once.

The page matches the query in its title, H1, and first paragraph; disambiguates the three intents the cluster mixes (visual graph vs. Node.js vs. ComfyUI) instead of assuming one; and gives each section a specific variant to target rather than restating the head term. Linked from the homepage hero (on the words "node-based"), the footer, and every `/alternatives/*` page via `ComparisonMesh`.

## Index hygiene

- **27 internal planning docs** (`plans/`, `sdk/`, PRDs, designs, audits, and the SEO docs themselves) get `noindex` + `sitemap: false` from `_config.yml` defaults, with the robots meta emitted in `_layouts/default.html`. They keep their URLs deliberately: published pages and `CLAUDE.md` link to them, so `exclude:` would fail Docs CI's html-proofer step. The config comment explains when to reach for which, and I dropped 9 entries that were dead config because those files are already under `exclude:`.
- **The six `docs/use-cases/*` mirrors** now carry `canonical_url` pointing at their marketing equivalents. This resolves the cross-domain duplicate without unpublishing content the docs use-case index links.
- `/apps` H1 now says "AI mini apps" — the `<title>` already did.

## Two corrections to the audit

Both came from reading the code instead of trusting the export, and both are recorded in the doc:

- **Finding 4** claimed a live `/apps` ↔ `/templates` duplication across 20 slugs. It was already fixed in `next.config.mjs` by the examples-revamp — the export was measuring URLs that had stopped resolving. Only the movie-trailer split was real, and that is what this PR fixes. The general lesson is now written into the finding: GSC lags the deploy, so check the route table before acting on a duplication finding.
- **Finding 5** called `/providers/kie` a spec dump needing a rewrite. It is already a full landing page with a differentiated title, tagline, blurb, FAQ, and generated catalog. "kie ai" is simply **navigational** — searchers want kie.ai — so 0.24% CTR at position 7 is the expected outcome, not a defect. The provider-page rewrite is struck from the roadmap; the winnable slice is the capability tail ("kie suno", "kling 3.0 local model"), which belongs on the model pages.

## Testing

- `tsc --noEmit` clean; `next lint` clean for every touched file (only pre-existing `` warnings elsewhere).
- Production build green (668 → routes regenerated, `/vs` output gone).
- **All 368 Playwright e2e tests pass** at CI's `workers: 1`. The smoke suite derives its routes from the page registry, so the new page is covered and the removed ones drop out with no test edit. One note for reviewers: a CLS test on `/` flakes under high parallelism but passes at CI's worker count and in isolation.
- Verified against a running production server: `/vs/comfyui` → **308** with `location: /alternatives/comfyui`, `/alternatives/comfyui` still 200, `/node-based-ai` 200, sitemap contains `/node-based-ai` and zero `/vs/` entries, and the ported sections render on `/alternatives/n8n`.
- `_config.yml` validated with Ruby's own YAML parser (it uses anchors) — Docs CI builds Jekyll on PRs touching `docs/**`, so the config change is covered there too.

## Open items from the audit, not in this PR

The rescue work on the nine page-two pages (finding 11), the billing-block-generator page, the "chain multiple LLMs in one workspace" page, and per-model cost text (findings 5 and 9). Roadmap items 4–6 in the doc.

https://claude.ai/code/session_01MciigRxNhjazbfockgHvZW

---
_Generated by [Claude Code](https://claude.ai/code/session_01MciigRxNhjazbfockgHvZW)_


georgi:
## CI status: two red checks, both failing on `main` too

This PR touches only `docs/**` and `marketing/**` — zero files under `packages/` or `web/`. Both failures reproduce on the base branch, so I'm not fixing them here; details below so whoever owns them can pick them up.

### 1. `Workflow Runner Browser E2E` — red on `main`

All 10 tests in `packages/workflow-runner` fail identically: the harness page never sets `window.workflowRunnerReady === true`, so every test times out at `page.waitForFunction` after 15s. That's a harness-boot failure, not 10 separate test failures.

Same job, same step ("Run workflow-runner browser E2E tests"), failed on `main` at `500defde` — [run 31365855431](https://github.com/nodetool-ai/nodetool/actions/runs/31365855431), which started at 07:26, ahead of this PR's 07:58 run.

### 2. `Quality Gate / docker` — red on `main`, and the root cause is nameable

The image build fails at `Dockerfile:75` (`cd web && npm run build`) with two missing-export errors from `vite-node-stubs`:

```
[MISSING_EXPORT] "realpathSync" is not exported by "vite-node-stubs/fs-stub.js"
[MISSING_EXPORT] "relative"     is not exported by "vite-node-stubs/path-stub.js"
  ../packages/node-sdk/dist/sandbox-pack-discovery.js:2-3
```

`sandbox-pack-discovery.js` imports `realpathSync` from `node:fs` and `relative` from `node:path`, and the web build's Node stubs don't export either. This looks like fallout from the sandbox pack-discovery work that landed just before this branch's base (`ba3fc55 fix(sandbox): harden Electron pack discovery`, `ff1ca49 Design: JS/WASM package system for the QuickJS sandbox`). The fix is presumably adding those two names to the stubs, or keeping `sandbox-pack-discovery` out of the web bundle — either way it belongs in a sandbox/web PR, not an SEO one.

The `docker` leg also failed on `main` at `500defde` at the same `Build image` step.

### What is green, including everything this PR could plausibly break

| Check | Result |
|---|---|
| **Docs CI** (`build-and-check`) — Jekyll build, html-proofer links/images, internal-docs privacy guard | ✅ |
| **Marketing** — Typecheck, lint & build | ✅ |
| **Marketing** — Playwright smoke tests (368 tests) | ✅ |
| **Marketing** — Lighthouse budget | ✅ |
| Quality Gate — lint, build, deps | ✅ |
| Workflow Integration Tests | ✅ |
| Markdown link check | ✅ |
| CodeQL — actions, python, javascript-typescript | ✅ |

Docs CI passing is the meaningful signal for the `_config.yml` noindex defaults and the `canonical_url` front matter: it builds Jekyll, checks every internal link resolves, and runs the guard that no internal doc leaks into the published site.

I'll pick these back up if either check goes green on `main` and this branch needs a rebase to inherit it.

---

georgi (commented):
Large PR (~1,594 lines across many marketing files). The Quality Gate checks (typecheck, lint, tests, build, etc.) all pass. The `docker` and `browser E2E` failures are pre-existing flaky infrastructure issues.

This consolidates comparison pages and adds a node-based AI entity page in the marketing site. Due to the size, a full code review was not feasible in this pass. The marketing-specific CI (typecheck, lint, build, Playwright smoke tests, Lighthouse) all pass, which is a good signal.

No cross-PR conflicts detected with other open PRs.

---

georgi:
Thanks — two notes, one of which matters beyond this PR.

### The `docker` failure is not flaky, and it will block the post-merge deploy

It reproduces deterministically with a named cause: the web build inside the image fails on two missing exports from `vite-node-stubs`.

```
[MISSING_EXPORT] "realpathSync" is not exported by "vite-node-stubs/fs-stub.js"
[MISSING_EXPORT] "relative"     is not exported by "vite-node-stubs/path-stub.js"
  ../packages/node-sdk/dist/sandbox-pack-discovery.js:2-3
```

Same failure, same step, on `main` at `500defde`. This is worth separating from the E2E one because of what the leg is for: per CLAUDE.md, `docker.yml` runs **only after a merge** and then auto-deploys the image to Fly, so the Quality Gate carries a `docker` leg to catch exactly this before it reaches prod. It is doing its job right now. Filed as intelligence rather than a fix here — it belongs in a sandbox/web PR, since `sandbox-pack-discovery` arrived in `ba3fc55` / `ff1ca49`, just before this branch's base.

The `Workflow Runner Browser E2E` failure is the one I'd call infrastructure: all 10 tests fail identically because the harness never sets `window.workflowRunnerReady`, so it's one boot failure, not ten flaky tests. Also red on `main`.

### Review guide, since size was the blocker

The diff is 663 insertions / 412 deletions across 24 files, and most of it is mechanical. The parts that actually need eyes:

| What | Where | Lines |
|---|---|---|
| The new entity page (all new logic) | `marketing/src/app/node-based-ai/page.tsx` | +325 |
| Two sections ported off the retired `/vs` template | `marketing/src/app/alternatives/[slug]/page.tsx` | +~60 |
| The redirect | `marketing/next.config.mjs` | +11 |
| The noindex set | `docs/_config.yml` | +80 (data, one entry per doc) |

Everything else is deletion or its consequence: `competitorEntries.ts` is −122/+30 (five dead fields × 13 records), and `src/app/vs/` is ~−248 (the deleted route). Reviewing the four rows above covers the whole behavioral change.

The claim most worth checking is the one I'd want checked: that no copy was lost when `/vs` folded in. `heroParagraph`, `competitorBullets`/`nodetoolBullets` with `competitorBulletTone`, and `explainerHeading`/`explainerParagraph` all still render — verified on `/alternatives/n8n` against a production server. The five deleted fields were the retired page's own title/OG metadata and its H1.

---

georgi:
Both inherited failures are now fixed here in `8c78a53`, at your request. They turned out to be **one** root cause, not two — which also corrects my earlier read that the E2E one was separate "infrastructure".

### The single cause

`packages/node-sdk/src/index.ts` does `export * from "./sandbox-pack-discovery.js"`, so every bundle importing node-sdk pulls in that module's `node:fs` / `node:path` imports. The browser stubs never had `realpathSync`, `lstatSync`, or `relative`.

- **docker** — the web build failed at bundle time on the missing named exports.
- **Workflow Runner Browser E2E** — the harness imports `@nodetool-ai/core-nodes` → node-sdk, so the same missing exports stopped `browser-entry.ts` from evaluating. `window.workflowRunnerReady = true` is the last statement in that file, so the flag was never set and all 10 tests timed out on it. One boot failure wearing ten hats.

### The fix

Added the names to both stub sets — `web/vite-node-stubs/` and `packages/workflow-runner/e2e/stubs/`. I found them by scanning every `node:fs` / `node:path` named import under `packages/` and `web/` rather than chasing what the bundler reported, because it stops at the first missing name per import statement — which is exactly why `lstatSync` was hiding behind `realpathSync` and would have failed the build a second time.

Two additions deliberately aren't throwers:

- `constants` is data — `accessSync(p, constants.R_OK)` reads it *before* it would ever call, so a throwing stub breaks at the wrong moment. Real POSIX values.
- `relative` returns a string callers use as a module id. Implemented properly and checked against `node:path.posix.relative` across sibling, ancestor, identical, and disjoint paths — all six cases match.

### Verified locally

After `npm run build:packages`:

- `vite build` in `web/` → **green, zero MISSING_EXPORT** (this is the step the docker leg fails on).
- `playwright test --config e2e/playwright.config.ts` in `packages/workflow-runner` → **10/10 passing**, WebGPU shader test included.

### One thing I did not fix

`web/vite-node-stubs/` and `packages/workflow-runner/e2e/stubs/` are two hand-maintained copies of the same stubs — which is precisely why one missing export took down two surfaces and why `main` went red without anyone noticing. Worth either unifying them or adding a test asserting the stubs cover every `node:` name the workspace imports. I left that out to keep this PR reviewable, but happy to do it here or in a follow-up — your call.

---

PR #4846, opened 2026-08-11
Tell the node-editor agent which workflow is open

georgi:
## The gap

Chatting from the node editor left the agent with no workflow id.

The turn already carried one — `GlobalChatStore.sendMessage` binds the open workflow onto `workflow_id` — but nothing rendered it into the system prompt. Only `ui_context` is rendered (`formatUiContext`), and the canvas composer never sent one. Meanwhile that same prompt tells the agent:

> Every `ui_*` tool requires the id of the document it should act on; pass one of the ids above.

…with no ids above. The tools happened to work — both the renderer bridge (`workflow_id ?? state.currentWorkflowId`) and the server fallback (`params.workflow_id ?? context.workflowId`) resolve the graph implicitly — but the agent was guessing at an id it had been told it needed, and could not name the workflow when the user asked.

## The change

**Frontend** — `CanvasMediaComposer` now sends `ui_context`, the same block `ChatView` already sends, built from the open graph rather than the tab store: the workflow's id and name as `focused`, plus the selected node ids as `selection.node_ids`. Nothing currently sends `node_ids`, and `formatUiContext` already renders selection. The graph is read imperatively via `useNodeStoreRef` at send time — subscribing to the selection would re-render the composer on every canvas click.

**Backend** — `buildChatAgentSystemPrompt` takes the turn's `workflowId` and names it when `ui_context` does not. This covers every client that binds a workflow without describing its UI, not just this composer. It folds into the existing "What the user is looking at" section instead of opening a second one, and is skipped entirely when `ui_context` already lists that workflow — that block says it better.

No protocol change: `ui_context` and `workflow_id` both already exist on `Message`. No change to the `ui_*` tool schemas: `workflow_id` is already optional with a correct fallback.

The stale comment about `workflow_id` routing into `handleWorkflowMessage` is corrected — the routing signal is `workflow_target`, not a bare `workflow_id`.

## Tests

- `web/…/CanvasMediaComposer.test.tsx` — asserts the outgoing message names the open workflow and its selected nodes. The two existing assertions (media generation must not set `workflow_id`) still hold.
- `packages/websocket/tests/chat-system-prompt-workflow-context.test.ts` — the backstop names a bound workflow, stays silent with none, defers to `ui_context` when it already lists the workflow, and never emits two headings.

## Verification

`tsc --noEmit` over `web` (which compiles the packages), eslint on both touched files, and the affected suites: the three `CanvasMediaComposer` cases, the four new prompt cases, and `chat-agent-parity` / `resident-tools` / `workflow-message-parity` (31 passing).

One pre-existing failure in this environment is unrelated and untouched by the diff: `packages/websocket/src/sandbox-catalog.ts` cannot resolve `@nodetool-ai/sandbox-compiler`, which is not linked into `node_modules` here.

https://claude.ai/code/session_018rix9aPJpDP5BZt4xRxDC1

---
_Generated by [Claude Code](https://claude.ai/code/session_018rix9aPJpDP5BZt4xRxDC1)_

### DIFF
--- a/packages/websocket/src/unified-websocket-runner.ts
+++ b/packages/websocket/src/unified-websocket-runner.ts
@@ -1469,21 +1469,51 @@ export const RESIDENT_TOOL_NAMES: ReadonlySet<string> = new Set([
  * `system_prompt` on the chat message — it is layered after the base prompt as
  * a context-specific addendum, never a replacement.
  */
-function buildChatAgentSystemPrompt(
+export function buildChatAgentSystemPrompt(
   mode: PermissionMode,
   extraSystemPrompt?: string | null,
-  uiContext?: UiContext | null
+  uiContext?: UiContext | null,
+  workflowId?: string | null
 ): string {
   const extra =
     typeof extraSystemPrompt === "string" && extraSystemPrompt.trim()
       ? `\n\n${extraSystemPrompt.trim()}\n`
       : "";
+  const uiBlock = formatUiContext(uiContext);
   return (
     CHAT_AGENT_SYSTEM_PROMPT +
     PERMISSION_MODE_PROMPTS[mode] +
-    formatUiContext(uiContext) +
+    uiBlock +
+    formatBoundWorkflow(uiContext, workflowId, uiBlock !== "") +
     extra
   );
+}
+
+/**
+ * Backstop for clients that bind a workflow to the turn (`workflow_id`) without
+ * naming it in `ui_context` — the canvas composer and every headless client.
+ * The graph `ui_*` tools take that id, so a turn carrying one and saying
+ * nothing about it leaves the agent guessing. Skipped when `ui_context` already
+ * names the workflow: that block says it better.
+ */
+function formatBoundWorkflow(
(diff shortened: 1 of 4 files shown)


georgi:
Code reviewed — looks correct.

**Backend** (`unified-websocket-runner.ts`): `formatBoundWorkflow` is a clean backstop that names the workflow only when `ui_context` doesn't already cover it, and folds into an existing "What the user is looking at" section rather than duplicating the heading. Exporting `buildChatAgentSystemPrompt` for testability is fine.

**Frontend** (`CanvasMediaComposer.tsx`): Reading `nodeStore` via ref at send time (not subscription time) avoids re-renders on every canvas click. `buildUiContext` populates focused workflow + selection correctly. The `workflow_id` field is intentionally omitted from the outgoing message to avoid routing to `handleWorkflowMessage`.

**Tests**: Both server-side (4 cases covering all branches of the new backstop) and client-side (verifying the outgoing message carries the correct `ui_context`) are thorough.

21/23 CI checks green — waiting on typecheck and Playwright visual snapshots. Ready to merge once CI completes.

---

PR #4884, opened 2026-08-13
Model Discovery and Media 

georgi:
Fix the agent's path from "pick a model" to "show me the image"

One chat session exposed a chain of defects, each of which broke a step an agent has to take to generate media and build a workflow around it. This branch fixes every link, plus the audit findings that followed.

The session that started it

generate 2 images with flux schnell and combine them took 7 actions and never finished. find_model couldn't search by name. The model it did pick came from a provider whose SDK wasn't installed. Reading the generated bytes back had no documented path, so the agent spent four actions probing the API. A later session hit the sequel: the workflow it built returned a 401 from a provider that had just worked in the same turn.

Model discovery

- find_model searches by name (c05165f12). query is free text over model id and name, punctuation-insensitive, so "flux schnell" finds black-forest-labs/FLUX.1-schnell. A name typed into task is read as a name search rather than filtering everything out, and a miss says so instead of ranking something unrelated first.
- Only rankable models are ranked (f1b145403). huggingface was counted as a locodel took a "downloaded" bonus and outranked the fal_ai copy that could actuallyinstalled. Reading the generated bytes back had no documented path, so the agent spent four actions probing the API. A later session hit the sequel: the workflow it built returned a 401
from a provider that had just worked in the same turn.

Model discovery

- find_model searches by name (c05165f12). query is free text over model id andtive, so "flux schnell" finds black-forest-labs/FLUX.1-schnell. A name typed into task is read as a name search rather than filtering everything out, and a miss says so instead of ranking something unrelated first.
- Only rankable models are ranked (f1b145403). huggingface was counted as a locodel took a "downloaded" bonus and outranked the fal_ai copy that could actuallyrun. BaseProvider.unavailableReason() now answers what a credential check cannot — an optional SDK nobody installed — and such a provider is skipped and named in the result note, at lookup time rather than on the paid call.
- Discovery is a direct tool call (a89abb4a5). find_model, list_models, list_provider_models, search_nodes, get_node_info, list_nodes are offered next to execute_code instead of only
behind it. DIRECT_TOOL_NAMES is the union of these and the core set; the two stubstitution can never reach a discovery tool. The belt keeps every tool, so
nodetool.models.pick still composes inside an action.

Reading media back

- media.* reads only what it may (3b7edd7ab) — paths resolve through the same c* uses.
- read_media_bytes and read_asset accept what a generation returns (f422783ed):d, a storage key, a data or http URL.
- nodetool.media.bytes(ref) makes it reachable (0bccd849d). The chat variant strips the media bridge, and the assets doc advertised only read(name) — a name no generation result carries. The new method takes the result object itself and never the host uri, which is a filesystem path the guest may not read. The chat contract now says where the stripped bridges went, so generate → image.composite → save is one action.

The 401

- Service-layer runs carry their own secrets (7da3a577a). buildWorkspaceExecutionContext set no resolver, so getSecret answered null for every key and providers saw only the environment — a key in the secret DB was invisible. This hit run_workflow, debug_workflow and POST /api/workflows/:id/run; the WebSocket paths wire their own and were unaffected, which is why the
same graph ran from the editor and 401'd from an agent. ExecutionSession's fallsame way now.

Persistence stopped depending on which entrance built the run

677411ab5, from an audit of every ProcessingContext construction site:

- Six hosts assembled model interfaces by hand and three forgot createAsset. setDefaultModelInterfaces lets a host install its persistence once; a context that wires its own still wins. The server installs serverModelInterfaces() at startup, the CLI installs an equivalent bundle from both entry points — replacing a hand-rolled createAsset that wrote no thumbnail and
left parent_id null, orphaning every CLI asset from folder listings.
- typeof context.createAsset === "function" is true on every context, wired or not. Three guards therefore could not fail, and the fallbacks behind them were dead code; SaveImage threw where it should have written a file. hasModelInterface is what they meant.
- debug_app built its script runner with no secret resolver, while the server s both pass one.

CLI

- The chat belt carries what the prompt documents (c13603e98). find_model was never in enabledTools, so nodetool.models.pick() threw "tool not in this toolbelt" on every turn. ALWAYS_ENABLED_TOOLS covers documents, discovery, generation and the read-back path, and reaches settings files written before they existed. Generation is no longer gated on OPENAI_API_KEY — these route by the model they're handed.
- Code actions render as title + program (ed786d022) instead of execute_code({.m as JSON arguments.

MCP

- /mcp teaches the guest contract (5c192598f). MCP has no system prompt, so thedbox summary ride in the execute_code description and on nodetool://sandbox.

Testing

New coverage for each fix, and each new test was checked against the old code: ior, media.bytes (4 cases in the real QuickJS sandbox), direct-tool membership,the CLI belt, secret resolution in a run context (3 cases), the interface default (3 cases), and SaveImage's two paths. Six agents test files had duck-typed fake contexts that define createAsset without hasModelInterface; they now implement the check rather than the guard being softened to accommodate them.

npm run typecheck, npm run lint and npm run test pass. Pre-existing and unrelatt.ts type errors and the vendored electron/.node-runtime/ lint output.

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -195,9 +195,10 @@ tool**, not a flat catalog.
 `execute_code` and `view_image`, which is direct because pixels cannot ride a
 sandbox action's JSON observation envelope. Everything else on that belt is
 reachable inside an action as `tools.<name>()`, through the `nodetool.*` object
-model, or found with `await searchTools("query")`. MCP has no system prompt, so
-the action contract and tool catalog ride in the `execute_code` description, and
-the machine-readable form is the `nodetool://capabilities` resource.
+model, or found with `await nodetool.searchTools("query")`. MCP has no system
+prompt, so the guest contract leads the `execute_code` description and is also
+the server `instructions` string. The machine-readable form is
+`nodetool://capabilities` (tools) plus `nodetool://sandbox` (guest surface).
 
 It used to register all ~95 bridged tools flat, which made a scoped MCP session
 NodeTool's largest surface anywhere — 120 tools and ~27k tokens of schema before
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -481,11 +481,12 @@ nodetool mcp status
 nodetool mcp uninstall
 ```
 
-The NodeTool server must be running (`nodetool serve`) for MCP to work.
+HTTP MCP (`/mcp`) needs the API server (`nodetool serve`). `nodetool mcp serve`
+is stdio and does not.
 
 ### What the MCP server exposes
 
-Exactly two tools land on `/mcp`:
+Exactly two tools land on `/mcp` and on `nodetool mcp serve`:
 
 - **`execute_code`** — the CodeAct action tool, built by the same
   `createChatCodeActSession` the in-app chat agent runs on, so the two surfaces cannot drift.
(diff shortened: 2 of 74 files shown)


georgi (commented):
Reviewed the diff. The changes look solid — model name search in `find_model`, `read_media_bytes` capability, MCP guest contract, discovery tools as direct tools, improved asset URI resolution, and the HuggingFace local-provider fix.

**CI is still running** — several Quality Gate jobs (`build`, `lint`, `deps`, `docker`) and the browser/integration tests are in progress.

**Merge conflicts expected**: five other PRs were just merged to main (#4879, #4880, #4881, #4882, #4883). PRs #4882 and #4883 edited `AGENTS.md`, `docs/AGENTS.md`, `packages/agents/AGENTS.md`, and `docs/cli.md` — the same files this PR touches. You will need to merge main into this branch and resolve conflicts before merging.

Will review again once CI completes and conflicts are resolved.

---

PR #4904, opened 2026-08-15
ci: add ten scheduled maintenance routines

georgi:
Completes the routine set the repo already started with `dead-code-cleanup.yaml`, `type-safety.yaml`, `performance-optimization.yaml` and friends. Ten new workflows, each following the house pattern: a deterministic scan writes an evidence file and sets an env flag, the agent step runs only when the scan found something, and a post-change step re-runs the quality checks against a pre-change baseline so a routine can only fail the run by *introducing* a failure.

All ten run daily, spread across 02:00–08:30 UTC at half-hour intervals so no two start together and none lands on the 03:00 slot the existing routines share.

| Workflow | What it does | Daily at |
|---|---|---|
| `flaky-test-fixer.yaml` | CI re-run history plus randomized repeat runs | 02:00 |
| `abstraction-police.yaml` | The five `check:*` scripts plus the import greps no script covers | 04:30 |
| `crash-fuzzer.yaml` | Mutates shipped examples and CLI fixtures, feeds them to `validate` / `jsscript validate` / `node run`, reports stack traces and hangs | 05:00 |
| `internal-only-shipper.yaml` | Finds dev/internal-gated surfaces, forces a ship-or-delete verdict on one | 05:30 |
| `logic-simplifier.yaml` | Ranks files by branch density and nesting; rewrites one, behavior-preserving | 06:00 |
| `logic-bugfixer.yaml` | Models one decidable function against an independent implementation, enumerates its inputs | 06:30 |
| `duplicate-unifier.yaml` | Sliding-window hash over normalized lines; merges one duplicate group | 07:00 |
| `useless-test-pruner.yaml` | Nominates tests that assert nothing, requires a mutation proof before touching one | 07:30 |
| `shipped-feature-inliner.yaml` | Boolean env gates and pinned constants, ranked by whether the repo ever sets them | 08:00 |
| `abstraction-improver.yaml` | Single-implementer interfaces, forwarding wrappers, re-export-only barrels | 08:30 |

Each opens at most one PR per run and skips when a PR from the same routine is already open, so the daily cadence raises the rate the backlog is worked through, not the number of PRs in flight.

## Design notes

**The scans are grounded in this repo, not generic.** `abstraction-police` runs `check:deps`, `check:circular`, `check:keys`, `check:coupled-deps` and `check:execution-boundary`, and its prompt states the package dependency order and the `ui_primitives` rule from AGENTS.md. `crash-fuzzer` seeds from the shipped examples and `packages/cli/tests/fixtures/js-script-*.json`, and derives its node types from those graphs so it only fuzzes types the registry actually has. `flaky-test-fixer` uses `jest --randomize` and `vitest --sequence.shuffle`, which target the order-dependence class of flake directly.

**Each routine names what it must not do.** Fuzzer fixes may not be a catch-all wrapper. Deflaking may not use retries, skips or widened timeouts. Flag inlining may not touch a kill switch or a safety gate. Layering fixes may not widen the grandfathered allowlist in `check-execution-boundary.mjs` — that list may only shrink. Test pruning may not delete a test whose mutation was never run.

**`useless-test-pruner` requires proof, not a heuristic.** The scan nominates; the agent must break the code under test and watch the test pass anyway before deleting it. A dedicated step then fails the run if a mutation was left behind in a non-test file.

## Verification

Every scan script was extracted and run against the working tree before landing, and each returns real findings — e.g. `electron/src/__tests__/ipc.test.ts:553` and `window.test.ts:219` are tests with zero assertions and a comment saying so.

Two bugs surfaced by running them, both of which would have shipped silently:

- `git ls-files 'dir/**/*.ts'` does not match files sitting directly in `dir` — it was skipping 42 of 119 electron sources, and the largest backend files entirely. Switched to `dir/*.ts`, which globs recursively under git's default pathspec (4395 → 4977 files scanned).
- Reading the config corpus through `execFileSync` overflowed the 1 MB default buffer with `ENOBUFS`, which crashed `shipped-feature-inliner`'s scan outright. Raised `maxBuffer` there and in the other four scans.

The first version of `shipped-feature-inliner` also found zero flags, because this repo gates on env vars rather than `const ENABLE_X = true` — a workflow that could never fire. It was rewritten around boolean env gates and now reports 5 of them with whether anything in the repo sets each one.

All 51 workflow files parse as YAML with no control characters, every `if: env.X` in the new workflows refers to a variable an earlier step sets, and no two schedules collide.

## Docs

`.github/workflows/README.md` gains the ten rows, the maintenance section is re-sorted alphabetically, and the summary line is corrected to the counts the table actually holds (51 files, 10 in rings, 41 maintenance — it read 43/12/31).

https://claude.ai/code/session_013H1r8Hmer6rdHjwdkfgzvM

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -43,22 +43,30 @@ see below.
 | `user-journeys.yml` | `journeys`: nightly Playwright journey suite (build a graph and run it, chat, mini app, library). `reliability-ring1` (on push to `main`, schedule, dispatch): full `reliability/journeys/*` suite on kernel+ws-server with `--diff`, plus one packaged-backend journey — gates `fly-deploy.yml` | 1 | `journeys`: advisory until 2026-08-15, then required. `reliability-ring1`: required |
 | `release.yaml` | Cross-platform signed release artifacts, packed-tree smoke, a packed-backend reliability journey per OS, updater assets | 2 | Required |
 | `example-smoke-debug.yml` | Nightly + manual real-provider smoke via `nodetool debug` | 2 | Nightly (spend-capped), also dispatch-only |
+| `abstraction-improver.yaml` | Scheduled agent flattens single-implementation interfaces, forwarding wrappers, re-export-only barrels | none/maintenance | Advisory (`continue-on-error`) |
+| `abstraction-police.yaml` | Scheduled agent fixes layering violations found by `check:*` plus the import greps no script covers | none/maintenance | Advisory (`continue-on-error`) |
 | `app-build-eval.yml` | Nightly `app-build` eval suite; reports the one-shot rate, gates nothing | none/maintenance | Advisory (report only) |
 | `aur-publish.yml` | Publish the AUR package on a GitHub release | none/maintenance | Required for its own job |
 | `claude-code-review.yml` | Claude reviews new/updated PRs | none/maintenance | Advisory |
 | `claude.yml` | Claude responds to `@claude` mentions and comments | none/maintenance | Advisory |
 | `copilot-setup-steps.yml` | Environment setup for the Copilot coding agent | none/maintenance | n/a (setup only) |
+| `crash-fuzzer.yaml` | Weekly fuzz of `validate` / `jsscript validate` / `node run` over mutated documents; agent root-causes each crash or hang | none/maintenance | Advisory (`continue-on-error`) |
 | `dead-code-cleanup.yaml` | Scheduled agent removes unused exports/imports/code | none/maintenance | Advisory (`continue-on-error`) |
 | `dependency-cleanup.yaml` | Scheduled agent prunes unused deps, aligns/updates versions | none/maintenance | Advisory (`continue-on-error`) |
 | `docs-ci.yml` | Docs site build + internal link/image check on PR | none/maintenance | Required for `docs/**` |
 | `docs-completeness.yaml` | Daily agent documents undocumented CLI commands, routes, env vars, packages | none/maintenance | Advisory |
 | `docs-correctness.yaml` | Daily agent checks docs claims against the code and fixes stale ones | none/maintenance | Advisory |
 | `docs-lint.yml` | Markdown lint on push/PR | none/maintenance | Required for `**/*.md` |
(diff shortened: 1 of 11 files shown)


georgi:
**Workflow Runner Browser E2E** is red on this PR, and it is red on `main` for the same reason. Not fixing it here — this PR changes only `.github/workflows/*` and cannot reach `packages/workflow-runner`.

**On this PR** ([job](https://github.com/nodetool-ai/nodetool/actions/runs/31880381392/job/95002486374)): all 27 specs fail identically, timing out at `page.waitForFunction(() => window.workflowRunnerReady === true)`. The harness page never boots. The cause is upstream of the tests, in the dev server:

```
[WebServer] ✘ [ERROR] Failed to resolve entry for package "https".
  [plugin vite:dep-pre-bundle]
  ../../../node_modules/@opentelemetry/otlp-grpc-exporter-base/node_modules/
    @opentelemetry/otlp-exporter-base/build/esm/transport/http-exporter-transport.js:34:66
[vite] (client) error while updating dependencies: Build failed with 2 errors
```

Vite's dep optimizer fails while pre-bundling a nested `@opentelemetry/otlp-exporter-base` that imports the Node builtins `https` and `http`. The optimizer never completes, the browser gets `504 (Outdated Optimize Dep)` on its module requests, and every spec times out waiting for the ready flag.

**On `main`** ([job](https://github.com/nodetool-ai/nodetool/actions/runs/31870335524/job/94977905829), commit `68c30fbf`, which predates this branch): the same job fails with the same 27 specs and the same `workflowRunnerReady` timeout at the same spec lines. `Workflow Integration Tests` and `Quality Gate / test-packages` are also failing on that commit, and the last four completed `test.yml` runs on `main` all concluded `failure`.

A real fix belongs in `packages/workflow-runner`'s Vite config — the browser bundle should not be pre-bundling a gRPC/Node OTel transport at all, so the fix is either excluding it from `optimizeDeps` or keeping the Node-only OTel exporter out of the browser entry's import graph. That is outside this PR's scope.

I will re-check when the base branch goes green and update this PR if anything changes.

---

PR #4905, opened 2026-08-15
fix: get the Quality Gate green — browser E2E, a leaked process, a 30-minute deadline, and four smaller defects

georgi:
Started as one job: **Workflow Runner Browser E2E**, red on `main`, all 27 specs timing out because Vite's dependency pre-bundle died. Fixing it uncovered a chain — each defect was hidden behind the one in front of it, and each fix let CI get one package further. Ten in total; two are real product bugs, not test problems.

All of them were red on `main`. None were introduced here.

## Browser E2E: 0/27 → 27/27

**1. A bare Node builtin in the dep optimizer.** `runtime/telemetry.ts` dynamically imports `@opentelemetry/sdk-node`, so Vite pre-bundled the OTel exporter chain — including a module that reaches its transport with a bare `import('https')`. A bare builtin name is not a builtin to a bundler: it resolves against `node_modules` first, and `pptxgenjs` depends on an empty squatter package literally named `https` (a `package.json`, `main: index.js`, no `index.js`). The harness stubbed only the `node:` form. Bare `http` survived by luck — nothing shadows that name — which is why exactly two errors appeared, both for `https`.

The stub table is now keyed by bare name with both specifier forms derived from it, so the esbuild plugin, the `resolveId` plugin and `resolve.alias` cannot drift. The browserify-shim names stay on the `node:` form deliberately: `buffer-stub.js` imports the real npm `buffer`, and `stream-stub.js` pulls `readable-stream`, which reaches npm `events`/`string_decoder`/`buffer` by bare name — stubbing those would make the stubs import themselves.

**2. `video.addAudio` hardcoded AAC.** Chrome ships no AAC *encoder* on Linux (Media Foundation on Windows, AudioToolbox on macOS), so Mediabunny dropped the audio track. Probed in the same headless Chromium: `avc1: true, opus: true, mp4a.40.2: false`. Now asks the platform via `getFirstEncodableAudioCodec(["aac", "opus"])`, probed with the source track's channels and sample rate — AAC still wins where it exists, so Node and macOS are unchanged.

**3. The MP4 fixture was truncated.** `mdat` declared 707 payload bytes and carried 690, leaving 5 of 7 packets readable. Not codec-related — it reproduced identically in Node while real MP4s converted fine. Re-encoded with Mediabunny and documented.

**4. `mediabunny` was discovered mid-run.** Nothing imports it statically, so Vite pre-bundled it when the media test first called `video.*` and reloaded the page out from under `page.evaluate`. Cold-cache only — invisible locally, guaranteed in CI. Fixed with `optimizeDeps.include`.

That fourth one is why this took several passes: `Execution context was destroyed` is what *all* of these look like from inside the suite, and it masked the real error every time. Running the spec alone was needed to see each one.

## Then the rest of the Quality Gate

**5. A sandbox worker that never released the process.** `app-build` was cancelled at its 45-minute job timeout on every run — while finishing its actual work in 0.8s. Its only ref'd handle was a `MessagePort`: the factory called `worker.unref()` and *then* attached the `message` listener, which starts the worker's public port and re-refs it. Isolated: unref-then-listen holds the process open, listen-then-unref exits. The ref now follows the lease, so an in-flight run still can't be cut short. **45 min → 1m47 in CI.**

**6. A 30-minute floor under every sandbox run.** The worker host sized its backstop `timeoutMs + suspendAllowanceMs + margin` unconditionally, and that allowance defaults to 30 minutes. A run with no clock cannot suspend at all, so a wedged worker outlived its own `timeoutMs` by half an hour. The contract was already documented — *"Only consulted when `clock` is set"* — and both other consumers already gate on `hasClock`; only this backstop ignored it. On the case its test describes, with `timeoutMs: 8000`: no settle within 120s before, **13016ms** after.

**7. `sandbox-fabric` loaded the browser build.** `import "fabric"` resolves to the browser entry, which reaches for `document`. Now tries `fabric/node` first, falling back to the default entry for the browser, where this host module *is* the page.

**8. A stale wrapper count.** `dsl-pack-guest` asserted `toBe(472)` against a pack shipping 457 since it was regenerated a day earlier. The number guarded something real — a half-matching regex making the next assertion vacuous — but a hardcoded total does that only until the pack changes. It now counts `createNode(` with a pattern too simple to miss one and compares the two.

## Four timing budgets that were really measuring worker startup

The same defect four times, in four places, and the last two only became visible once the failures in front of them were gone. Spawning the QuickJS worker is the expensive part of a sandbox run — under tsx each spawn recompiles the whole sandbox graph — and a contended CI runner stretches it to seconds. Any budget that starts its clock before the spawn is measuring startup, not the thing it is named after.

**9. Two suites racing the default timeout.** `code-nodes` at 20s and `base-nodes` at the bare 5s default. One run lost six `base-nodes` tests at exactly 5000ms. Both now 60s, matching what `sandbox-compiler` already carried. I enumerated the rest rather than patching only what failed: `core-nodes`, `node-sdk` and `protocol` name `nodetool.code.Code` only as a string in fixtures and spawn no worker.

**10. Two budgets that needed splitting, not raising.**

- `reliability-harness` / `ws-transport-faults` failed `expected 'timeout' to be 'completed'`. The journey declared `timeoutMs: 4000` where every other pipeline journey declares 30s — because that number is also the dead wait the three black-hole faults sit through, waiting for a frame that never arrives. So the completion budget for `ws-delay`/`ws-fragment` was set by how long the suite should spend on runs whose answer is known, and the first run in the file (the one paying the cold spawn, since `nodetool.text.ToUppercase` executes as a Code node) reported `"timeout"` for a transport that was fine. Measured: 1480ms cold against 896ms warm, on a runner 3-4× slower than this one. The manifest now declares the completion budget; the black-hole cases shorten it for themselves, and only for the faulted half — the clean run that follows is the assertion, so it keeps the full budget.
- `agents` / `interrupts a CPU-bound loop once cancellation has landed` asserts the interrupt arrives within 3s, timing from before `runInSandbox`. Cancelling terminates the worker and drops it from the pool, so each of these tests spawns fresh: 904ms cold against 16ms warm here, 11191ms in CI. Both cancellation tests now leave an idle worker in the pool first, so the budget measures interrupt latency (21ms).

## Verification

Every fix was reproduced before it was written and re-broken after, rather than trusted because CI went green:

- The dep-optimizer failure was reproduced locally, same two errors, same file and line.
- Reverting the Opus fallback **with the good fixture in place** brings back the original error — the codec fix is load-bearing, not made redundant by the fixture.
- `tests/e2e-vite-config.test.ts` pins the stub table; both assertions were watched failing (remove `https` from the bare set, add `buffer` to it).
- `tests/sandbox-av-media.test.ts` is the first unit coverage the video bridge has had; two of its six cases fail if the codec is hardcoded again, so they hold even on a host that *has* AAC.
- Two new pool cases pin the lease/ref contract — removing the idle unref fails `holds the process only while a worker is leased`.
- The wrapper guard was checked by injecting a `createNode(SOME_VAR, …)` the detailed regex cannot match: `expected 457 to be 458`.
- The `ws-delay` assertion was re-broken by putting the budget below the run time, printing the message CI printed; the interrupt bound was re-broken by tightening it.

Local: browser E2E **27/27** on two consecutive cold-cache runs · `agents` 2859/2859 · `base-nodes` 2021/2021 · `code-nodes` 212/212 · `sandbox-compiler` 208/208 · `workflow-runner` 23/23 · `reliability-harness` 157 passed / 4 skipped · `image-nodes` 229/229 (needs `mesa-vulkan-drivers`/lavapipe locally, per AGENTS.md) · `npm run lint` and `tsc --noEmit` clean.

Confirmed green in CI on earlier commits of this branch: Workflow Runner Browser E2E, app-build, Run workflow suite in browser, typecheck, lint, build, deps, bundle, docker, test-app, Workflow Integration Tests, Load every route, CodeQL.


georgi:
CI result on `54111e9`: the Vite fix works — **26 passed, 1 failed**, up from 0/27. The harness boots, 27 tests actually run, and there is no `Failed to resolve entry for package` anywhere in the job log.

The one remaining failure is **not caused by this PR and I am not fixing it here** — it needs a call from whoever owns the media harness.

### What is failing

```
[chromium] › e2e/tests/sandbox-media.spec.ts:3:1 › composes and transforms media through browser WebCodecs
Error: page.evaluate: Execution context was destroyed, most likely because of a navigation.
```

That message is misleading — it is what the failure looks like inside the full suite. Run the spec on its own and the real error appears:

```
video.addAudio: required tracks cannot be encoded as MP4
```

### Why

The spec mixes a WAV soundtrack into an MP4, so `Conversion.init` needs an AAC encoder. Probing the encoder directly in the same headless Chromium the suite launches:

```
avc1.42001f: true     vp8:  true
mp4a.40.2:   false    opus: true
```

Chrome ships no AAC **encoder** on Linux — AAC encode goes through Media Foundation on Windows and AudioToolbox on macOS. `audioConversion.utilizedTracks` therefore excludes the audio track and `sandbox-av-media.ts:740` throws. This is deterministic on `ubuntu-latest`, not flaky.

`sandbox-media.spec.ts` and `sandbox-av-media.ts` both landed yesterday in 37f9dbbb, while this job was already red from the dep-optimizer bug — so the spec has most likely never executed in CI, and the gap only became visible once the harness started booting again.

### Options

1. Gate the AAC-dependent assertion on `AudioEncoder.isConfigSupported({codec: "mp4a.40.2"})` and keep the trim / extract-frame / info coverage running everywhere. Green on Linux, full coverage on macOS.
2. Have `video.addAudio` fall back to Opus-in-MP4 when AAC is unavailable — a behavior change to shipped sandbox media, so it wants its own review.

Option 1 is the smaller change, but it trades away CI coverage of a feature that shipped yesterday, and option 2 changes product behavior on every platform. Either is a design decision on someone else's new code rather than a mechanical fix, so I left it alone instead of picking one. **This check stays red until that is decided.**

Worth fixing separately: the in-suite error masks the real one. Whoever picks this up should run `npx playwright test --config e2e/playwright.config.ts e2e/tests/sandbox-media.spec.ts` to see it.

---

georgi:
Went with the Opus fallback (option 2). Pushed in 2853ef7 — **the suite is now 27/27**, verified on two consecutive cold-cache runs.

The codec fix alone wasn't enough. Two more defects were sitting behind it, each hidden by the previous one:

1. **`video.addAudio` hardcoded AAC** → now `getFirstEncodableAudioCodec(["aac", "opus"])`, probed with the source track's channel count and sample rate. AAC still wins where it exists (Node, macOS unchanged); Linux gets Opus, which MP4 also carries.

2. **The MP4 fixture was truncated.** With Opus in, the error became `First packet must be a key packet` — and that one reproduces in **Node** too, so it was never codec-related. The inline `SILENT_MP4_BASE64` declared 707 `mdat` payload bytes and carried 690, leaving 5 of 7 packets readable. Re-encoded with Mediabunny and documented.

3. **`mediabunny` was discovered mid-run.** Nothing imports it statically, so Vite pre-bundled it when the media test first called `video.*` and reloaded the page out from under `page.evaluate`. Cold-cache only — invisible locally on a warm cache, guaranteed on every CI run. Fixed with `optimizeDeps.include`.

That third one is also why this took a few passes to read: `Execution context was destroyed` is what all of these look like from inside the suite, and it masked the real error every time. Running the spec on its own was needed to see each one.

Both media fixes are pinned rather than assumed. Reverting the codec change **with the good fixture in place** brings back `required tracks cannot be encoded as MP4`, so it is load-bearing. And `tests/sandbox-av-media.test.ts` adds the first unit coverage this bridge has had — two of its cases fail if the codec is hardcoded again, so they hold on a host that has AAC, where the browser condition can't be reproduced.

---

georgi:
Two red checks on `2853ef7`. **Neither is caused by this PR**, and one of them reproduces on `main`.

### `Quality Gate / test-packages` — pre-existing on `main`

`@nodetool-ai/sandbox-compiler` fails one assertion:

```
FAIL tests/dsl-pack-guest.test.ts > every node type the DSL offers > exists in the live node registry
AssertionError: expected 457 to be 472
```

That line counts `createNode(...)` calls in the **checked-in** DSL pack under `packages/sandbox-packs/sandbox-dsl/sandbox/generated/`, before the registry is consulted. It reads files on disk, not anything this branch builds:

```
$ grep -ho 'createNode("[^"]*"' .../sandbox/generated/*.js | wc -l
457          # on this branch
457          # on origin/main
```

The pack was regenerated in 218ca2ff (Aug 14, `feat(jsscript): use typed widgets for run-dialog inputs`); the test's hardcoded `472` was last touched in 4cb2d7fe (Aug 13). So the constant went stale a day before this branch existed, and recent `test.yml` runs on `main` are failing.

Worth noting the neighbouring test — *"covers the registry with no duplicate wrapper for one type"* — **passes**, and it asserts the pack is total against the live registry. So 457 is the correct current count and the stale number is the literal `472`, not the pack. The fix is a one-line update, but it belongs to whoever regenerated the pack rather than to this PR, so I have not touched it.

This PR's five files are `packages/agents/src/sandbox-av-media.ts`, its new test, and three under `packages/workflow-runner/e2e/` — none of them the pack.

### `Run workflow suite in browser` — intermittent, watching it

```
Error: manifest should list workflows
Expected: > 0   Received: 0
```

The spec waits up to 90s for `window.__E2E__.manifest()` to fill, swallows the timeout, then asserts. The job's test took 1.6m — the 90s wait plus overhead — so the manifest never arrived in time. That is a startup-timing failure, not a content one.

Evidence it is not this change:
- It **passed** on the previous commit of this same branch (54111e91).
- It passes locally on this branch on both a warm and a deliberately cold Vite cache (2/2, 1.9m and 2.1m).
- The diff adds no new dependency to the web bundle — `getFirstEncodableAudioCodec` is one more named import from `mediabunny`, which `sandbox-av-media.ts` already imported `Conversion`, `Input` and `Output` from.

I'll confirm on the next run rather than call it settled. If it goes red again I'll dig into the 90s manifest wait properly.

### Unaffected

`Workflow Runner Browser E2E` — the job this PR exists to fix — is queued in the current run and passed 27/27 locally on two consecutive cold-cache runs.

---

georgi:
Final CI state for `2853ef7`, compared job-for-job against the latest `main` run ([31880034124](https://github.com/nodetool-ai/nodetool/actions/runs/31880034124), head `fcea107`):

| Job | `main` | this PR |
| --- | --- | --- |
| **Workflow Runner Browser E2E** | ❌ failure | ✅ **success** |
| `test-packages` | ❌ failure | ❌ failure |
| `app-build` | ⚪ cancelled (45 min) | ⚪ cancelled (45 min) |
| `quality` (aggregate) | ❌ failure | ❌ failure |
| changes, lint, deps, build, typecheck, bundle, test-app, docker, Workflow Integration Tests | ✅ | ✅ |

Every red on this PR is red on `main` in the same job for the same reason. The one difference is the job this PR exists to fix, which is now green.

Two notes on the remaining reds, neither of them this branch's:

- **`app-build` is a 45-minute job timeout, not a test failure.** It ran 43 minutes and was cancelled at the job's `timeout-minutes: 45` — on `main` (11:15:51 → 11:59:00) and here (12:48:51 → 13:32:12) alike. Identical duration, so it is hanging in the same place regardless of this diff. Its `cancelled` result is what drags the `built` leg (and therefore the `quality` aggregate) down alongside `test-packages`.
- **`test-packages`** is the stale `expect(wrappers.length).toBe(472)` against a 457-wrapper checked-in DSL pack, covered in the comment above.

That accounts for every failing check. I'm not touching either in this PR — both predate it and neither is in its five files — but they do mean `main` cannot go green until someone updates that constant and looks at why `app-build` hangs.

---

georgi:
Pushed b7bf473 taking on the two remaining reds. Both were red on `main` too, so this fixes them there as well.

### `app-build` — a 45-minute timeout on a 0.8-second job

The suite was never slow. Both deterministic cases finish green in **0.8s**; the process then sits idle until CI kills it at the job's `timeout-minutes: 45`.

Probing the hung process, its only ref'd handle is a `MessagePort` — the sandbox worker's public port:

```
=== ACTIVE (ref'd) handles: 2
>>> handle=Socket      (stdout)
>>> handle=MessagePort hasRef=true
```

`defaultSandboxWorkerFactory` called `worker.unref()` and *then* attached the `message` listener. Attaching one starts the worker's public port, which **re-refs it**, so the unref was silently undone. Isolated, with nothing else involved:

```
unref-first      HANGS (held open)
listener-first   exits cleanly (code 0)
```

Moving the unref after the listeners makes it take effect — but an unref'd worker on its own would let the event loop drain while a run is still in flight. So the ref now follows the lease: `acquire` refs, returning to idle unrefs, and a worker handed straight to a waiting run stays ref'd. `wasm-sandbox/workers.ts` had the identical ordering and is fixed with it.

Result: `app-build` exits in **5s**, both cases green. And a Code node that sleeps 3s still returns `42`, so the in-flight case is covered rather than traded away.

### `test-packages` — a count that could only go stale

`dsl-pack-guest` asserted `wrappers.length === 472` against a pack that has shipped 457 since 218ca2ff regenerated it, a day before this branch existed.

The number was there for a real reason — to stop a half-matching regex making the next assertion vacuous — but a hardcoded total does that only until the pack changes. It now counts `createNode(` with a pattern too simple to miss one and compares the two, so the same guard holds and cannot go stale. Verified it still bites: injecting a `createNode(SOME_VAR, …)` the detailed regex can't match fails the test with `expected 457 to be 458`.

### Verification

- `app-build`: exits 0 in 5s, 2/2 green (was: cancelled at 45 min).
- `sandbox-compiler`: 208/208.
- `agents`: 2856 passed, with two new cases pinning the lease/ref contract — removing the idle unref fails `holds the process only while a worker is leased`.
- Workflow Runner Browser E2E: 27/27 on a cold cache.
- `base-nodes` 2021/2021, `image-nodes` 229/229 (needed `mesa-vulkan-drivers`/lavapipe locally, per AGENTS.md), `npm run lint` clean, `tsc --noEmit` clean.

### Still outstanding

`Run workflow suite in browser` failed once on the previous commit and I have not been able to reproduce it: it passes locally on warm and cold Vite caches, passed on this branch's first commit, and `main` is green on it across the last 30 runs. Its failure mode is the 90s `window.__E2E__.manifest()` wait expiring — startup timing, not content. This push re-runs it; if it goes red again I'll dig into that wait rather than keep calling it intermittent.

---

georgi:
`b7bf473` landed the worker fix, and CI confirms both of the things it targeted:

| Job | before | on `b7bf473` |
| --- | --- | --- |
| `app-build` | ⚪ cancelled at 45 min | ✅ **success in 1m47** |
| `Run workflow suite in browser` | ❌ failure | ✅ success |
| Workflow Runner Browser E2E | ✅ | ✅ |
| typecheck, lint, build, deps, bundle, docker, test-app, Integration Tests, Load every route, CodeQL | ✅ | ✅ |
| `test-packages` | ❌ (sandbox-compiler) | ❌ (code-nodes — different, see below) |

`app-build` going from a 45-minute timeout to under two minutes is the ref-ordering fix working on a real runner. `sandbox-compiler` passes. And the browser suite flake did not recur, which is the second clean run for it.

### The one remaining failure, and why it is not a regression

`test-packages` failed on `code-nodes`, but on a different thing:

```
FAIL tests/code-node-script-link.test.ts > runs the materialized body with no resolver on the context
Error: Test timed out in 20000ms.
```

It is the **first** test in the file — the one that pays for the QuickJS worker spawn. The same CI run measured the other cold spawns at **7.2s** and **12.9s**, both passing. Warm, this test takes **1.4s**; it passes 3/3 locally.

The config had already been raised from vitest's 5s default for exactly this, with a comment saying the worker's "startup cost can exceed the default 5s under CI's contended parallel task load". Under tsx every spawn recompiles the whole sandbox graph, and CI runs ten of these files against a contended runner, so the first spawn in a file is the slowest thing in the suite. 20s wasn't a wrong bound, just one with no margin.

Worth stating plainly, since the timing changed in the same commit that touched the worker: **the ref change cannot affect spawn speed.** It alters ref state only while a worker is *idle* — during a run the worker was ref'd before this PR (the `message` listener re-ref'd it) and is ref'd after. There is no path by which it makes a spawn slower.

3d24279 raises it to 60s with the measurements recorded in the comment, so the next person sees the numbers rather than an unexplained constant.

---

georgi:
`3d24279` fixed `code-nodes`, and `test-packages` moved on to `base-nodes` — same root cause, different package. 0eb3f8e closes it.

Six tests failed, **every one at exactly 5000ms** — vitest's default:

```
FAIL tests/code-node-examples-run.test.ts   > aggregates six orders by region…
FAIL tests/code-node-examples-run.test.ts   > parses an Atom feed…
FAIL tests/gallery-examples-run.test.ts     > Chunk a Transcript…
FAIL tests/gallery-examples-run.test.ts     > Redact and Tidy a Log Line…
FAIL tests/pure-node-examples-run.test.ts   > applies each string transform…
FAIL tests/pure-node-examples-run.test.ts   > matches, extracts, filters…
Error: Test timed out in 5000ms.   (×6)
```

All six execute example graphs containing a Code node, which spins up the QuickJS worker thread. Under tsx that spawn recompiles the whole sandbox graph, so the first one in a file is the slowest thing in the suite — and `base-nodes` was running it on the bare 5s default.

Every other package that spawns this worker already has an allowance: `code-nodes` and `sandbox-compiler` at 60s, `agents`/`execution`/`dsl`/`websocket` at 30s. `base-nodes` had none, so it was the one left racing the default.

Rather than fix only the package that happened to fail, I enumerated the rest: `core-nodes`, `node-sdk` and `protocol` name `nodetool.code.Code` only as a string inside graph fixtures and never call `runInSandbox`, so they spawn no worker and keep the default. `base-nodes` was the only real gap.

I should have caught this two pushes ago — `pure-node-examples-run.test.ts > applies each string transform` failed at 5027ms in one of my local full-suite runs and I recorded it as load flake instead of following the 5000ms to its cause. It is the same test in the list above.

### Where the PR stands

Confirmed green on `3d24279`: Workflow Runner Browser E2E, app-build, Run workflow suite in browser, typecheck, lint, build, deps, bundle, docker, test-app, Workflow Integration Tests, Load every route, CodeQL. `test-packages` is the last one, and this is the fix for it.

---

georgi:
`0eb3f8e` fixed `base-nodes`, and `test-packages` reached `agents` — the last package in the chain. fe51c4d fixes it. **agents now passes 2859/2859**, the first fully green run it has had.

Both failures were red on `main` and had never been *seen*: turbo stopped at an earlier failing package every time, so this suite never ran to completion in CI. Each fix I pushed let it get one package further — 58 → 56 → 60 → 62 successful tasks — which is why this looked like whack-a-mole rather than one queue draining.

### `sandbox-fabric` — the browser build on Node

```
FAIL host-modules.test.ts > sandbox-fabric in the guest > renders a scene to SVG
FAIL host-modules.test.ts > sandbox-fabric in the guest > exports data URL
```

The module imported `"fabric"`, which resolves to the package's **browser** build. Constructing a canvas there reaches for `document`:

```
ReferenceError: document is not defined
  at dt.createLowerCanvas (fabric/dist/index.min.mjs)
```

Fabric ships `fabric/node`, which wires a real canvas backend — verified directly: `toSVG` and `toDataURL` both work through it. That subpath does not resolve in a browser, where this host module *is* the page, so it is tried first and falls back to the default entry rather than branching on a runtime flag.

### The marshaling test — a 30-minute floor under every run

That test wasn't slow, it *could not finish*. The worker host sized its backstop as:

```ts
timeoutMs + Math.max(0, suspendAllowanceMs) + DEADLINE_MARGIN_MS
```

unconditionally — and `DEFAULT_SUSPEND_ALLOWANCE_MS` is **30 minutes**. A run with no clock cannot suspend at all, so a wedged worker outlived its own `timeoutMs` by half an hour. I confirmed the shape by dropping `timeoutMs` to 8s: still no settle after 120s.

The contract was already documented — *"Only consulted when `clock` is set"* — and both other consumers of the allowance already honour it: the engine abort in `js-sandbox.ts` and `executionTimeout` in `interpreter.ts` both gate on `hasClock`, and the run message already carries the flag. This one backstop ignored it.

Measured on the case the test describes (a host global returning a raw 16 MB `Uint8Array`, which OOMs the guest mid-marshal), `timeoutMs: 8000`:

| | settle |
| --- | --- |
| before | none within 120s |
| after | **13016ms** — the timeout plus the 5s margin |

and it now reports `the sandbox worker did not answer within 13000 ms and was terminated` instead of hanging. That is a real robustness bug in the sandbox, not just a test problem: any run whose worker wedges was unbounded for half an hour.

### Where this leaves the PR

Green on the previous commits and unaffected by this one: Workflow Runner Browser E2E, app-build, Run workflow suite in browser, typecheck, lint, build, deps, bundle, docker, test-app, Workflow Integration Tests, Load every route, CodeQL. `test-packages` was the last red and `agents` was its last package.

---

PR #4990, opened 2026-08-17
Bolt: O(N^2) Array finding optimization in Dataframe Components

georgi:
## What
Replaced a nested array search (`Array.find()`) inside a `while` loop with a pre-computed `Set` in the `addColumn` functions of `DataframeProperty`, `DataframeEditorModal`, and `RecordTypeProperty`.

## Why
When adding a new column, the code generated a unique name ("Column N") by checking if the name already existed. It did this by calling `columns.find(...)` inside a `while` loop that incremented `N`. Since `find` is $O(N)$, doing this inside a loop that increments up to the number of existing columns resulted in $O(N^2)$ time complexity. For dataframes with a large number of columns, this iterative full-array search blocks the UI thread unnecessarily.

## Impact
Reduced the complexity of finding a unique column name from $O(N^2)$ to $O(N)$. By extracting the column names into a `Set` before the loop, the `has()` check inside the loop is $O(1)$. 
In a benchmark with 5000 columns, the uniqueness generation dropped from ~278ms to ~3.3ms (an ~80x speedup).

## Verification
- Wrote a standalone test script simulating 5000 columns to verify the performance claim and correctness.
- Ran `npx tsc --noEmit` in `web` (ignoring pre-existing errors, no new errors introduced).
- Ran `npx oxlint web/src`.
- Ran `npm run test` on `web/src/components/properties/` to ensure no regressions in property component rendering and interaction.

---
*PR created automatically by Jules for task [5489065635338318794](https://jules.google.com/task/5489065635338318794) started by @georgi*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -89,3 +89,6 @@
 ## 2026-05-25 - O(N*M) lookup optimization in eval case selection
 **Learning:** Found an $O(N \times M)$ performance bottleneck in `packages/cli/src/commands/eval.ts` where `picked.some((c) => c.id === id)` was called inside `[...wanted].filter(...)`. When evaluating a large suite with many specific cases requested, this nested array iteration creates unnecessary overhead.
 **Action:** Replaced `.some()` inside the `.filter()` loop with an $O(1)$ `Set` lookup. Extracting the `picked` IDs into a `Set` before the loop reduces the complexity from $O(N \times M)$ to $O(N + M)$.
+## 2026-05-25 - O(N^2) Array finding optimization in Dataframe Components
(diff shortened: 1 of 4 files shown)


georgi (commented):
Reviewed and verified independently:

- All three `addColumn` functions on main use the `while (columns.find(...))` pattern exactly as described.
- The `Set` pre-computation is correct — `Set.has()` is semantically equivalent to the `find()` check for uniqueness.
- Searched the entire codebase for similar `while (*.find(` patterns — these three files are the only occurrences, so nothing was missed.
- `.jules/bolt.md` pre-exists on main (not a new file).
- No cross-PR conflicts with other open PRs (#4989 was just merged, touches only test files in `packages/agents/`).
- 18/19 CI checks passed; `test-packages` is still running. This PR only touches frontend component files, so the backend test suite is unlikely to be affected. Ready to merge once that check finishes green.

---

PR #5006, opened 2026-08-17
fix(web): offer `any` handles in the connectable-nodes menu

georgi:
## What

Dragging a connection onto empty canvas opens a menu of nodes that can take the edge. The filter behind it (`filterTypesByInputType` / `filterTypesByOutputType`) dropped every node whose only compatible handle was typed `any`:

```ts
// Exclude matches that are only connectable because the output is "any"
if (output.type.type === "any") { return false; }
```

So an `any` output — a Code node, a dynamic output, an untyped slot — could never be wired to a typed input from that menu, even though the connection is legal everywhere else: `isConnectable(any, T)` returns true, the direct handle-to-handle drag creates the edge, `Graph.validateEdgeTypes` accepts it, and the graph validator treats `any` as permissive.

## Changes

- The `any` exclusion is a **ranking** concern now, not a filter. An `any` match sorts below every typed match (new priority 3; "no match" moved to 4), so exact matches still lead the menu and `any`-only nodes are reachable at the bottom.
- `ConnectableNodes.getPreferredConnectableInput` prefers a typed property over an `any` one when the node has both, so the auto-created edge lands on the meaningful slot.
- The output side had a related defect: it wired `metadata.outputs[0]` blindly, which can be an output the target input never accepted. It now picks the first *compatible* output, typed before `any` (`getPreferredConnectableOutput`).
- The NodeMenu's explicit type-filter dropdown (`filterDataByType`) keeps its strict semantics — selecting "any" there still means "nodes with an `any` handle".

## Verification

- Six new tests in `web/src/components/node_menu/__tests__/typeFilterUtils.test.ts` cover both directions, the ranking, and that incompatible nodes are still dropped. Restoring the old exclusion turns three of them red, so they bite.
- `web` typecheck, lint, and the full Jest suite pass: 1135 suites / 13137 tests.
- Root `npm run lint` fails in this container on a missing `@oxlint/plugins` package (config load, before any file is read) — an install gap unrelated to this diff; `npm run lint` inside `web/` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_018dpnHyuDathH6HnUPAX8Kn)_


georgi:
CI is red on two checks; neither comes from this diff.

**`lint` — fails on the base branch.** `lint:design` reports two `design-tokens/spacing-tokens` errors in `web/src/components/node/SubgraphNode/SubgraphNode.tsx:81`, a file this PR does not touch. It arrived on `main` in 16cbba41 ("feat(editor): workspace chip, inline subgraph inputs, subgraph open button"):

```tsx
<Box sx={{ px: "8px", pt: "8px", flexShrink: 0 }}>
```

Reproduced by linting `main`'s copy of that file on its own (2 errors, exit 1); linting this branch's `web/src` end to end passes (exit 0). The fix belongs on `main`, not here — `px: SPACING.md, pt: SPACING.md` (8px on the 4px grid).

**`changes` — infrastructure.** Both attempts died fetching `dorny/paths-filter` from codeload (429, then 502, then 429), before any check body ran. Re-runnable, nothing to diagnose.

I'll merge `main` in and re-run once the spacing error is fixed there.

---

PR #5014, opened 2026-08-17
fix(dashboard): decide the page before rendering it

georgi:
## What

The dashboard picks between two layouts — the first-run page (welcome hero, tutorials, templates) and the returning page (compact command bar, two columns of the user's own work). `useDashboardMode` answered `"first-run"` while the workflow count was still loading.

For a user with unfinished onboarding and a workflow library, that means the first-run page paints, then the returning page replaces it. Both orders carry the same sections — templates, workflows, documents, tutorials — so the swap reads as duplicated or uncleaned content rather than as loading.

## Change

- `useDashboardMode` returns `"pending"` while the answer is unknowable, instead of guessing. New type `DashboardModeState = DashboardMode | "pending"`.
- `Portal` renders nothing below the downloads banner until the mode is decided. `DashboardHero` still takes a decided `DashboardMode`; TypeScript narrows it at the call site.

Onboarding is persisted (synchronous), so a user who has finished it is answered on the first render and waits for nothing. The wait only applies to the cold-cache case that used to flash, and only until the workflow list arrives — a warm TanStack cache makes `isLoading` false, so navigating back to the dashboard in-session renders once.

## Tests

- `useDashboardMode` test updated for the new `"pending"` answer, plus one that rerenders through the load and asserts the sequence is `pending → returning`, never `first-run → returning`.
- `npx jest src/components/portal src/hooks/__tests__/useDashboard` — 10 suites, 64 tests, green.
- `oxlint` clean on both changed files. Web typecheck reports only the pre-existing `@nodetool-ai/audio-nodes` / `image-nodes` module-resolution errors from unbuilt packages; nothing in the touched files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AYXQsfeKEqti1igWR1FsdC)_

### DIFF
--- a/web/src/components/portal/Portal.tsx
+++ b/web/src/components/portal/Portal.tsx
@@ -27,12 +27,7 @@ import DashboardFooter from "./DashboardFooter";
 import { useStartTrackChat } from "../../hooks/useStartTrackChat";
 import { openSettingsTab } from "../workspace/openPageTab";
 import { WELCOME_TRACKS, type WelcomeTrackId } from "./welcomeTracks";
-import {
-  Box,
-  BORDER_RADIUS,
-  SPACING,
-  getSpacingPx
-} from "../ui_primitives";
+import { Box, BORDER_RADIUS, SPACING, getSpacingPx } from "../ui_primitives";
 
 const styles = (theme: Theme) =>
   css({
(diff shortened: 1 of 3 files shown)


georgi:
CI is red on the three jobs that build the backend packages (Quality Gate `build`, Workflow Integration Tests, Workflow Runner Browser E2E), all on the same failure:

```
packages/runtime/src/providers/custom-provider-registry.ts(33,5): error TS2345:
Argument of type 'typeof CustomOpenAIProvider' is not assignable to parameter of type 'ProviderConstructor'.
  Type 'Record<string, unknown>' is missing the following properties from
  type 'CustomOpenAIProviderConfig': _providerId, _baseUrlKey, _apiKeyKey
```

That is failing on the base branch too — main's own run for 13f60c3d (the commit this branch is based on) is red with it, and it reproduces in a clean checkout of that commit with no changes from this PR applied. It arrived with 76dd17c5 "feat(providers): add OpenAI-compatible endpoints from settings": `registerProvider` wants `new (kwargs: Record<string, unknown>)`, and `CustomOpenAIProvider`'s constructor takes the narrower `CustomOpenAIProviderConfig`.

This PR is web-only — two files under `web/src` — so it cannot affect the runtime build. I'll re-run CI when main recovers.

---

PR #5048, opened 2026-08-18
revert(marketing): restore the previous hero headline and subhead

georgi:
Puts `NodeToolHero.tsx` back to the copy it had before #5046 — H1 returns to "The agent-first creative workspace" and the subhead to its previous wording.

Hero only. `marketing/NARRATIVE.md` and `StatusQuoSection.tsx` from #5046 are untouched.

## One thing left inconsistent, deliberately

`NARRATIVE.md` records "Describe the piece. Keep the workflow." as the positioning line and states that the category descriptor is not the H1. With this revert, the site does not match that. I did not edit the doc, because which of the two is right is the call to make, not the mismatch:

- keep the doc as the target and the hero as it stands today, or
- change the doc's positioning line back to the category descriptor.

Say which and I will follow up.

## Verification

Diff is 7 insertions / 6 deletions inside one JSX block — no props, imports, or structure touched; the file parses clean through esbuild's tsx loader. As before, `marketing/` has no installed `node_modules` in this environment, so its `tsc` fails repo-wide on unresolved imports regardless of this diff. No test asserts the hero string; the Playwright smoke test only counts h1 elements, and the count is unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TGJSZvvscvrHCvqUAyR2Mf)_

### DIFF
--- a/marketing/src/components/NodeToolHero.tsx
+++ b/marketing/src/components/NodeToolHero.tsx
@@ -26,10 +26,10 @@ export default function NodeToolHero() {
             id="hero-title"
             className="mt-6 text-5xl font-bold leading-[1.05] tracking-tight text-white md:text-6xl"
           >
-            Describe the piece.
+            The agent-first
             <br />
             <span className="bg-gradient-to-r from-rose-400 via-fuchsia-400 to-amber-300 bg-clip-text text-transparent">
-              Keep the workflow.
+              creative workspace.
             </span>
           </h1>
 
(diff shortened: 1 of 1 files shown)


georgi:
CI red on **Workflow Runner Browser E2E** — 27 of 27 tests in `packages/workflow-runner/e2e` failed, including `browser-workflow.spec.ts:33 › harness boots in a real browser`. This is not from this diff, and it is failing on the base branch too:

- `main` @ `581c96a8` — job green (run 32139828840). That was this PR's base until an hour ago.
- `main` @ `2effc760` — **same job failed** ([job 95733049305](https://github.com/nodetool-ai/nodetool/actions/runs/32143925007/job/95733049305)), on the push-to-main run, with no PR involved.
- This PR's base is `2effc760`, and its diff is 7 lines of hero copy in `marketing/src/components/NodeToolHero.tsx` — a file the workflow-runner harness never loads.

Between those two commits sits only #5047, which moved 118 lines out of `packages/agents/src/js-sandbox.ts` into `network-guard.ts` / `workspace-paths.ts`. Sixteen of the failures are `sandbox-modules.spec.ts` cases driving `SANDBOX_MODULE_FIXTURES` from `@nodetool-ai/agents`, and the whole-suite failure pattern (the boot test fails too) reads like the browser bundle failing to build rather than individual assertions breaking. Circumstantial, not confirmed — I have not bisected it.

Not fixing it here: it belongs to `main`, not to a copy revert. I will merge the base in and re-run once `main` is green again.

---

PR #5054, opened 2026-08-18
chore: update package dependencies across multiple packages

heavy-d:
## chore(deps): npm audit fix + manual bumps for the leftovers

`npm audit fix` cleared the auto-fixable set but left several "fixable" advisories in place (a second pass changed nothing). This bumps those by hand.

**68 → 33 vulnerabilities (critical 1 → 0, high 22 → 15).**

| Package | Change | Fixes |
|---|---|---|
| `web/react-router-dom` | `^7.16.0` → `^7.18.2` | 5 react-router advisories (high) |
| `@openclaw/fs-safe` (storage, automation-nodes) | `^0.1.2` → `^0.5.6` | 0.1.2 pinned `tar@7.5.13`; 0.5.6 has no deps. We only use `root`/`FsSafeError` |
| `tar` | lock refresh → 7.5.22 | the critical (`cmake-js` accepts `^7.5.6`) |
| `agents/mailparser` | `^3.9.3` → `^3.9.15` | linkify-it, nodemailer (high) |
| `agents/imapflow` | `^1.2.12` → `^1.7.1` | drops nodemailer dependency |
| `cli/@inquirer/prompts` | `^5.0.0` → `^8.5.2` | tmp/external-editor (high); only `confirm`/`input` used, same API |
| `runtime/@opentelemetry/*` | sdk-node + exporter-trace-otlp-proto `^0.221.0`, resources + sdk-trace-base `^2.10.0` | all 19 OTel core entries (moderate) |

Verified: `tsc --noEmit` clean for runtime, agents, cli, storage; storage, cli, and agents email suites pass. Remaining runtime/cli/agents failures on my machine are pre-existing Windows-only issues (`spawn EFTYPE`, symlink/chmod, path-separator asserts).

**Left alone (33):** no upstream fix (`expr-eval`, `adm-zip` via epub2, `ip` via node-av); real majors or downgrades npm mislabels as fixes (`monaco-editor`/dompurify, `electron` 39→43, `pptxgenjs`, `office-text-extractor`, `exceljs`, `drizzle-kit`, `vite-plugin-node-polyfills`); dev-only transitive (`qs` under stryker, `deepmerge-ts` inside html-to-text@10).

--

- Updated `imapflow` to version 1.7.1 in `packages/agents/package.json`.
- Updated `mailparser` to version 3.9.15 in `packages/agents/package.json`.
[...]

PR #5070, opened 2026-08-19
feat(code): infer sandbox packages from imports, drop the packages setting

georgi:
A Code node declared its sandbox packages twice: once in a `packages` property a picker wrote, and once in the `import` statements the body actually made. Two lists that had to agree, and every disagreement was a run that failed on an import the body plainly makes. A JS script already resolved by import alone, so the same body behaved differently in the two places.

The imports are now the declaration everywhere.

## What changed

- **`nodetool.code.Code`** parses its own body and resolves the specifiers it imports against the installed catalog. The `packages` property is gone. This is the rule `mountJsScriptSandbox` already ran, now shared as `resolveImportedPacks` (`packages/agents/src/js-script-sandbox.ts`) and used by the JS script path, the hermetic `run_code`/`test_code` harness, and the browser runner's prefetch.
- **`validateCodeNodeBody`** drops `declaredPackages` / `allowInstalledPackages` and the `code_unused_package` warning; every caller now gets one behavior — imported specifiers checked against the catalog.
- **Removed with the property**: `SandboxPackagesProperty` and the `sandbox_packages` editor type, `ui_code_set_packages`, `ui_jsscript_set_packages`, the `packages` field on `JsScriptDocument`, and `parseSandboxModuleDeclarations` (no callers left). A leftover `packages` key on a saved graph is ignored, not turned into a dynamic input.
- **Two new static checks** replace what the undeclared-import error used to catch, and both are decidable with no catalog present: an import of a Node builtin (the guest is QuickJS and has none), and a direct import of a `nodetool:` host bridge, which the loader serves only to a generated facade.
- A legacy `yield`-contract body is not valid module syntax, so the node reads its imports off the same `yield` → `yield_` rewrite it runs — a legacy streaming body can still import a pack.
- Docs and prompts that told an author to declare a pack are updated: 38 `SKILL.md` files, `AGENTS.md`, `docs/javascript-sandbox.md`, `docs/sandbox-package-design.md`, the editor chat prompt, and the Code assistant prompt.

## Verification

- `npm run build:packages`, `npm run test:packages` (108/108 suites), `npm run typecheck` (web + electron), `npm run lint`, full web Jest (1146 suites), electron Jest (63 suites).
- `nodetool harness gate` — 6/6 selfchecks, after regenerating the one `linear-text-pipeline` reliability golden that recorded the removed property (diff is exactly that key).
- The new checks were shown to fail, not just pass: a body importing `@nodetool-ai/sandbox-not-real` and `fs` produces `code_package_unavailable` and `code_module`; the same workflow unmodified validates clean.
- End to end with nothing declared: `nodetool node run nodetool.code.Code --props '{"code":"import { count } from \"@nodetool-ai/sandbox-tokens\"; …"}' --no-secrets` → `{"output":4}`.

Image-node tests need a Vulkan ICD locally (documented in AGENTS.md); with lavapipe they pass 231/231.

https://claude.ai/code/session_01HqPoa2gJvU78qEYh9CYRQv

---
_Generated by [Claude Code](https://claude.ai/code/session_01HqPoa2gJvU78qEYh9CYRQv)_


georgi:
This PR has merge conflicts (`mergeable_state: dirty`) — main has moved forward since it was opened.

Additionally, the **Workflow Runner Browser E2E** check failed (the check output was empty, so the failure cause is not visible from the summary) and **Load every route without errors** was cancelled. The rest of the Quality Gate (build, typecheck, lint, deps, bundle, app-build, test-packages, test-app, docker) passed.

No file overlap with the other open PRs (#5062, #5054), so no cross-PR conflicts.

To unblock: merge main into this branch to resolve conflicts, then the full CI suite will re-run and the E2E failure can be investigated.

---

georgi:
Both are handled; main is merged (twice now — `702e05df`, then `bee00713`), and the E2E failure had a cause worth naming since the check output was empty.

Nine `sandbox-modules` deny cases reported `LOADED`: the denied import succeeded. The guest module host was built only when a run resolved at least one module, so a run that resolved nothing got no loader — and the QuickJS wrapper's own Node-compat modules stayed reachable. `import "node:buffer"` loaded and ran.

That was reachable before this branch, for any Code node that declared no packages. Inferring packages from imports made it the common case, which is why the browser suite caught it now. The Node suite never did, because it hands each fixture's modules straight to `runInSandbox` and so never ran an empty resolution.

`createGuestModuleHost` now always returns a host, and the interpreter builds one unconditionally: with nothing resolved it serves no specifier and denies each by name. Each deny fixture also runs a second time with no modules at all — ten of them fail against the pre-fix loader, which I confirmed by checking the old file back out.

---

PR #5130, opened 2026-08-22
feat(execution): run the workflow preflight in ExecutionSession

georgi:
Implements item 2 of `docs/failure-mode-roadmap.md`.

## The gap

`runWorkflow` refuses bad models, providers, and credentials before the job row
exists. `ExecutionSession.create` did not, so every host that goes through the
facade instead of the run service — the CLI `workflows run` and `run`, the
debug harness, the websocket runner, the headless job runner, the app
simulator — still failed at the node that needed the key, after the upstream
half of the graph had already run and billed.

## What changed

- **`packages/execution/src/preflight.ts` (new).** `modelSelectionErrors` and
  `unconfiguredProviderErrors` move here, out of `service/workflow-run.ts`,
  because that module imports the models database and `session.ts` must not.
  Adds `ExecutionPreflightError` — `.issues` is the machine-readable list,
  `.message` the text a host prints — plus `collectPreflightIssues` and
  `assertPreflight`. `workflow-run.ts` re-exports the moved symbols, so its
  public surface is unchanged.
- **`ExecutionSession.create`** selects its `ProcessingContext` ahead of the
  bridge (the preflight resolves credentials through `context.getSecret`) and
  refuses after normalization — before `bridgeFactory` runs and before
  `persistence.onAccepted`, the two side effects that make a late refusal
  expensive.
- **New options `catalogs` and `providerConfiguration`,** both defaulting to
  the process-wide provider registry — the same implementations `runWorkflow`
  uses. A host whose providers are not in that registry passes both; passing
  one and not the other would refuse a graph it can in fact run, so the option
  doc says so.
- **`runWorkflow`** builds its 400 from the same issue list through
  `formatPreflightDetail`, so the refusal text is one contract rather than two
  copies.

## Hosts

Each `ExecutionSession.create` host either answers the refusal on its own
channel or is listed as a propagator with who answers it:

| Host | Disposition |
|---|---|
| `unified-websocket-runner.ts` (job start) | terminal `job_update` via `emitBeforeRunFailure` — a bare throw reached `handleCommand` as a generic `invalid_command` the UI never associates with the job |
| `unified-websocket-runner.ts` (`run_node`) | `{ error }` bag, like its other preparation failures |
| `headless-job-runner.ts` | finalizes the already-created job row — a throw stranded it at `running` |
| `cli/src/debug/server-runner.ts` | a failed `ServerRunReport`; the harness's job is to report why a run did not happen |
| `execution/service/app-run-server.ts` | a failed run report the simulator complains about |
| `cli/src/nodetool.ts` | prints `error.message` alone, so the secret name is not buried behind the class name |
| `agents/execute-agent-graph.ts`, `workflow-agent.ts` | propagate to the agent loop, detaching their forwarding listener on the way out |
| `cli/src/run-dsl-supervised.ts`, `cli/src/evals/graph-runner.ts` | propagate; documented in the audit |

## Tests

`packages/execution/tests/session-preflight.test.ts` (11 cases): unknown
provider, unknown model, missing credential, credential resolved through an
injected context, credential resolved through the default (store-backed)
resolver and through env, model issues reported alone, a custom-provider host's
own catalogs, no bridge connection and no `persistence.onAccepted` on refusal,
and a graph with no model selection left untouched.

A new audit in `execution-session-hydration-audit.test.ts` enumerates every
host and asserts it found more than five before checking them.

Both gates were inverted once:

- `if (false) await assertPreflight(...)` in `session.ts` → **6 of 11 cases
  fail** (`npx vitest run tests/session-preflight.test.ts`).
- Removing one propagator entry → the audit names the file
  (`npx vitest run tests/execution-session-hydration-audit.test.ts`).

Suites run: `execution` (275 passed), `cli` (631), `agents` (3308),
`websocket` (2320). Lint and typecheck clean on every changed file.

## Second commit: a red test this PR did not cause

`workflow-run-credentials.test.ts` → *"ignores a saved model replaced by an
incoming data edge"* was already failing on the branch base. `normalizeGraph`
prunes an edge whose source or target is not in the node list, and the fixture's
`model-source` was never declared — so the edge was pruned before it could
shadow the saved model, and the case asserted the opposite of what it
exercised. The pruning is correct, so the fixture is the bug; the neighbouring
bypass case already declares its source node and this one now does too.

Verified both ways: with the phantom source the helper reports one credential
error, with a real source it reports none — the edge does the shadowing the
case is named for.

## Not from this change

`packages/cli` build fails with `Cannot find module '@nodetool-ai/telegram'`.
`packages/telegram` exists but is not linked into `node_modules` in the
sandbox this was developed in; CI does not hit it.

### DIFF
--- a/docs/failure-mode-roadmap.md
+++ b/docs/failure-mode-roadmap.md
@@ -42,38 +42,7 @@ Acceptance:
 - Production without the enable flag returns 404.
 - `npm run test --workspace=packages/websocket` passes.
 
-## 2. Run workflow preflight in `ExecutionSession`
-
-Status: OPEN.
-
-`runWorkflow` rejects bad models, providers, and credentials before job
-acceptance. `ExecutionSession.create` does not, so `workflows run` can fail
-after upstream work has started.
-
-Implementation:
-
-1. Add optional model catalogs and an optional provider-configuration checker
-   to `ExecutionSessionOptions`. Default both to the runtime implementations
-   used by `runWorkflow`. A custom provider host must supply both.
-2. Select the `ProcessingContext` before bridge connection.
-3. After normalization, call `modelSelectionErrors` and the selected
-   provider-configuration checker.
-4. Resolve secrets with `(key) => context.getSecret(key)`. Do not import the
-   models database into `session.ts`.
-5. Refuse before Python bridge connection and `persistence.onAccepted`.
-6. Add a typed `ExecutionPreflightError` whose issue list is the common refusal
-   contract used by the CLI and service adapters.
-7. Update all hosts enumerated by
-   `execution-session-hydration-audit.test.ts`. Custom-provider hosts must pass
-   their catalogs.
-
-Add `packages/execution/tests/session-preflight.test.ts`. Cover unknown models,
-unregistered providers, missing credentials, default and injected secret
-resolvers, no bridge connection, and no persistence acceptance.
-
-Run `npm run test --workspace=packages/execution -- session-preflight`.
-
-## 3. Include missing secrets in `validate_workflow`
+## 2. Include missing secrets in `validate_workflow`
 
 Status: OPEN.
 
@@ -97,7 +66,7 @@ Tests must show that a missing key produces an issue, an available key clears
 it, and an absent callback preserves current output. Run
 `npm run test --workspace=packages/agents -- workflows`.
 
-## 4. Make media resolution the browser rendering boundary
+## 3. Make media resolution the browser rendering boundary
 
 Status: OPEN. Follow-up to #4873, #4929, #5028, #5078, #5122, and #5123.
 
@@ -122,7 +91,7 @@ Acceptance: upload -> graph input -> render works for image, video, and audio;
 the primitives cover stored locators and HTTPS URLs; the lint fixture proves
 the rule can fail.
 
-## 5. Preserve provider behavior through decorators
+## 4. Preserve provider behavior through decorators
 
 Status: OPEN. Issue #5109.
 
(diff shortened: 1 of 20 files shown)


georgi:
CI red on **Quality Gate / test-app**, failing on base too — not pushing a change for it.

The failure is `web/src/components/appbuilder/__tests__/AppDataPanel.test.tsx` › *"pins a named document of the selected kind"*, 1 failed of 13,346:

```
Unable to find role="button" and name `/add resource/i`
  at AppDataPanel.test.tsx:294
```

**Why it isn't this PR's.** The diff touches no file under `web/`, `electron/`, or `mobile/` — it is 17 files under `packages/` plus two docs. So this branch's web tree is byte-identical to `main`'s.

**Reproduced on base.** Checking out `main`'s own copy of that test (`git checkout origin/main -- web/src/components/appbuilder/__tests__/AppDataPanel.test.tsx`) and running it here fails identically, at the same line 294 CI reports. Notably `main`'s newest commit, c27d44a *"fix tests"*, was itself an attempt to fix this case — it added a 5 s timeout to the preceding `findByRole("combobox", …)` — and the failure survives it. CI runs on the merge ref, so the run above already includes that fix.

**What I found, for whoever picks it up.** The `Add resource` button is unconditionally rendered (`AppDataPanel.tsx:437-445`), so this is an accessibility-tree problem, not a rendering one: after the second `SelectField` (the storyboard picker) is opened and an option clicked, the button is not reachable by role. My first guess — that MUI's popover still held `aria-hidden` on the app root and a `findByRole` retry would clear it — is wrong: swapping in `findByRole` still times out. So the popover appears not to close on the option click, and the fix is upstream of the failing assertion, in the kind → document select interaction rather than at line 294.

I did not widen this PR to chase it. Happy to take it as a separate change if you want.

---

PR #5132, opened 2026-08-22
Claude/sdk trpc consolidation ut79fs

heavy-d:
### Summary

This PR implements the server-side foundation for SDK v1 contract convergence: freeze the current public contract, declare it once in a typed operation registry, publish a deterministic contract bundle on release, and route all SDK v1 HTTP, WebSocket, and tRPC traffic through a shared transport-neutral service boundary. **No public API changes** — existing `/api/sdk/v1` paths, WebSocket commands, MessagePack envelopes, auth policies, and feature flags behave the same; Phase 0 golden tests enforce byte-exact parity.

Completes Phases 0–3 of [`docs/sdk/sdk-trpc-consolidation.md`](docs/sdk/sdk-trpc-consolidation.md). Phases 4+ (C# SDK consumption, cross-repo CI, release migration) are follow-up work.

---

### Phase 0 — Freeze and measure

- Captures **20 byte-exact fixtures** (11 HTTP, 8 WebSocket) under `packages/protocol/fixtures/sdk-v1/`
- Adds golden replay tests through real Fastify plugins, `handleApiRequest`, and `UnifiedWebSocketRunner`
- Pins route/dispatcher inventories (11 Fastify registrations vs 6-route `http-api.ts` subset) and the four SDK-related tRPC procedures
- Documents baseline policies, error shapes, and drift checks in `docs/sdk/phase-0-baseline-2026-08-22.md`

### Phase 1 — Declare and publish

- Introduces typed operation registry: `sdk-v1-operations`, `sdk-v1-http-operations`, `sdk-v1-websocket-operations`
- Every public operation gets a stable ID, implementation status, schemas, auth/feature policy, and declared errors
- OpenAPI/AsyncAPI artifacts are now **generated from declarations** (committed artifacts remain byte-identical)
- Adds `build:sdk-contract-bundle` and `diff:sdk-contract` for deterministic bundle output and semantic change classification
- Release CI builds and attaches the contract bundle to GitHub Releases
- Documents compatibility rules in `docs/sdk/protocol-v1-compatibility.md`

### Phase 2 — Transport-neutral service boundary

- Adds `SdkV1Service` facade and `SdkV1ServiceError` with transport-specific mappings
- Introduces typed handler map keyed by operation IDs
- Migrates all implemented operations (discovery, lifecycle, model catalog/downloads, preflight, temporary asset upload) behind one service path
- Keeps domain services focused; multipart parsing stays at the HTTP edge

### Phase 3 — Converge adapters

- Adds declaration-driven Fastify route plugin (`routes/sdk-v1.ts`) as the single owner of SDK v1 HTTP registrations
- Removes duplicate SDK dispatch from scattered route files and `http-api.ts`
- Routes SDK WebSocket lifecycle commands and the four retained tRPC procedures through the shared boundary (replacing in-process tRPC hops where safe)
- Shadow parity tests validate old vs new paths before cutover

---

### Notable files

| Area | Key additions |
|------|---------------|
| Protocol | Operation registry, bundle/diff scripts, schema conformance tests |
| WebSocket server | `sdk-v1-service.ts`, `sdk-v1-handler-map.ts`, `routes/sdk-v1.ts` |
| Tests | `sdk-v1-http-goldens`, `sdk-v1-ws-goldens`, route/TRPC inventories, service + shadow parity tests |
| CI/Release | Contract bundle attachment in `.github/workflows/release.yaml` |
| Docs | Updated convergence plan, baseline record, compatibility rules |

**83 files changed**, ~13k insertions / ~1.7k deletions.

---

### Test plan

- [ ] `cd packages/protocol` — run protocol tests (`sdk-v1-operations`, `sdk-contract-bundle`, `sdk-contract-diff`, golden schema conformance)
- [ ] `cd packages/websocket` — run SDK v1 golden suite:
  ```bash
  npx vitest run tests/sdk-v1-http-goldens.test.ts tests/sdk-v1-ws-goldens.test.ts tests/sdk-v1-route-inventory.test.ts tests/sdk-v1-trpc-caller-inventory.test.ts tests/sdk-v1-service.test.ts tests/sdk-v1-fastify-shadow-parity.test.ts
  ```
- [ ] Verify existing SDK v1 handler tests still pass (capabilities, preflight, model catalog/download, temporary asset upload, lifecycle RPC)
- [ ] Confirm release workflow builds and attaches `nodetool-sdk-v1-contract-*.tar`
- [ ] Smoke-test a few SDK v1 HTTP endpoints and WebSocket commands against a running server (responses should match pre-PR behavior)
- [ ] Confirm feature flags still work independently: `NODETOOL_REQUIRE_SDK_AUTH_V1`, `NODETOOL_DISABLE_SDK_WORKFLOW_INTERFACE_V1`, `NODETOOL_DISABLE_SDK_LIFECYCLE_V1`

---

### Follow-ups (out of scope)

- **Phase 4:** Pin bundle in `nodetool-sdk` and generate/verify C# wire layer
- **Phase 5:** Cross-repository conformance CI as a release gate
- **Phase 6–8:** Benchmarks, C# session facade simplification, safe migration/cleanup


georgi:
Blockers
1. CI is red — merge is blocked
Quality Gate failed on this head (c216fac):
•  test-packages: packages/protocol — publishes schema bundles with resolvable component references in sdk-protocol-artifacts.test.ts timed out at 5s.
•  quality: aggregate failure from that.
Cause is almost certainly Ajv compiling the huge lifecycle $defs tree after refs were scoped under each component. The test is valuable (it is why they moved off z.fromJSONSchema), but it cannot live at the default 5s timeout.
Fix options: raise timeout for that case, compile once and reuse the Ajv instance, or compile only the $defs that goldens / operations actually reference. Do not drop the check.
2. Semantic diff can miss or under-classify contract changes
Copilot is right on both of these; they are not nits for a tool that is supposed to be the v1 compatibility gate.
Response metadata removals are invisible. In diff-sdk-contract.ts, content type / status / schema refs are compared only when both sides are present. Dropping response.content_type (or status, or schema) produces no change and exit 0.
Request $defs edits are risky, not breaking. diffSchemaDefs only special-cases response names. A required field added to PreflightRequest becomes schema-def-changed → exit 2. The compatibility doc says request schema changes are breaking (exit 3).
Also worth fixing in the same pass: the operations manifest omits HTTP parameter metadata (version=1 requiredness, path params, multipart field). Those can change with no manifest diff.
3. “Shadow parity” no longer compares old vs new
sdk-v1-fastify-shadow-parity.test.ts registers sdkV1Routes twice (production path vs /__sdk-v1-shadow). After cutover that only proves prefix rewriting. A regression in the new plugin cannot fail this test.
Either:
•  compare Fastify plugin vs the pre-cutover handleApiRequest adapters for the six legacy routes, or
•  rename it to a subpath / encoding test and stop calling it parity.
Same class of issue: HTTP goldens still label captures via: "http-api-dispatcher" and owner: routes/nodes.ts, but after Phase 3 they all go through sdk-v1.ts. The bytes can stay; the metadata should match reality or the pin is lying.
4. Standalone handleApiRequest / createHttpApiServer lost SDK ownership
Phase 3B intended to drop the second dispatcher. That is fine for Fastify (the production server). It is not free:
•  createHttpApiServer is still a public export of @nodetool-ai websocket.
•  /api/sdk/v1/* on that path is now 404 { detail: "Not found" }.
•  GET /api/workflows/:id/interface no longer has a dedicated branch, so it can fall through handleWorkflowById instead of the interface handler.
If the standalone server is supported for SDK clients (tests, test-ui-server, embeddings), this is a public behavior change despite “no public API changes.” Either keep those routes on the dispatcher via the same boundary, or document that Fastify is the only supported SDK HTTP transport and add a regression test that createHttpApiServer is not an SDK surface.
The route-inventory test currently skips the compatibility interface path when asserting dispatcher 404s, so it cannot catch the fall-through.

Should-fix before merge
Wrong-method fallbacks change Fastify 404 → handler 405 for undeclared methods.
LEGACY_METHOD_HANDLER_IDS registers DELETE/PUT/PATCH/… onto the same handler so goldens can keep {detail:"Method not allowed"}. That also means PUT /api/sdk/v1/capabilities is no longer Fastify’s 404. If the claim is byte-exact public contract, pin the methods you care about and leave the rest as 404.
Custom userIdHeader + preflight.
bridge() forwards identity as x-user-id. handleSdkPreflight reads options.userIdHeader. A non-default header yields AUTHENTICATION_REQUIRED on every Fastify preflight. Resolve principal from the Fastify request.userId the plugin already sets.
OpenAPI security is global [{ bearerAuth: [] }, {}].
Authenticated operations (downloads, preflight, upload) look optional to generated clients. Not a runtime break, but it fights the new “declaration is the contract” story. Per-operation security from auth would match the registry.
Plan doc vs this PR.
docs/sdk/sdk-trpc-consolidation.md marks Phase 4/5/7 done in nodetool-sdk. This PR is only the producer. Easy to misread as “C# consumer is in this merge.” Keep status, but say clearly what lands here.

Nits / follow-ups
•  Fixture route.owner still points at nodes.ts / workflows.ts / assets.ts.
•  HTTP vs WS node-type goldens disagree on python_bridge_ready (false vs true) because the harnesses default differently. Not introduced as a product change, but the goldens now freeze the inconsistency.
•  uploadTemporaryAsset size-checks in both the HTTP adapter and the service. Harmless; pick one owner.
•  Bundle test expects fixtures/sdk-v1-baseline.json and the new tree — good — but the 100-byte USTAR name limit will bite if fixture paths grow.
•  Release upload uses if-no-files-found: ignore then later fail_on_unmatched_files: true. Fine if the Linux build always runs; a skipped/failed bundle step will fail late in publish.

Suggested merge bar
Do not merge until:
1.  Protocol tests are green (Ajv timeout).
2.  Diff tool treats request-schema edits as breaking and response-metadata removals as a change.
3.  Shadow / inventory tests assert the story they name (or get renamed).
4.  An explicit decision on createHttpApiServer + /interface (keep on dispatcher or document + test as unsupported).
After that, this is a solid foundation for pinning the C# SDK to a bundle. The goldens and the single service path are the parts I would protect most carefully in review of follow-up PRs.

PR #5133, opened 2026-08-22
fix(validate): close three gaps in the example and workspace gates

georgi:
## What

Three gates that were reporting green over ground they never touched.

### 1. `nodetool affected` never named `reliability/`

The command scanned `packages/` only, so a change under `reliability/harness`
— a workspace declared in the root `package.json` — mapped to no workspace at
all, and the reliability journeys it runs mapped to nothing either.

Workspaces now come from the root `package.json`. A workspace can also declare
directories it owns outside its own tree (`EXTRA_WORKSPACE_PATHS`), which is
how `reliability/journeys/` reaches the harness that runs it.

```
$ nodetool affected reliability/harness/src/index.ts reliability/journeys/linear-text-pipeline/workflow.json
  dependent @nodetool-ai/cli
  changed   @nodetool-ai/reliability-harness
```

### 2. `npm run validate:examples` never scanned `examples/workflows/`

The scan covered `packages/**/examples/` only, leaving the 33 CLI samples the
repo's own README tells people to run as the shipped graphs nothing validated.
Both roots are scanned now, and an empty result from either fails the script
rather than passing quietly.

### 3. What the wider scan then found

Four of the 33 failed, all on `unknown_node` — but only two of them for a
Python-only node:

- **`nodetool.text.*` types are not Python-only.** `TrimWhitespace`,
  `RegexMatch`, `Chunk` and friends were *removed* with a
  `NODE_TYPE_MIGRATIONS` entry; `Graph.loadFromDict` rewrites them to
  `nodetool.code.Code` on every load, so those graphs run fine. The validator
  was refusing what the runner accepts. `validateGraph` now migrates the graph
  first, exactly as the runner does — so it validates what will execute.
- **Two migration snippets silently dropped properties.** `TrimWhitespace`
  ignored `trim_start`/`trim_end`, `IndexOf` ignored `case_sensitive`; a graph
  setting `trim_start: false` got both ends trimmed anyway. Both bodies now
  read them from the dynamic inputs, matching the pattern the neighbouring
  migrations already use. Surfaced by the `code_unused_input` warnings the
  migrated graphs produce.
- **The two genuinely Python-only examples** (librosa MFCC, pedalboard Reverb)
  get a named exemption for exactly those node types. Every other issue in
  those files still fails the gate, and an entry that stops firing fails it
  too, so an exemption cannot outlive the reason it was written.

The `workflow-execution` harness surface now lists `examples/workflows/`, so
`harness gate` runs the validate selfcheck on a diff that touches one.

## Proving the new checks can fail

Each was inverted once and watched go red, then restored:

| Inversion | Result |
|---|---|
| bogus node type in `examples/workflows/variables_cli.json` | `FAIL … 251/252`, exit 1 |
| extra dead name in a `PYTHON_ONLY_NODES` entry | `stale PYTHON_ONLY_NODES entry … lib.pedalboard.NoLongerUsed` |
| exemption naming a file the scan cannot reach | `PYTHON_ONLY_NODES names 1 file(s) the scan does not reach` |
| `validateGraph` without the migration pass | new node-sdk test fails |

## Verification

- `npm run validate:examples` — 252/252 (was 248/252 once the root tree was in scope)
- `npm run test:packages` — 108/108 tasks
- `npm run typecheck` — clean (web, electron, mobile)
- `npm run lint` — clean
- `npm run test:web` / `:electron` / `:mobile` — green except the
  already-red-on-main `AppDataPanel › pins a named document of the selected
  kind`, which this diff does not touch
- `nodetool harness gate --base main` — 3/3 selfchecks
- `packages/base-nodes` needed `mesa-vulkan-drivers` locally (the documented
  headless-WebGPU gap); with it installed, 2027 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01AusSZiKUaxYqaQicNwWHsz)_

### DIFF
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -216,7 +216,9 @@ jobs:
           #      commit 2c0fd8fad), hence is-test:false — they stay a required
           #      gate even when skip-tests drops the bulk test legs.
           #   3. `npm run validate:examples` — rot detector for the shipped
-          #      example workflows: `nodetool validate` against the current node
+          #      example workflows, both `packages/**/examples/` (templates and
+          #      app bundles) and `examples/workflows/` (the CLI samples):
+          #      `nodetool validate` against the current node
           #      registry (unknown node types, missing required properties,
           #      unselected models, dangling/mis-typed edges).
           #      --warnings-as-errors (set in the script) so a warning-level
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1723,6 +1723,12 @@ rebuild/test: the owning package plus its downstream dependents, and a
 `build:packages` only when a decorator package (loads from `dist/`) is affected.
 Avoids reflexively running the full 1–2 min build.
 
+Workspaces come from the root `package.json`, not from a scan of `packages/` —
+`reliability/harness` is a workspace too, and a scan of one directory reported
+every change under it as belonging to nothing. `reliability/journeys/` maps to
+the harness that runs it (`EXTRA_WORKSPACE_PATHS` in
+`packages/cli/src/affected/affected.ts`).
+
 ```bash
 npm run dev:nodetool -- affected                       # uses git working-tree changes
 npm run dev:nodetool -- affected --base main           # diff against a ref
--- a/packages/cli/src/affected/affected.ts
+++ b/packages/cli/src/affected/affected.ts
@@ -27,7 +27,24 @@ export interface PackageInfo {
   dir: string;
   /** Internal `@nodetool-ai/*` dependencies declared in package.json. */
   internalDeps: string[];
+  /** Extra repo-relative dirs this workspace owns — see EXTRA_WORKSPACE_PATHS. */
+  ownedPaths?: string[];
 }
+
+/**
+ * Directories that belong to a workspace without living inside it.
+ *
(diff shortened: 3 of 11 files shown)


georgi:
`Quality Gate / test-app` is red on one test, and it is not this PR's:

```
AppDataPanel › pins a named document of the selected kind
  web/src/components/appbuilder/__tests__/AppDataPanel.test.tsx:294
  Unable to find an accessible element with the role "button" and name /add resource/i
Tests: 1 failed, 13374 passed
```

This diff touches no frontend code — `git diff origin/main -- web/ electron/ mobile/` is empty, so the test and everything it renders are byte-identical to `main`, where the same test is already failing. I reproduced it locally on this branch with the identical assertion and message.

The sibling case one block up (`pins an asset resource`) passes, so the Resources section does open; what breaks is the kind → document select interaction — after choosing `storyboard` and an option, the "Add resource" button is no longer in the tree. The fix belongs in that interaction, not at the assertion, and not in this PR — widening a validation-gate change into an app-builder component fix would make both harder to review.

Every other check on this head is green: lint, build, deps, bundle, app-build, typecheck, test-packages, Workflow Integration Tests, Workflow Runner Browser E2E, CodeQL.

---

PR #5166, opened 2026-08-23
feat(agents): let model rankings decide an agent's default model

georgi:
## What changed

`find_model` is where an agent picks a model, and its rank term sat below the recommended bonus (≤ +80 against +100), so the hand-pinned `RECOMMENDED_MODELS` list still decided every default — `find_model("text_to_video")` answered with `openai:sora-2`, a model the rankings artifact does not rank at all, over the rank-1 route. Scoring is now an explicit ladder (`SCORE_TIERS`), each tier above the sum of everything below it: model hint 2000, provider hint 1000, `prefer_local` 500, leaderboard 200 + `normalized × 80`, recommended 100, downloaded 30. A ranked candidate does not also take the recommended bonus, or the static list could still reorder the leaderboard; preferences the caller actually stated keep winning, and the pinned list keeps ordering what the artifact says nothing about (language, embedding and ASR models, local models, anything unmatched). The answer also collapses to one row per canonical model — `gpt-image-2` is reachable through five routes, so a top-5 used to be five copies of one model; the best-scoring route survives and names the rest in `alternate_routes`. Tool descriptions, the `nodetool.models.pick` prompt text, and the design doc say what the default now is.

## Verification

- [x] `npm run test:affected` — green except `packages/base-nodes` `image-*-examples-run`, which fail on this box with `No WebGPU adapter available (Node/Dawn)`: no Vulkan ICD installed, the environment gap AGENTS.md documents, unrelated to this diff. `npm run test --workspace=packages/agents` on its own: 3349 passed, 5 skipped.
- [x] `npm run typecheck` — web and electron clean (`typecheck:web` / `typecheck:electron` exit 0). `typecheck:mobile` fails on this box because `mobile/node_modules` is not installed (`Cannot find module 'expo-file-system/legacy'`, `react-native`, …); it fails identically without this diff.
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base main` — `Gate: 8/8 selfchecks passed`

Inverted to prove the new tests bite (restored afterwards):

- `SCORE_TIERS.ranked: 200 → 0` → `× leads the hand-pinned recommended list`
- `RANK_BONUS_MAX: 80 → 0` and `ranked → 0` → 6 failures, including `× orders find_model's answer by the leaderboard`
- dropping the `bestRoutePerModel` call → `× answers with one row per model, not one per route`

## Agent capabilities

- [x] `npm run capabilities:check` passes — `find_model`'s contract fingerprint is re-synced (its description changed)
- [x] `find_model` already names `capability-suites` plus its suites in `packages/cli/src/harness/capability-table.ts`; the ordering rules are pinned in `packages/agents/tests/capabilities-models-rankings.test.ts`

## New checks

No new rule or audit — the change is covered by tests, each inverted once above.

https://claude.ai/code/session_01SwZhNgc3igs7dc5XJnXT49

---
_Generated by [Claude Code](https://claude.ai/code/session_01SwZhNgc3igs7dc5XJnXT49)_

### DIFF
--- a/docs/canonical-model-intelligence-design.md
+++ b/docs/canonical-model-intelligence-design.md
@@ -3,8 +3,9 @@
 Status: passes 1–3 and 4a are implemented (accessors + artifact in
 `packages/model-pricing`, the rankings sync + nightly workflow, the
 `find_model` rank term, the enriched `/api/models/recommended/*`
-endpoints). The picker work in pass 4 and the first real rankings
-artifact (which lands via the nightly sync PR) are still open.
+endpoints), and the first real artifact has landed — 168 routes, ranked
+across five tasks. Rankings now *lead* the agent's default model choice
+(§4 below). The picker work in pass 4 is still open.
 
 ## The problem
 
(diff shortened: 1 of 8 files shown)


georgi:
`Quality Gate / test-packages` failed on `@nodetool-ai/model-pricing`, and the failure predates this PR: `model-rankings.test.ts > the shipped artifact > is empty but valid` asserts `generatedAt` is `null` and `models` is `{}`, which stopped holding when #5164 synced the first real artifact (168 routes). Nothing in this branch touches `packages/model-pricing/src`.

I fixed it here rather than only reporting it, because it blocks every PR against `main` and it is the same staleness this PR already had to fix in the sibling `packages/agents` test. `ee3adee` pins what holds whatever the nightly sync writes — schema version and source, every key parsed as `provider:model_id` with a canonical id behind it, every route to one canonical model carrying identical `tasks` (the invariant the `find_model` rank term relies on), and the accessors answering for a route the artifact carries and with `null`/`[]` for one it does not. An empty artifact stays a valid shipped state.

Proven able to fail: perturbing the artifact — one route's rank changed, one key without a colon — reds the two structural assertions.

---

PR #5168, opened 2026-08-23
fix(web): fix iOS double-tap needed to select a model

georgi:
## Summary

- Fixes a report that the model select menu on mobile browsers needed two taps to select a model.
- Root cause: model picker rows (`web/src/components/model_menu/ModelList.tsx`) render as `<div role="button">` with matching `:hover` CSS rules (MUI's `ListItemButton` hover style, plus the app's favorite-star/pin hover reveal). iOS Safari requires a first tap to "materialize" `:hover`/`:active` on any element that isn't natively interactive (`<a href>`, `<button>`, etc.) but matches a `:hover` rule, so the real tap only registers on the second try. This is WebKit-specific and doesn't reproduce in Chromium-based emulation, which synthesizes `click` on a single tap regardless.
- Fix: bind one no-op `touchstart` listener to `document` in `MobileClassProvider` (mounted once at the app root). This is WebKit's own documented escape hatch — once any element on the page has a real touch listener bound, Safari stops requiring the double-tap materialization app-wide, fixing not just the model picker but any other list/menu with the same hover-styled-non-native-element pattern.
- Considered rendering the rows as real `<button>` elements instead, but reverted: the rows contain nested favorite/pin buttons, and nesting `<button>` inside `<button>` is invalid HTML with unpredictable touch behavior.

## Test plan

- [x] `npx tsc --noEmit` and `npx oxlint` on the changed file — clean
- [ ] Manually verify on an iOS Safari device: model picker no longer needs a double tap to select a model

---
_Generated by [Claude Code](https://claude.ai/code/session_01FXyH3tu4HFsguAT6ZDjNGy)_

### DIFF
--- a/web/src/components/MobileClassProvider.tsx
+++ b/web/src/components/MobileClassProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useLayoutEffect } from 'react';
+import React, { useEffect, useLayoutEffect } from 'react';
 import { useMediaQuery } from '@mui/material';
 import { useTheme } from '@mui/material/styles';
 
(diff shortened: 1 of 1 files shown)


georgi:
CI red on `Quality Gate / test-packages` (`@nodetool-ai/agents#test`), failing on `tests/edit-search-tools.test.ts > EditFileTool symlink containment > refuses to create through a DANGLING symlink that escapes the workspace`.

Verified this is pre-existing on `main`, not caused by this PR: this branch is `origin/main` (4176a73f) plus one commit that only touches `web/src/components/MobileClassProvider.tsx`, and the same test fails identically when run locally against 4176a73f directly (`ENOENT: no such file or directory, mkdir '.../outside'` instead of the expected "outside the workspace" message). Not fixing it here — out of scope for this mobile double-tap fix. Will keep watching CI and rebase once the base check recovers.

---

PR #5178, opened 2026-08-24
Fix the failure modes a media build session hit end to end

georgi:
## What changed

A session asked to build a short video generated three clips and then spent
the rest of its budget failing to score them, probe them, title them or cut
them — every failure a hole in this repo, not in the model. This fixes nine
of them: the FAL music catalog is 70 entries of which 22 generate music (the
rest are Kokoro voices, a dialogue model, sound effects and audio transforms,
all filed under one `text_to_audio` module name); `find_model` tagged a music
hit `tts_model`, which the property refuses and the graph validator then
mis-diagnoses against the TTS catalog; with no task named, the leaderboard
ranked a music search on the *speech* board, which has no music task at all;
model rows carried `model_id` while the ref beside them carries `id`, so two
catalogs printed as `provider/undefined`; a generated `asset://` could not
reach `ffprobe` or `download_file`; ffprobe's all-strings JSON broke the
arithmetic that followed; `generate_speech` reached past the context to
`getProvider`, so on a context without it every FAL voice reported "fal_ai
does not support textToSpeech"; there was no music capability at all and no
way to create a timeline; and `update_shot` plus `toBase64` each failed
silently. Two new capabilities close the gaps the rest exposed:
`generate_music` and `create_timeline`.

## Verification

- [x] `npm run test:affected` — green except three barrel-import "exports all
      public API" tests (`text-nodes`, `chat`, `deploy`) that time out under
      the parallel turbo run and pass standalone in 1–10s; a different one
      each run, none touched by this diff, and `deploy`'s also times out on a
      clean tree under the same load. The three packages this diff touches
      pass in full: `agents` 3393, `runtime` 2954, `cli` 660.
- [x] `npm run typecheck` — `typecheck:web` and `typecheck:electron` pass;
      `typecheck:mobile` fails on `Cannot find module 'react-native'` because
      `mobile/node_modules` is not installed in this container (mobile is not
      a root workspace, and this diff touches no mobile file).
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base main` — `Gate: 6/6
      selfchecks passed`

Image-node tests need a Vulkan ICD; run with `VK_DRIVER_FILES` pointed at
lavapipe as `AGENTS.md` describes, or they fail on "No WebGPU adapter
available".

## Agent capabilities

- [x] `npm run capabilities:check` passes — `capability table is current (219
      capabilities)`
- [x] `generate_music` and `create_timeline` each name a suite in
      `packages/cli/src/harness/capability-table.ts`
      (`capabilities-media.test.ts`, `capabilities-timelines.test.ts`), and
      the re-declared contracts of `ffprobe` and `download_file` name the two
      new suites written for them

## New checks

Every check added here was inverted once and observed failing:

| Reverted | Test that went red |
|---|---|
| `text_to_music` ref type; default rank task | `hands back a music_model ref for text_to_music`, `does not score a music search on a speech leaderboard` |
| music evidence rule | 4 in `manifest-models.test.ts`, incl. `keeps speech, sound effects and audio transforms out of the FAL music catalog` |
| ffprobe `inputs` staging | `ffprobe stages an inputs ref the way ffmpeg does`, `…refuses an inputs name that escapes the workspace` |
| `download_file` stored refs | 3 in `capabilities-web-download.test.ts` |
| encoded-TTS error kept | `reports why the encoded path gave up, not the streaming stub's message` |
| `create_timeline` | `creates an empty sequence the caller can then edit` |
| `update_shot` field check | `refuses a shot field it does not set, instead of dropping it` |
| `toBase64` guard | `names an un-awaited Promise instead of encoding it to nothing` |
| `id` alias | `reports each row's id under both names`, `lists every configured provider's models, sorted` |

The music reclassification was measured against the three shipped manifests
before and after rather than asserted from reading: FAL music 70 → 22 with no
real music model lost, FAL TTS 39 → 52, replicate TTS 14 → 25, kie unchanged.

https://claude.ai/code/session_01NowaRUvQwvdHDnT3Wpis6g

---
_Generated by [Claude Code](https://claude.ai/code/session_01NowaRUvQwvdHDnT3Wpis6g)_


georgi:
Reviewed the full diff (2096 lines, 24 files). Nine real bugs from one live session, each with a clear root cause and a test that was inverted to prove it can fail.

Key fixes verified:

1. **FAL music catalog**: `moduleName: "text_to_audio"` no longer counts as music evidence — FAL files speech, music, and sound effects alike under it. New classifiers plus positive music-keyword evidence. Pinned against the real FAL manifest (70→22 music, 39→52 TTS).
2. **`text_to_music` → `music_model`** ref type fix
3. **`id` alias** on model results to prevent `undefined`
4. **Leaderboard task default** prevents scoring music on the speech board
5. **`encodedSpeech` helper** tries `context.textToSpeechEncoded` before `getProvider`; error now reports the real reason
6. **`generate_music`** and **`create_timeline`** capabilities fill real gaps
7. **ffprobe `inputs` staging + numeric `summary`**
8. **`download_file` stored refs** (`asset://`, `data:`, `/api/storage/`)
9. **`update_shot` field check** and **`toBase64` guard** — silent drops become explicit errors

No cross-PR conflicts. CI still running — will merge once green.

---

Issue #5184, opened 2026-08-24
Gated HuggingFace models cannot be downloaded on a rented worker


georgi:
## What happens

Downloading a gated repo on a RunPod worker fails:

```
Unauthorized to access 'https://huggingface.co/briaai/RMBG-2.0/resolve/main/.gitattributes'.
Status: 401. Check your Hugging Face token and permissions. Token present: False
```

`Token present: False` is the whole story — the worker has no token and no way to be given one.

## Why

The worker resolves a token in `nodetool.integrations.huggingface.hf_auth.get_hf_token`: a per-user secret store first, then `os.environ["HF_TOKEN"]`. On a rented pod there is no secret store, so it comes down to the pod's environment.

Nothing puts it there. `RunpodPodProvider.provision` (`packages/compute/src/providers/runpod.ts:73`) sends:

```ts
env: { ...spec.env, HF_HOME: WORKER_HF_HOME },
```

`spec.env` comes from the profile's `spec` blob via `specFromProfile`, and nothing populates it with `HF_TOKEN` — not the manager, not the Workers settings panel, and the CLI has no `--env` flag either (same family as the missing `--disk`, see #5182).

So a worker can only ever download public repos.

## Impact

The gated set is most of the interesting models: FLUX.1-dev, FLUX.1-Fill-dev, Qwen-Image, RMBG-2.0, Stable Diffusion 2.1. A user who has stored `HF_TOKEN` locally and can download these on their own machine cannot download them on a worker, and the failure only appears after provisioning a GPU.

## Reproduce

With `HF_TOKEN` in the local secret store, on any RunPod worker:

```
models.download { repo_id: "briaai/RMBG-2.0", model_type: "hf.image_segmentation" }
→ 401, Token present: False
```

## Notes toward a fix

Passing the token as a pod env var puts a long-lived credential in the provider's API request and in the pod's environment, where anything on the box can read it. Worth weighing against forwarding it per-request over the bridge instead — `ModelDownloadRequest` has no token field today, and the worker would then hold it only for the life of the call.

Found while testing HuggingFace nodes on a worker; see also the missing-`refs` bug (nodetool-core#1007) and the idle-heartbeat gap filed alongside this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0138g6jBUaPu6E7bxxYPVGb6

georgi:
[...]

PR #5187, opened 2026-08-24
feat(models): send an HF token with a worker model download

georgi:
## What

`models.download` now carries an optional HuggingFace token, and the two worker-bound call sites resolve one and send it.

TypeScript half of #5184. Worker half: nodetool-ai/nodetool-core#1008 (already open) — this PR sends the field that PR taught the worker to read.

## Why

A rented GPU worker is a bare container. It has no per-user secret store and nothing puts `HF_TOKEN` in its environment, so `get_hf_token()` there resolves to nothing and every gated repo answers:

```
Unauthorized to access 'https://huggingface.co/briaai/RMBG-2.0/resolve/main/.gitattributes'.
Status: 401. Check your Hugging Face token and permissions. Token present: False
```

`Token present: False` is the whole diagnosis. The gated set is FLUX.1-dev, FLUX.1-Fill-dev, Qwen-Image, RMBG-2.0 and Stable Diffusion 2.1 — most of what a GPU gets rented for. A user who can download these on their own machine could not download them onto a worker.

The alternative — baking `HF_TOKEN` into the pod's environment at provision time — puts a long-lived credential in the provider's API request and leaves it readable by anything on the box for the pod's whole life. Per request, the worker holds it for the length of one call.

## What changes

- `ModelDownloadRequest.token?: string | null` (`packages/runtime/src/python-bridge-types.ts`).
- `downloadModel` drops a blank token instead of forwarding it. `Authorization: Bearer ` fails differently than sending no header at all, and worse for a public repo. One implementation, so every caller gets the rule and it matches the worker's own `_request_token`.
- `resolveWorkerHfToken` (`packages/huggingface/src/hf-auth.ts`): the user's stored `HF_TOKEN` first, then this process's environment and HF token file via the existing `getHfToken`. Blank at either step is not a credential. A secret store that throws — locked, absent — falls through to the environment rather than failing the download. It takes the secret reader as a parameter, so `@nodetool-ai/huggingface` gains no dependency on `@nodetool-ai/models`.
- Server: `relayWorkerDownload` resolves the token for the requesting user. It is the single server-side worker download path — both the `/ws/download` socket and the SDK download service go through it — so `startWorkerDownload` grew a `userId` first parameter to match `startHuggingFaceDownload`, and the socket handler passes `req.userId ?? "1"`.
- CLI: `worker models download` resolves for the local user (`"1"`).

## Not the schema

`packages/protocol/src/bridge-frames.ts` models only the six **inbound** frame types the dispatcher switches on (`discover`, `result`, `error`, `chunk`, `progress`, `comfy.event`). Outbound requests like `models.download` are deliberately not modeled — `getBridgeFrameSchema` returns undefined for them — so there was nothing to widen and no unknown-field rejection to clear. Verified by driving a real download through the fake worker, which validates every frame it receives.

## Credential handling

- Resolved on the host, never read off `msg` — that JSON comes from the client over `/ws/download`. A test pins that a client-supplied `token` is not relayed onward.
- Not logged. The bridge logs no request payload on this path (`python-bridge-base.ts` logs a dispatched `execute`'s node type, a malformed inbound frame's type/id, and a provider call's error message; the WebSocket transport logs URLs and socket errors). Nothing serializes the download request.
- Not echoed into a progress or error frame; a test asserts the token appears in no update the worker sends back.
- Not persisted: the SDK service retains `repo_id`/`path`/`model_type`, and `token` is not part of `SdkV1ModelDownloadStartRequest` — it is attached inside the relay.

## Blast radius

Additive on the wire. An install with no token resolvable sends no field and behaves exactly as before; an old worker ignores the field; a new worker with an old host takes its own fallback. The local (non-worker) download path is untouched — that process already resolves its own credential, so forwarding one buys nothing.

`nodetool models download-hf` is local-only (`DownloadManager` against the local cache) and is deliberately not changed.

## What is proven, and what is not

Reported by a maintainer, on a RunPod A40 with `HF_TOKEN` in the pod environment:

```
Unauthorized to access 'https://huggingface.co/briaai/RMBG-2.0/resolve/main/.gitattributes'.
Status: 403. Check your Hugging Face token and permissions. Token present: True
```

Against the same request without a token — 401, `Token present: False` — that is
the mechanism this PR changes: a token supplied by the host arrives where the
download happens. The download still failed, because that credential is not
authorized for the repo's contents; the same token gets 403 from a developer
machine while `/api/models/briaai/RMBG-2.0` returns 200 with `gated = auto`.

**A gated repo completing a download on a worker is unproven.** It needs a
credential authorized for that repo, and it is worth pinning with one once this
lands.

## One uncovered line

The CLI call site in `packages/cli/src/commands/worker-models.ts` has no test.
`packages/cli`'s vitest config replaces `@nodetool-ai/runtime`,
`@nodetool-ai/models` and `@nodetool-ai/config` with empty stub modules, so a
test there can build neither a real bridge nor a real `getSecret`. The four
lines mirror the server relay, which is covered.

## Verification

Reverting each part turns its tests red.

`downloadModel`'s blank-token normalization removed:

```
 × omits the field entirely when no token is available
 × treats an empty string as absent rather than sending a blank credential
 × treats a whitespace-only token as absent
 × treats null as absent
 × trims a padded token before sending it
AssertionError: expected { repo_id: 'org/gated', token: '' } to not have property "token"
      Tests  5 failed | 2 passed (7)
```

The relay no longer passing the token:

```
 FAIL  tests/models-api-worker.test.ts > relayWorkerDownload HF token > forwards the token this server can resolve
AssertionError: expected undefined to be 'hf_server_token'
      Tests  1 failed | 19 passed (20)
```

The resolver's secret-store branch disabled:

```
 × prefers the user's stored secret over the environment
 × trims surrounding whitespace off a stored token
AssertionError: expected 'hf_from_env' to be 'hf_from_secret_store'
```

The runtime tests read the payload the fake worker actually received, so they measure the wire rather than the call.

One trap the tests had to handle: both layers memoize. `getSecret` caches an env-sourced value per user and `getHfToken` caches its env/token-file read, so without clearing both a later case reads an earlier one's token — which is how the "no token available" case first passed for the wrong reason.

`npm run test:affected`, `npm run typecheck` and `npm run lint` all pass.

https://claude.ai/code/session_0138g6jBUaPu6E7bxxYPVGb6

### DIFF
--- a/packages/cli/src/commands/worker-models.ts
+++ b/packages/cli/src/commands/worker-models.ts
@@ -14,6 +14,9 @@ import {
   WebsocketPythonBridge,
   type ModelDownloadUpdate
 } from "@nodetool-ai/runtime";
+import { resolveWorkerHfToken } from "@nodetool-ai/huggingface";
+import { getSecret } from "@nodetool-ai/models";
+
 import { printTable, asJson } from "./output.js";
 
 // ---------------------------------------------------------------------------
@@ -31,6 +34,9 @@ function formatBytes(n: number | null | undefined): string {
   }
   return `${value.toFixed(1)} ${units[i]}`;
 }
+
+/** Local single-user id used across the TS stack for secret-store reads. */
+const LOCAL_USER_ID = "1";
 
 function collectOption(value: string, previous: string[] | undefined): string[] {
   return previous ? [...previous, value] : [value];
@@ -205,12 +211,19 @@ function registerDownload(
             }
           };
 
+          // The worker holds no HuggingFace credential, so a gated repo
+          // (FLUX.1-dev, RMBG-2.0, ...) 401s unless this host supplies one.
+          const token = await resolveWorkerHfToken((key) =>
+            getSecret(key, LOCAL_USER_ID)
+          );
+
           await bridge.downloadModel(
             {
               repo_id: opts.repoId,
               path: opts.filePath ?? null,
               allow_patterns: opts.allowPatterns ?? null,
-              ignore_patterns: opts.ignorePatterns ?? null
+              ignore_patterns: opts.ignorePatterns ?? null,
+              token
             },
             render
           );
--- a/packages/huggingface/src/hf-auth.ts
+++ b/packages/huggingface/src/hf-auth.ts
@@ -154,3 +154,47 @@ export async function resolveHfToken(
   if (disableImplicit) return null;
   return cached;
 }
+
+// ---------------------------------------------------------------------------
+// Token for a remote worker
+// ---------------------------------------------------------------------------
+
+/**
+ * Reads one credential out of NodeTool's secret store. Injected rather than
+ * imported so this package keeps no dependency on `@nodetool-ai/models`.
+ */
+export type SecretReader = (
+  key: string
+) => Promise<string | null | undefined> | string | null | undefined;
+
+/**
+ * Resolve the HuggingFace token a host should forward to a remote worker.
+ *
+ * A rented worker has neither a secret store nor `HF_TOKEN` in its
+ * environment, so it can only download public repos. The host resolves the
+ * credential instead and sends it with the `models.download` request
+ * (nodetool-ai/nodetool#5184).
+ *
+ * Order: the user's stored `HF_TOKEN` first, then this process's own
+ * environment and HF token file via {@link getHfToken} — the same token the
+ * local download path already uses. A blank value at either step is not a
+ * credential and yields `undefined`, so the worker takes its own fallback
+ * rather than receiving an empty Bearer header. A secret store that throws
+ * (locked, absent) falls through to the environment instead of failing the
+ * download.
+ */
+export async function resolveWorkerHfToken(
+  readSecret?: SecretReader
+): Promise<string | undefined> {
+  if (readSecret) {
+    try {
+      const stored = await readSecret("HF_TOKEN");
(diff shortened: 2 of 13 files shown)


georgi:
Correction to one line in the PR body, and the error is mine, not the author's.

The body says:

> The premise is confirmed rather than assumed: a pod launched with `HF_TOKEN` in its environment downloaded the same gated repo successfully.

That did not happen. I asserted it to the author before I had run the test, and they reasonably took it as evidence. What the test actually showed, on a RunPod A40 with `HF_TOKEN` injected into the pod environment:

```
Unauthorized to access 'https://huggingface.co/briaai/RMBG-2.0/resolve/main/.gitattributes'.
Status: 403. Check your Hugging Face token and permissions. Token present: True
```

401 → 403, and `Token present: False` → `True`. So the token **reaches the download path on the worker**, which is the mechanism this PR changes, and that much is genuinely confirmed. The download still failed, because the credential I used is not authorized for that repo's contents — from my own machine, same token:

```
GET /api/models/briaai/RMBG-2.0        -> 200   (gated = auto)
GET /briaai/RMBG-2.0/resolve/main/...  -> 403
```

So the accurate claim is narrower: **a token supplied by the host arrives where the download happens.** A gated repo completing a download on a worker is still unproven, and needs a credential authorized for the repo. Worth pinning with one the CI account can actually read, once this lands.

The same correction is on issue #5184.

PR #5273, opened 2026-08-26
feat(web): untie the workspace explorer from the open workflow

georgi:
## What changed

The Workspace Explorer showed files only while a workflow was open ("Open a workflow to access its workspace files"), reset its folder to the active workflow's workspace on every tab switch, and patched-and-saved the workflow when the user picked a different folder to browse. Browsing is a viewing choice, and the panel mounts under `/workspace` where there may be no workflow at all. The tree now reads its own persisted selection — `WorkspaceExplorerStore` resolved by `useWorkspaceExplorer`, falling back to the default workspace the server creates while answering the list call — so it works with no workflow open, keeps the folder the user picked across workflow switches, and writes nothing to a workflow. `useCurrentWorkspace` is unchanged and still answers where the *active workflow's run* writes, for the composer chip, the workflow form, and the save-destination hint. The loading branch moved ahead of the empty branch so the panel shows its skeleton while the workspace list is in flight instead of flashing "No workspace selected"; `themeMock` gains the `Skeleton` palette slot MUI derives only in a real CSS-variables palette, which is what let that branch render under Jest.

## Verification

- [x] `npm run test:affected -- --base HEAD` — 264 suites, 2294 passed, 1 skipped. (Plain `test:affected` diffs against a stale local `origin/main` here and selects all four trees; that run additionally fails `@nodetool-ai/image-nodes` on `vkCreateInstance failed with VK_ERROR_INCOMPATIBLE_DRIVER` — the missing Vulkan ICD documented in AGENTS.md, in a package this diff does not touch.)
- [x] `npm run typecheck` — `web` and `electron` clean (0 errors under `web/`). `mobile` fails on absent Expo/React Native deps, which are not installed in this container and are unrelated to the diff.
- [x] `npm run nodetool -- harness gate --base HEAD~1` — 1 surface touched (`web-editor`), 0 selfchecks to run.

Both new `WorkspaceTree` tests were run against the pre-change component (`git stash push -- web/src/components/workspaces/WorkspaceTree.tsx`) and both went red, so they pin the behavior rather than passing vacuously:

```
--- original WorkspaceTree ---
    ✕ lists the default workspace's files with no workflow open (50 ms)
    ✕ browses the workspace the user picked, not a workflow's (4 ms)
Tests:       2 failed, 2 total
--- restored ---
    ✓ lists the default workspace's files with no workflow open (178 ms)
    ✓ browses the workspace the user picked, not a workflow's (312 ms)
Tests:       2 passed, 2 total
```

## Agent capabilities

## New checks

No check, rule, or audit added — the two new suites are behavior tests, inverted above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DTSKz7vB1AjZP8TYspXEGi)_

### DIFF
--- a/web/src/__mocks__/themeMock.ts
+++ b/web/src/__mocks__/themeMock.ts
@@ -91,6 +91,7 @@ interface MockVarsPalette {
   FilledInput?: { bg: string; hoverBg: string; disabledBg: string };
   TableCell?: { border: string };
   Tooltip?: { bg: string };
+  Skeleton?: { bg: string };
   OutlinedInput?: { border: string; hoverBorder: string };
   Slider?: Record<string, string>;
   Chip?: {
@@ -479,6 +480,9 @@ draft.vars.palette.TableCell = {
 };
 draft.vars.palette.Tooltip = {
   bg: "rgba(97, 97, 97, 0.92)",
+};
+draft.vars.palette.Skeleton = {
+  bg: "rgba(255, 255, 255, 0.11)",
 };
 draft.vars.palette.OutlinedInput = {
   border: "rgba(255, 255, 255, 0.23)",
--- a/web/src/components/workspaces/WorkspaceTree.tsx
+++ b/web/src/components/workspaces/WorkspaceTree.tsx
@@ -1,7 +1,6 @@
 /** @jsxImportSource @emotion/react */
 import { css } from "@emotion/react";
 import React, { memo, useCallback, useEffect, useRef, useState } from "react";
-import { useShallow } from "zustand/react/shallow";
 import isEqual from "../../utils/isEqual";
 import { useQuery } from "@tanstack/react-query";
 import { FileInfo } from "../../stores/ApiTypes";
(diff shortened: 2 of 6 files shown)


georgi:
Auto-merge is armed but this PR cannot merge, and the reason is not in the diff.

The **Test** workflow run for head `8bad6e6` ([run 32985460348](https://github.com/nodetool-ai/nodetool/actions/runs/32985460348)) ended in `startup_failure` at 15:33 UTC. It produced **zero jobs** — the run died before any job body started — so the required **Quality Gate** check never reported, and the PR sits at `mergeable_state: blocked` indefinitely.

Evidence it is not this PR's:

- The diff touches only `web/src/**`. It changes no workflow file, no action, no lockfile.
- Across the last 30 `test.yml` runs, this is the **only** `startup_failure`. Runs #13188–#13193 on other PRs, all after mine, ran the same `test.yml` and succeeded; #13186 before it ran too.
- `list_workflow_jobs` on the run returns `total_count: 0`, so nothing in the repo was executed and nothing in the repo can explain it.

Everything that did run is green: **Page Load Smoke** / "Load every route without errors" passed. The three CodeQL `Analyze` jobs have been `queued` — never started — since 15:33; that is runner availability, not a failure.

I tried the one re-run this warrants (`POST /actions/runs/32985460348/rerun`) and GitHub refused it: `403 This workflow run cannot be retried`. GitHub does not allow retrying a `startup_failure`. `test.yml` has no `workflow_dispatch` trigger, so there is no dispatch path either.

That leaves only a new `pull_request` event to re-fire the gate: a real push to the branch, or closing and reopening the PR. I am not going to push a no-op commit or close/reopen to kick CI. There is no test to make robust here either — nothing ran.

So this needs one of: a push carrying a real change, or a close-and-reopen by a maintainer. I will keep watching the PR and will act on anything that goes red once the gate actually runs.

---

PR #5277, opened 2026-08-26
feat(atlascloud): ship the 27 newest AtlasCloud image and video models

georgi:
## What changed

AtlasCloud's catalog ranks 28 image/video models above everything the manifest carried (priority 90035–90051 against a ceiling of 90032); this ships 27 of them as nodes — Wan 3.0 and Wan 3.0 Prime, Seedance 2.5, MiniMax H3, Grok Imagine Image 2.0, Qwen Image 3.0 and 3.0 Pro, Reve 2.1, and Youchuan v8.2. Every field is generated from the model's own published `Input` schema, so `sync-atlascloud-manifest.mjs --check` reports no writable drift on any new entry. Two of them needed a mechanism the factory lacked: Wan 3.0 and MiniMax H3 reference-to-video take one mixed `refers: [{url, type}]` array rather than per-kind lists, so a field can now declare `wrapInto` and the manifest splits that array into typed `reference_images` / `reference_videos` / `reference_audios` inputs — the editor keeps its asset wiring, and the media kind each input declares is exactly the `type` the objects require. The runtime provider reads the same flag, so its generic `imageToVideo` posts `refers` rather than a `reference_images` the endpoint would reject. Numeric option lists are typed `int`, not `enum`: NodeTool serializes an `enum` prop as a string and AtlasCloud rejects `"5"` where it wants `5`, which is how the shipped Seedance and Youchuan entries already type `duration`, `quality` and `n`. AtlasCloud's chat models need no change — the provider discovers them from `GET /v1/models`.

`bytedance/seedream-v5.0-pro/layer-decomposition` is the 28th and is deliberately left out: `pickOutputUrl` returns the first output, and that model's point is the *set* of layers it returns, so the node would silently drop what the call was for. A model returning N interchangeable variants (`n`, `num_images`) is fine; one whose outputs are not interchangeable is not. `packages/atlascloud-nodes/AGENTS.md` records both that rule and the `wrapInto` convention.

Two things found and **not** fixed here, since neither is this change's scope:

- 27 writable drifts against upstream schemas on models already shipped — `nano-banana-2` / `-pro` / `-2-lite` gained `seed`/`top_p`/`temperature`, `seedream-v5.0-pro` gained `prompt_optimization_mode` and `background`, and `happyhorse-1.1`'s `resolution` enum is cased `"1080P"` where upstream now says `"1080p"`. The last one looks like a live bug. `node scripts/sync-atlascloud-manifest.mjs` applies all of them; it wants its own review.
- The catalog carries 376 further models we do not ship, mostly older generations.

## Verification

- [x] `npm run test:affected` — green. `web` 1174/1174 suites and 13476 tests passed; `electron` 63/63 suites and 684/684 tests passed. Two failures needed environment fixes rather than code: `document-nodes` / `data-nodes` failed on a missing `dist/` in a fresh checkout (fixed by `npm run build:packages`), and `image-nodes` failed with `VK_ERROR_INCOMPATIBLE_DRIVER` — the documented headless-WebGPU gap, which passes 24/24 files and 231/231 tests once `mesa-vulkan-drivers` (lavapipe) is installed, as CI already does. One flake: `packages/cli` `run-dsl.test.ts` timed out at 30s under the parallel load (38s observed) and passes 10/10 in isolation.
- [x] `npm run typecheck` — exit 0
- [x] `npm run lint` — exit 0 (pre-existing warnings only, none in the touched files)
- [x] `npm run dev:nodetool -- harness gate --base main` — run as `--base HEAD~1` against the commit; `Gate: 4/4 selfchecks passed` (5 Ring 0 reliability journeys, provider contract probes, and 2 more).

Beyond the gate: all 27 new classes instantiate from the built manifest and register their 186 properties; `sync-atlascloud-manifest.mjs --check` reports no writable drift on any new entry.

## Agent capabilities

## New checks

Three tests were added and each was inverted once and observed failing:

- `merges wrapInto asset fields into one {url, type} array` — with `appendWrapped` tagging every entry `"image"` instead of the field's kind: `Test Files 1 failed | 1 passed (2)`, `Tests 1 failed | 80 passed (81)`.
- `sends one 'refers' array on a model whose manifest wraps its references` — with the provider's `wrapInto` lookup forced to `undefined`: `Test Files 1 failed (1)`, `Tests 1 failed | 27 passed (28)`.
- `only wraps asset fields, and never collides with a declared field name` is a manifest invariant that asserts it found its targets (`expect(wrapped.length).toBe(9)`), so it cannot pass by matching nothing.

https://claude.ai/code/session_01Uz2G7JafBHqXSMQeQLQFi1

### DIFF
--- a/packages/atlascloud-nodes/AGENTS.md
+++ b/packages/atlascloud-nodes/AGENTS.md
@@ -27,8 +27,23 @@ Package specifics from shipped fixes:
   for extension-less / `asset://` URIs.
 - **Model multi-image inputs as `list[image]`**, not a single `image` with
   `array: true`.
+- **A mixed reference array is `wrapInto`, not one untyped list.** Wan 3.0 and
+  MiniMax H3 reference-to-video take a single `refers: [{url, type}]` covering
+  images, videos and audio. The manifest splits that into typed
+  `reference_images` / `reference_videos` / `reference_audios` inputs, each
+  carrying `"wrapInto": "refers"`; the factory appends their resolved URLs into
+  one array in field order and tags each entry with the kind its input
+  declared. The wire name is never a node property — the runtime provider reads
+  the same flag so its generic `imageToVideo` posts `refers`, not
+  `reference_images`. Seedance 2.5 keeps the per-kind names because its API
+  really does take three arrays.
 - **Don't expose an API option that yields an extra output** the single-output
   node can't surface (the `return_last_frame` Seedance option was dropped).
+  The same rule keeps whole models out: `pickOutputUrl` returns the first
+  output, so `bytedance/seedream-v5.0-pro/layer-decomposition` — whose point is
+  the *set* of layers it returns — is deliberately not shipped. A model that
+  returns N interchangeable variants (`n`, `num_images`) is fine; one whose
+  outputs are not interchangeable is not.
 - **The manifest is generated, not hand-edited.** `node
   scripts/sync-atlascloud-manifest.mjs` reconciles every entry's fields against
   the `Input` schema AtlasCloud publishes for that model (reachable from the
--- a/packages/atlascloud-nodes/README.md
+++ b/packages/atlascloud-nodes/README.md
@@ -3,8 +3,9 @@
 AtlasCloud.ai nodes for [NodeTool](https://nodetool.ai).
 
 Run [AtlasCloud](https://www.atlascloud.ai) image and video models inside NodeTool
-workflows — GPT Image, Nano Banana, Seedream, Qwen Image and Wan for images;
-Seedance, Veo, Kling, Wan, HappyHorse and Grok Imagine for video.
+workflows — GPT Image, Nano Banana, Seedream, Qwen Image, Reve, Youchuan and Wan
+for images; Seedance, Veo, Kling, Wan, MiniMax, HappyHorse and Grok Imagine for
+video.
 
 Node fields are generated from AtlasCloud's published model schemas. Re-sync them
 with `node scripts/sync-atlascloud-manifest.mjs` (or `--check` to detect drift).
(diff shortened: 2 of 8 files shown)


georgi:
`Quality Gate (npm run check) / test-packages` went red on 3730e63. The failing workspace is `@nodetool-ai/reliability-harness`, and the single failure is a **setup hook**, not a test body:

```
FAIL tests/journeys/client-reconnect-mid-run.test.ts > journey 9: client reconnect mid-run (task D2)
Error: Hook timed out in 10000ms.
 ❯ tests/journeys/client-reconnect-mid-run.test.ts:156:3
   156|   beforeEach(async () => {
   157|     srv = createTestUiServer({ host: "127.0.0.1", port: 0 });
   158|     await srv.listen();
```

That `beforeEach` binds a Fastify test server and a `WsFaultProxy` on loopback. It never reached the test body — everything else in the workspace passed (26/27 files, 155 tests).

**Why this is not this PR's failure.** The diff is the AtlasCloud node manifest, the AtlasCloud node factory, and `AtlasCloudProvider.imageToVideo`. This journey exercises client reconnect against the ws-server and touches none of it — the only `atlascloud` string anywhere under `reliability/` is a stale `tsbuildinfo` build artifact. The same workspace passes locally on this commit: 27 files, 156 passed, 5 skipped.

**The likely cause is a config gap, not the diff.** `reliability/harness/vitest.config.ts` raises `testTimeout` to 30 s but leaves `hookTimeout` at vitest's 10 s default, so a `beforeEach` that binds two listeners gets a third of the budget the test body gets. Three sibling packages already raise it for the same reason (`image-nodes` 30 s, `automation-nodes` 60 s, `sandbox-compiler` 60 s). The job ran 216 s wall with the suites in parallel, which is where a 10 s bind budget runs out.

No fix ported: adding `hookTimeout` to the harness config would widen this PR into an unrelated workspace, and the one-line change wants its own review. Re-running the failed jobs once now — a hook that died before any test body ran is the case a re-run is meant to confirm. If it comes back red, it is real and I will root-cause it rather than re-run again.

---

PR #5278, opened 2026-08-26
feat(kie): add newest kie.ai models and fix non-media parameter typing

georgi:
## What changed

Regenerates the KIE configs and manifest from live `docs.kie.ai`, taking the node count from 151 to 163. The twelve new models are Kling 3.0 Omni (text-to-video, image-to-video, reference-to-video, transformation), Grok Imagine Image 2.0 (text-to-image, image-edit, segment map, segment edit), Bytedance Seedance 2.5, Wan 3.0 Video and Video Prime, and Seedream 5.0 Pro Layer Decomposition. Five of them exposed parameters the generator could not type, so four parser rules are fixed in the same change. The largest: a non-media array was typed `list[image]`, which gave the node an asset handle with no upload config — and `kie-factory.ts` skips asset-typed fields, so the parameter never reached the request at all. Kling 3.0 Omni's `multi_prompt` and `elements` are object arrays and Grok's `mask_indexs` is an integer array; every non-media array is now typed off its item schema (`list[dict]`, `list[int]`, `list[float]`, `list[str]`) and sent by value, which also repairs nine existing nodes including Kling 3.0 and the Gemini/ElevenLabs TTS speaker lists. The other three: Wan 3.0 declares every media parameter as one empty `type: object` YAML anchor, so `first_frame_url` came out a plain string — the `_url` suffix now wins over the declared type, and a first/last frame is named an image rather than letting the description sniff read "frame of the video" as a video handle; `reference_file_urls` (documents) and `reference_link_urls` (web pages) stay `list[str]` instead of being read as video lists off the same "…-to-video generation" phrasing; and a string array named for a medium without the `_urls` suffix is a media list, which is what HappyHorse's `reference_image` (1–9 image URLs, previously no upload config at all) needed. The `llms.txt` snapshot is refreshed and four docs fixtures are added, one per rule, so the drift gate covers them.

## Verification

- [x] `npm run test:affected` — packages 110/110 tasks, web 13476 passed / 1174 suites, electron 684 passed / 63 suites. The mobile leg needed `npm --prefix mobile install` first (its `node_modules` was empty in this environment); after that, 1026 passed / 75 suites. The `packages/image-nodes` suite needed `apt-get install -y mesa-vulkan-drivers` per the WebGPU note in AGENTS.md; with lavapipe present it is 231 passed / 24 suites.
- [x] `npm run typecheck` — web and electron clean. Mobile reports 36 errors (`sync_mode`, `TRPC_MAX_BATCH_SIZE`, `application_id`); the identical 36 appear on the base commit with this diff stashed, and no mobile file is touched here.
- [x] `npm run dev:nodetool -- harness gate --base main` — ran as `harness gate --base origin/main`: 1/1 selfchecks passed (`provider-codegen`: fal 5 outputs match, kie 7 outputs match, both `--strict`).

Each new parser rule was inverted once and observed failing:

```
$ # array item typing reverted to `type = "list[image]"`
     × types a non-media array by its item schema
Received: "list[image]"
 ❯ tests/schema-parser.test.ts:711:36
     expect(typeOf("multi_prompt")).toBe("list[dict]");

$ # the object-typed `_url`, file/link, and media-name-array rules reverted
     × reads a URL parameter KIE declares as a bare object
     × keeps document and web-page URL lists as plain strings
     × maps a media-named string array onto an AssetRef list
      Tests  3 failed | 1 passed | 8 skipped (12)

$ # VALUE_LIST_TYPES narrowed back to list[str] only
     × sends object and number lists by value instead of stringifying them
     × omits an empty value list rather than sending an empty array
     × parses a JSON string wired into a list[dict] parameter
     × exposes value lists as input handles
      Tests  4 failed | 10 skipped (14)
```

All 163 manifest entries were loaded through `loadKieNodesFromManifest` to confirm the new nodes register with the handles the fix gives them, e.g. `kie.video.Kling30OmniTextToVideo | inputs: prompt,multi_prompt,elements` and `kie.video.Wan30Video | inputs: prompt,first_frame,last_frame,reference_images,reference_videos,reference_audios,reference_file_urls,reference_link_urls`.

## Agent capabilities

## New checks

No new check, rule, or audit — the four added tests pin parser and factory behavior rather than gating a diff. Each was inverted and observed failing; the output is in **Verification** above.

## Known gap

54 nodes already carried a description whose trailing blurb is a Markdown heading (`## Create Task`, `## Query Task Status`) because that is the first line KIE's `llms.txt` gives as the page summary. The new models inherit it, taking the count to 65. It is cosmetic, predates this change, and is left alone rather than churning 65 descriptions here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FmS3CJJiTUKAqS8rfoeH5R)_


georgi:
The **Quality Gate** required check has never dispatched on this PR, on either head. That is what is holding `mergeable_state` at `blocked` — nothing here is failing.

What did run on `3ac24d53`: CodeQL only (`Analyze (actions)`, `Analyze (python)`, `Analyze (javascript-typescript)`, all green). `test.yml` — whose `Quality Gate` job is the required check — produced no run at all. `list_workflow_runs` for `test.yml` filtered to `claude/kie-ai-newest-models-ddsmvx` returns `total_count: 0`, while PRs opened at 16:21 and 16:31 both dispatched theirs within seconds.

What I tried:

1. The branch was cut from `6697b2dc` while the base is `b0fb4bca`, so it was three commits behind. Merged `main` in (`1a10d243`) — clean, the three commits are tutorial docs and demo casts with no overlap on `packages/kie-*`. Re-validated the merged head before pushing: `generate:kie:check` 7 outputs match, `kie-codegen` 82 passed, `kie-nodes` 62 passed, `kie-codegen` lint clean. CodeQL re-dispatched on the new head; `test.yml` still did not.
2. `Provider Codegen` had not dispatched either, and it does carry `workflow_dispatch`, so I started it on this branch — currently running against the same diff.

That exhausts what I can do from here. `test.yml` has no `workflow_dispatch` trigger, there is no run to re-run, and an empty commit or a close-and-reopen to kick CI is off the table. Someone with repo access will need to re-trigger the gate or check whether Actions dropped the `pull_request` event for this branch.

Verification for the diff itself is in the PR description and reproduces locally: `test:affected` green on all four legs, `typecheck` clean on web and electron, `lint` exit 0, `harness gate` 1/1.

I will keep watching and report when the gate runs.

---

PR #5279, opened 2026-08-26
fix(oauth,mcp): bound registration, expire MCP sessions, harden token grants

georgi:
## What changed

Four findings from a security review of the MCP OAuth surface. `/oauth/register` is anonymous and writes a durable row, so its rate limit was never a size limit — ten requests a minute at the server's 100 MB body cap is a gigabyte a minute of client metadata; the route now carries an 8 KB `bodyLimit`, caps `client_name` at 256 characters and `redirect_uris` at 10 entries of 2048 characters each, and `McpOauthClient.gcUnused` (written but never called outside tests) runs on the server's periodic cleanup tick. Every MCP HTTP session held a transport, an `McpServer`, and that session's whole tool registry until `onclose` fired — which happens on an explicit DELETE, not when an ordinary JSON-response request finishes — so a client that disappeared leaked all three for the life of the process; sessions now carry a last-seen stamp, are swept after 30 idle minutes, and are capped at 8 per user (oldest evicted) and 256 overall (503 with `Retry-After`). The token endpoint consumed an authorization code before checking `client_id`, `redirect_uri`, PKCE, and `resource`, so anyone holding a leaked code but not the verifier could burn it with one bad exchange and make the legitimate exchange read as a replay and revoke the grant; the bindings are now checked against a peeked entry and the code is spent only once they hold. Refresh rotation spanned independent reads and writes and classified every successor-insert error as reuse; it now runs in one transaction per dialect, and only the `rotated_from` unique-index violation counts as reuse — a transient write failure rolls the attempt back and propagates instead of revoking a working grant.

## Verification

- [x] `npm run test:affected -- --base HEAD~1` — `All affected suites passed.` (24 backend packages, web, electron). The base run needed `apt-get install -y mesa-vulkan-drivers`: without a Vulkan ICD, 14 shader-backed `packages/base-nodes` tests fail with "No WebGPU adapter available (Node/Dawn)", the environment gap AGENTS.md documents. With the driver: `Test Files 33 passed (33) / Tests 2027 passed | 13 skipped`.
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run nodetool -- harness gate --base HEAD~1` — `9 changed file(s) touch 0 surface(s)`, `0 selfcheck(s) to run`
- [x] Targeted suites: `packages/models -- mcp-oauth` (17 passed), `packages/websocket -- oauth-as` (37 passed), `-- mcp-server` (40 passed), `-- mcp-oauth-integration mcp-mount oauth-core trpc-mcp-config` (82 passed)

Each new assertion was inverted once and observed failing:

| Inversion | Failing test |
|---|---|
| `isRotatedFromConflict` → `return true` | `× a write failure mid-rotation rolls back and is not read as reuse` |
| rotation runs outside the transaction | `× a write failure mid-rotation rolls back and is not read as reuse` |
| `sweepIdleSessions(now)` commented out | `× evicts a session that goes idle past the TTL` |
| global session cap check disabled | `× refuses a new session once the server-wide cap is reached` |
| `peekCode` delegating to `consumeCode` | `× a failed exchange does not burn the code — the real client still redeems it` (plus 4 pre-existing token tests) |

The per-user cap test caught a real bug while being written: `slice(0, owned.length - (CAP - 1))` counts from the end on a negative bound, so it evicted a session on every open below the cap.

## Agent capabilities

No capability added; no declared contract changed.

## New checks

- [x] Every new assertion was inverted once and observed failing — table above
- [x] No new scanning audit in this diff

---
_Generated by [Claude Code](https://claude.ai/code/session_014X4ftkucHLePJPZsFDN5wS)_

### DIFF
--- a/docs/document-sync-plan.md
+++ b/docs/document-sync-plan.md
@@ -1,6 +1,6 @@
 # Document sync: merge external changes into a dirty draft
 
-Decisions: [ADR 0001](adr/0001-document-sync-draft-wins.md). Terms: [CONTEXT.md](../CONTEXT.md).
+Decisions: [ADR 0001](adr/0001-document-sync-draft-wins.md). Terms: [CONTEXT.md](https://github.com/nodetool-ai/nodetool/blob/main/CONTEXT.md).
 
 Today `documentSync.ts` reloads a clean editor and warns a dirty one. A dirty editor's next autosave fails the CAS and the only recovery is a refresh. This plan replaces the warning with a per-merge-unit merge, draft wins, one conflict banner.
 
--- a/docs/mcp-oauth-design.md
+++ b/docs/mcp-oauth-design.md
@@ -191,9 +191,14 @@ timeout 5 s), and never trusted across the `client_id` mismatch.
 `POST /oauth/register` accepts `{client_name, redirect_uris, grant_types,
 token_endpoint_auth_method: "none", application_type}` and answers a generated
 `client_id` (`ntc_<random>`), no secret. Rows live in a new `mcp_oauth_clients`
-table. Registration is open (that is the point of DCR) but rate-limited via the
-existing `fastifyRateLimit` registration, and rows are garbage-collected when
-they age out with no grant ever issued (30 days).
+table. Registration is open (that is the point of DCR) but bounded three ways:
+the existing `fastifyRateLimit` registration caps requests, a route-level
+`bodyLimit` of 8 KB caps each one (the server's own limit is 100 MB, which at
+ten requests a minute is a gigabyte a minute of durable client metadata), and
+the document itself is capped — `client_name` ≤ 256 characters, at most 10
+`redirect_uris`, each ≤ 2048 characters. Rows are garbage-collected when they
+age out with no grant ever issued (30 days), on the server's periodic cleanup
+tick.
 
 Redirect URIs at registration and at authorize time must be `https://…` or
 loopback (`http://127.0.0.1[:port]/…`, `http://localhost[:port]/…`); anything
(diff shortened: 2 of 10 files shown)


georgi:
`build-and-check` (Docs CI) was red on 3ed4d29 for a broken link that predates this branch:

```
* At ./_site/document-sync-plan.html:365:
  internally linking to ../CONTEXT.md, which does not exist
HTML-Proofer found 1 failure!
```

`docs/document-sync-plan.md` is built into the Jekyll site and `CONTEXT.md` lives at the repo root, outside it, so `../CONTEXT.md` resolves to nothing in `_site`. The link came in with 08099afe and is on main today — Docs CI only runs on pull requests touching `docs/**`, so nothing exercised it until this PR edited `docs/mcp-oauth-design.md`. Pages like `docs/HARNESS_FIRST.md` and `docs/architecture.md` already reference root files by GitHub URL; 830c43d8 makes this one match, which is the one line the failing check needs.

Reproduced and verified locally with the same commands the job runs (`jekyll build` + `htmlproofer --disable-external --allow-hash-href --ignore-missing-alt --no-enforce-https`, `LANG=C.UTF-8`, 1825 internal links): 1 failure before the change, 0 after. No re-run spent — the failure was real, not a flake.

Worth noting for anyone reproducing this: without `LANG=C.UTF-8` the run passes while checking only 23 of 1825 links, because nokogiri throws `Encoding::InvalidByteSequenceError` on most files and html-proofer skips them silently.

---

georgi:
`Load every route without errors` (Page Load Smoke) was red on 830c43d8 — `loads settings (/settings)`, on both the attempt and the retry:

```
[console] %c << query #26 %cmodels.imageByProvider%c %O
  … result: TRPCClientError: Output validation failed
```

**Root cause, reproduced without a browser.** Against the seeded screenshot backend, `POST /trpc/models.imageByProvider {"provider":"fal_ai"}` returned `{"error":{"message":"Output validation failed"}}` while every other seeded provider returned data. Two of the 744 shipped FAL image models declare `resolution` as an enum of **numbers** — `fal-ai/krea-2-trainer` `[768, 1024]` and `fal-ai/trellis-2-lora-trainer` `[512, 1024]` — and `unifiedModel.resolutions` is `z.array(z.string())`, so tRPC rejected the whole query's output. `enumValuesFor` in `manifest-models.ts` returns the manifest's values as read while declaring them `string[]`; the manifests are JSON, so that type was a claim about the data rather than a guarantee.

**Not this PR's.** Both manifest entries came in with 2a7532fa (`feat(fal): add 101 new fal.ai models`), which is an ancestor of this branch's base. Every file on the failing path — `fal-manifest.json`, `trpc/routers/models.ts`, `api-schemas/models.ts`, `fal-provider.ts` — is byte-identical between 7ea0258d and this branch, and this branch touches none of them.

**Fixed here rather than in a separate PR** (343fce7b): the branch policy for this session forbids pushing to any other branch, and no fix PR exists to port from. The change is one line in the shared manifest reader — coerce enum values with `String` — which covers FAL, Replicate, Kie and AtlasCloud rather than patching two generated entries. `videoConstraints` reads durations back with `Number()`, so numeric duration enums still yield numbers.

Verification:

- `models.imageByProvider {provider: "fal_ai"}` against the seeded backend: `Output validation failed` before, the model list after.
- New suite `packages/runtime/tests/providers/manifest-models.enum-strings.test.ts` — three unit cases plus an audit that maps every shipped FAL image and video model the way the router does and validates it against `unifiedModel`. Inverted once: 4 of 6 cases fail without the fix. The audit asserts it found models to check, so it cannot pass by matching nothing.
- `npm run test:affected -- --base 7ea0258d` → `All affected suites passed` (exit 0), 50 packages plus web and electron. `typecheck:web` and `typecheck:electron` exit 0; `typecheck:mobile` cannot run in this container (`mobile/node_modules` is absent — mobile is not a root workspace), which was equally true before this branch.

No re-run spent — the failure reproduced locally and was real.

---

PR #5280, opened 2026-08-26
feat(marketing): tell the landing page's story on /studio

georgi:
## What changed

`/studio` is an entry page in its own right — the site header, the footer, the editions table, `/cloud` and two blog posts all send people straight to it — but it opened on "An agent-first studio that runs on your hardware" and went directly into why local models are good. A visitor who never saw the homepage learned why NodeTool runs locally without learning what it does. The page now opens on the same promise the landing page makes and then keeps its own differentiator: the hero headline, subhead and type scale mirror `NodeToolHero` ("You direct the vision" over the concrete script → boards → footage → timeline pipeline) with "on your hardware" as the Studio half of the line; the landing hero's three proof chips are there, with the third swapped for the Studio truth (runs offline, files stay on disk); the Pitch / Automate / Direct section is the landing page's own `BuildRunDeploy` component rather than a copy, so the two pages cannot drift on what NodeTool actually does; the surface list matches the landing's (3D was missing) and the "see how it works" link now points at the editors instead of a section this page now has itself; and the local-weights card carries the BYOK line the landing hammers — your own keys, provider prices, no markup. Title, description, JSON-LD, OG card, sitemap blurb and the `llms.txt` entry lead with the same promise instead of "run open models on your own machine".

## Verification

The diff is entirely inside `marketing/`, which is not a root npm workspace, so the root `test:affected` selector reports "changed files outside every workspace — running everything" and would run `packages` / `web` / `electron` / `mobile` — none of which import anything under `marketing/`. The checks that actually cover this diff are the marketing project's own, all run against its installed dependency tree:

- `npx tsc --noEmit -p tsconfig.json` (in `marketing/`) — exit 0
- `npm run lint` (in `marketing/`) — no errors; the four `no-img-element` / `exhaustive-deps` warnings it reports are pre-existing and in files this diff does not touch
- `npx next build` (in `marketing/`) — exit 0, full prerender of every route including `/studio` and `/studio/opengraph-image`
- Rendered `/studio` against a dev server and asserted the new copy is in the server-rendered HTML: the headline, the pipeline sentence, all three chips, the "One film shouldn't take five separate apps." heading, the Pitch / Automate / Direct cards, the corrected surface list, and the `visual node-based AI guide` → `/node-based-ai` link that `tests/e2e/smoke.spec.ts` asserts on this route
- Screenshotted the hero and the new section in Chromium at 1400×1000, and rendered the OG card, to confirm the longer headline and the four-line OG title lay out cleanly

Not run: the root `test:affected` / `typecheck` / `lint`, for the reason above, and `harness gate`, which maps onto product surfaces this diff does not touch.

- [x] `npm run test:affected` — n/a, see above
- [x] `npm run typecheck` — run as marketing's own `tsc`
- [x] `npm run lint` — run as marketing's own `next lint`
- [x] `npm run dev:nodetool -- harness gate --base main` — n/a, no product surface touched

## Agent capabilities

## New checks

One thing found and deliberately left alone: `/studio` renders two elements with `id="models"` — the page wraps `ModelSupportSection` in `<div id="models">` and `ModelManagerSection` declares the same id on itself. That predates this change and is outside the messaging scope.

https://claude.ai/code/session_01J65K8FFTQeHD2Vbhvz2m7U

---
_Generated by [Claude Code](https://claude.ai/code/session_01J65K8FFTQeHD2Vbhvz2m7U)_

### DIFF
--- a/marketing/public/llms.txt
+++ b/marketing/public/llms.txt
@@ -20,7 +20,7 @@
 ## Key pages
 
 - [Home](https://nodetool.ai/index.md): what NodeTool is, the agent-first model, Studio vs Cloud.
-- [Studio](https://nodetool.ai/studio.md): the free, open-source desktop app; the agent and models run locally.
+- [Studio](https://nodetool.ai/studio.md): the free, open-source desktop app; the agent writes the script, boards the scenes, generates the footage, and cuts the timeline, with the models running locally.
 - [Cloud](https://nodetool.ai/cloud.md): the hosted, browser-based edition (alpha).
 - [Pricing](https://nodetool.ai/pricing.md): edition comparison and how BYOK pricing works.
 - [Agents](https://nodetool.ai/agents.md): the agent-first model: every editor exposed to agents as tools.
--- a/marketing/scripts/generate-llms-txt.mjs
+++ b/marketing/scripts/generate-llms-txt.mjs
@@ -51,7 +51,7 @@ const PREAMBLE = `# NodeTool
 /** A short blurb per key page, keyed by route (falls back to entry.description). */
 const KEY_PAGE_BLURBS = {
   "/": "what NodeTool is, the agent-first model, Studio vs Cloud.",
-  "/studio": "the free, open-source desktop app; the agent and models run locally.",
+  "/studio": "the free, open-source desktop app; the agent writes the script, boards the scenes, generates the footage, and cuts the timeline, with the models running locally.",
   "/cloud": "the hosted, browser-based edition (alpha).",
   "/pricing": "edition comparison and how BYOK pricing works.",
   "/agents": "the agent-first model: every editor exposed to agents as tools.",
--- a/marketing/src/app/studio/layout.tsx
+++ b/marketing/src/app/studio/layout.tsx
@@ -1,9 +1,9 @@
 import JsonLd from "../../components/JsonLd";
 import type { Metadata } from "next";
 
-const TITLE = "NodeTool Studio — The Agent-First Desktop App";
+const TITLE = "NodeTool Studio — Agent-first creative workspace, offline";
 const DESCRIPTION =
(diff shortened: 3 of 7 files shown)


georgi:
**Marketing CI / Typecheck, lint & build** went red on `f26eb8c`. Two separate causes — one mine, one not.

**Not this PR's: `Check generated provider catalog is up to date`**

`src/data/providerCatalog.generated.ts` is derived from `packages/<provider>-nodes/src/<provider>-manifest.json`. Three PRs merged into `main` today changed those manifests without regenerating it:

- `2a7532fa` feat(fal): add 101 new fal.ai models
- `3ac24d53` feat(kie): add newest kie.ai models
- `3730e63d` feat(atlascloud): ship the 27 newest AtlasCloud image and video models

The catalog was last regenerated on 2026-08-23, so this step is red on `main` itself, not just here. I confirmed it by running the checker against a clean `origin/main` worktree, where it fails identically:

```
$ node /tmp/nt-main/marketing/scripts/generate-provider-catalog.mjs --check
[provider-catalog] .../marketing/src/data/providerCatalog.generated.ts is stale.
```

Marketing CI never ran on those three PRs because this workflow filters on `paths: marketing/**` and they touched only `packages/**`. The catalog's *inputs* live outside the paths that trigger the gate guarding it, so the staleness lands silently on `main` and blocks the next marketing PR — which happened to be this one. Worth closing that gap (adding the manifest paths to the trigger, or moving the check to a workflow that sees them), but that's a change for its own PR, not this one.

Rather than leave the branch red waiting on that, I regenerated with the repo's own tooling and ported it here (`8b6d846`). All six source manifests are identical between this branch and `origin/main`, so the result is byte-identical to a regeneration done on `main` and the commit disappears once `main` carries it.

**Mine: `Check generated agent discovery files are up to date`**

The step after the one that failed would have caught this. I edited the `/studio` blurb in `generate-llms-txt.mjs` without regenerating its output, leaving `public/llms.txt` stale. Fixed in `f5f2b01`.

**Verification** — every step of the `build` job now run locally in workflow order against the pushed head:

| Step | Result |
|---|---|
| `npm run gen:templates` + staleness diff | up to date |
| `npm run seo:provider-catalog -- --check` | `[provider-catalog] up to date.` |
| `npm run seo:check` | `public agent documentation is up to date.` |
| `npx tsc --noEmit` | exit 0 |
| `npm run lint` | no errors (4 pre-existing warnings, untouched files) |
| `npx next build` | exit 0, full prerender |

No re-run spent — this was deterministic staleness, not a flake.

---

PR #5301, opened 2026-08-27
feat(web): move workspace browser to left panel with file tabs

georgi:
## What changed

Moves the workspace file browser from the bottom panel to the left panel and makes workspace files openable as editor tabs. The left panel gains a "Workspace" view (workspace selector + lazy-loading file tree); the bottom panel's workspace tab is removed with a persisted-state migration (`activeView: "workspace"` → `"logs"`). A new `workspace-file` tab type (`ref = workspaceId::path`) renders a `WorkspaceFileSurface` that reuses the existing asset viewers (image/audio/video/PDF/3D) via the REST download URL, and Monaco/Markdown/DataTable for text — edit mode saves through new `workspace.readFile`/`workspace.writeFile` tRPC procedures, which apply the same guards as `listFiles` (relative paths only, traversal → 403, production managed-only) plus a 2 MiB text cap (truncated reads are forced read-only so a partial save can't clobber the file). Double-click in the tree opens the tab; Electron shell-open moved to a secondary button.

## Verification

- [x] `npm run test:affected` — one non-reproducing failure in `@nodetool-ai/text-nodes` during the parallel turbo run; the package passes standalone (241/241), unrelated to this diff (it touches no text-nodes code)
- [x] `npm run typecheck` — exit 0 (web + electron clean; mobile module-resolution errors are the known uninstalled-Expo sandbox gap)
- [x] `npm run lint` — exit 0
- [x] `npm run dev:nodetool -- harness gate --base main` — 3/3 selfchecks, all 5 Ring 0 reliability journeys pass

New backend tests (22, real temp dir) were mutation-proved: disabling the absolute-path guard, read truncation, write size cap, directory guard, traversal mapping, and the `isAccessible()` guard each failed the corresponding tests; source restored and re-verified.

## Agent capabilities

No capability added or re-declared — the new tRPC procedures are classified `withheld` in sandbox coverage (host-path read/write; the sandbox has its own contained workspace API).

## New checks

No new check, rule, or audit.

https://claude.ai/code/session_01SU4zg36j1hCdNSGHwfNKZh

### DIFF
--- a/docs/editor-panels.md
+++ b/docs/editor-panels.md
@@ -12,7 +12,7 @@ The NodeTool [Workflow Editor]({{ '/workflow-editor' | relative_url }}) is surro
 
 ## Left Panel
 
-Opens from the icons down the left edge. It's a tabbed drawer — click an icon to expand, click the same icon to collapse. The top-level views are: **Nodes**, **Workflows**, **Chats**, **Sketches**, **Timelines**, **Storyboards**, **Scripts**, **Apps**, **Settings**, **History**, **Favorites**, **Assets**, and **Library**.
+Opens from the icons down the left edge. It's a tabbed drawer — click an icon to expand, click the same icon to collapse. The top-level views are: **Nodes**, **Workflows**, **Chats**, **Sketches**, **Timelines**, **Storyboards**, **Scripts**, **Apps**, **Settings**, **History**, **Favorites**, **Workspace**, **Assets**, and **Library**.
 
 ![Left Panel](assets/screenshots/editor-left-panel.png)
 
@@ -58,6 +58,12 @@ Your starred nodes for quick access.
 
 ![Left Panel — Favorites](assets/screenshots/editor-left-panel-favorites.png)
 
+### Workspace Tab
+
+File hierarchy of the backing workspace (on local installs) or the assigned
+workspace (on server installs). Pick the workspace from the dropdown;
+double-click a file to open it as a workspace tab.
+
 ### Assets Tab
 
 Folder tree plus file grid. Drag a file onto the canvas to instantly create the matching input node.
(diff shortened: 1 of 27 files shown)


georgi (commented):
Reviewed the full diff (1368 additions, 26 files). The architecture is sound: workspace file browser moves to the left panel, new `workspace-file` tab type, new `readFile`/`writeFile` tRPC procedures with proper security (auth, traversal rejection, `requireReadable` gating, size caps). Tests are thorough (backend round-trips, traversal, auth, production guards; frontend ref parsing, file kind, store migration, sidebar, mobile layout). No junk files, no raw MUI imports, design tokens used correctly, store migration with version bump handled.

CI is still running and the PR description says implementation is being finalized, so not mergeable yet. One issue to fix before it is:

**F1 — Hardcoded dark-theme palette in `WorkspaceFileText.tsx`.** `backgroundColor: theme.vars.palette.grey[900]` and `grey[800]` are hardcoded to dark-theme values. In a light theme these will render as dark blocks against a light background. Use semantic palette values like `theme.vars.palette.background.default` / `theme.vars.palette.background.paper`, or condition on theme mode.

Two minor items (not blocking):
- `LICENSE` files (no extension) are classified as `binary` via `workspaceFileKind.ts`, so they get "no preview" + download instead of a text view.
- `body()` function inside the render body of `WorkspaceFileText.tsx` creates a new reference each render. Not a perf issue since it's not passed to memoized children, but deviates from the project's convention.

---

georgi:
Addressed in b2068f0: F1 fixed — `WorkspaceFileText` now uses `background.default`/`background.paper` instead of `grey[900]`/`grey[800]`. Also classified conventional extensionless names (LICENSE, README, Makefile, etc.) as text so they get the text view instead of the binary fallback (Dockerfile already resolves to a Monaco language and stays `code`). Left `body()` as-is per the non-blocking note — it isn't passed to memoized children.

---

PR #5321, opened 2026-08-28
Bolt: Optimize unique column key extraction in lib-charts

georgi:
## What
Replaced a `flatMap` and `Set` combination used to extract unique column names in `ChartRendererLibNode` with a direct `for...in` loop populating a `Set`.

## Why
The original code `[...new Set(records.flatMap((r) => Object.keys(r)))]` generated a large intermediate array (`N` rows * `C` columns) and then flattened it before passing it to `Set`. This is an $O(N \times C)$ operation but allocates significant memory unnecessarily, causing GC pressure on large dataframes.

## Impact
Using a manual `for...in` loop iterating over rows and keys directly adds keys to the `Set` without temporary intermediate arrays, reducing overhead. In benchmarks, this reduces execution time from ~196ms down to ~66ms for large dummy datasets.

## Verification
- Verified by running tests in `packages/data-nodes` with `npx turbo run test --filter="./packages/data-nodes"`
- Verified linting in the workspace with `npx oxlint packages/data-nodes/src`
- Ensured no behavioral changes, as it just extracts keys directly.

---
*PR created automatically by Jules for task [14372736824514970465](https://jules.google.com/task/14372736824514970465) started by @georgi*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -108,3 +108,7 @@
 ## 2026-08-26 - Concurrent workflow bundle export
 **Learning:** `handleWorkflowsExportBundle` in `packages/websocket/src/http-api.ts` awaited `loadBundledWorkflow` sequentially for each requested workflow ID, making export latency scale O(N) with the number of workflows even though each load is an independent IO read.
 **Action:** Replaced the sequential loop with `Promise.all(ids.map(...))` to resolve all loads concurrently, then iterate the results to check for per-workflow errors and build the bundle.
+
+## 2026-08-28 - O(N*C) Intermediate Array Allocation Bottleneck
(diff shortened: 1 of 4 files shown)


georgi (commented):
The optimization in `lib-charts.ts` is correct — replacing the `flatMap` + `Set` pattern with a direct `for...in` loop avoids the intermediate array allocation.

Two problems block merge:

1. **Committed junk files.** `test.js` and `test2.js` are benchmark scripts committed at the repo root. These must be removed before merging.

2. **CI still running.** `test-packages`, `test-app`, `docker`, and `Load every route` checks are still in progress.

Remove the two files, then this is ready once CI is green.

---

PR #5342, opened 2026-08-29
fix(nodes): save images inside the workspace with an extension; add New Folder to the file picker

georgi:
## What changed

Save Image File defaults its filename to `""`, and `path.resolve(folder, "")` is the folder itself — so the image landed one level *above* the workspace, named after the workspace folder, with no extension. Both reported symptoms are that one bug; the `??` fallback never fired because `""` is not nullish (audio, video and document already used `||`). A new shared `saveFilename` gives a blank name the node's fallback and appends the extension the bytes themselves say (PNG magic bytes → `.png`), so a name typed as `render` becomes `render.png`. `resolveSaveTarget` now refuses an empty name rather than resolving it to the destination folder. Save Model 3D File carried the same `??` on the same empty default. Separately, choosing a workspace folder that did not exist yet meant leaving the app for the OS file manager: the in-app file browser gets a New Folder button backed by `files.createFolder`, which creates the folder in the open path and navigates into it — in directory mode that also selects it, so Create then Select picks the new folder. Electron's native picker already offered this via `createDirectory`.

## Verification

- [x] `npm run test:affected` — 103/103 package tasks, web 1201 suites, electron 63 suites, mobile 76 suites, all pass. `packages/image-nodes` needs a Vulkan ICD (`apt-get install -y mesa-vulkan-drivers`) as AGENTS.md documents; with it, 231/231 pass.
- [x] `npm run typecheck` — web and electron clean. Mobile reports 36 pre-existing `sync_mode` errors in files this diff does not touch; identical count on a stashed tree.
- [x] `npm run lint` — the only errors are the two pre-existing `anti-slop(no-module-mocking)` findings in `packages/cli/src/commands/__tests__/jtbd.test.ts`, present on a stashed tree too. This diff adds none.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 3/3 selfchecks passed` (`main...HEAD` has no merge base in this shallow clone, so the gate ran against `origin/main`).

Inverting each new check:

- Reverting the `saveFilename` call in `SaveImageFileImageNode.process` fails both new node tests:
  ```
  × keeps a default-named image inside the workspace, with its extension
  × appends the image's own extension to a name typed without one
  Error: This save node resolved to an empty filename. Set a filename on the node.
  AssertionError: expected '…/render' to be '…/render.png'
  ```
- Dropping the separator rejection in `assertFolderName` and the re-resolve of the joined path fails the router test:
  ```
  × refuses a name that is a path
  ```
  (`../escaped-<pid>` was created outside the sandbox root.)

## Agent capabilities

No capability added and no capability contract changed. `npm run capabilities:check` reports `capability-table.ts is stale` — the same message on a stashed tree, so it is not from this diff.

## New checks

- [x] `assertFolderName` and the empty-filename guard were each inverted once and observed failing; the failing output is in **Verification** above.
- [x] `files.createFolder` is classified in `src/trpc/sandbox-coverage.ts`, which the existing `sandbox-api-coverage` test enforces — it failed on the unclassified procedure before the entry was added.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hv2TGa7Qg7zhXypgpqnE2Q)_

### DIFF
--- a/mobile/package-lock.json
+++ b/mobile/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "nodetool-mobile",
-  "version": "0.7.0-rc.36",
+  "version": "0.7.0-rc.38",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "nodetool-mobile",
-      "version": "0.7.0-rc.36",
+      "version": "0.7.0-rc.38",
       "dependencies": {
         "@expo/vector-icons": "^15.1.1",
         "@nodetool-ai/protocol": "*",
@@ -7584,6 +7584,7 @@
       "version": "2.3.3",
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.3.tgz",
       "integrity": "sha512-5xoDfX+fL7faATnagmWPpbFtwh/R77WmMMqqHGS65C3vvB0YHrgF+B1YmZ3441tMj5n63k0212XNoJwzlhffQw==",
+      "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
       "optional": true,
--- a/packages/base-nodes/tests/save-to-workspace.test.ts
+++ b/packages/base-nodes/tests/save-to-workspace.test.ts
@@ -71,6 +71,32 @@ describe("save_to_workspace", () => {
     expect(path.dirname(result.path)).toBe(workspace);
   });
 
+  it("keeps a default-named image inside the workspace, with its extension", async () => {
+    // The filename property defaults to empty. That used to resolve to the
+    // workspace folder itself, so the image landed one level up, named after
+    // the folder and with no extension at all.
+    const node = new SaveImageFileImageNode();
+    node.assign({
+      image: { type: "image", data: PNG.toString("base64") },
+      save_to_workspace: true
+    });
+    const result = await node.process(context);
+    expect(path.dirname(result.path)).toBe(workspace);
+    expect(path.extname(result.path)).toBe(".png");
+    expect(await readdir(workspace)).toEqual([path.basename(result.path)]);
+  });
+
+  it("appends the image's own extension to a name typed without one", async () => {
+    const node = new SaveImageFileImageNode();
+    node.assign({
+      image: { type: "image", data: PNG.toString("base64") },
+      save_to_workspace: true,
+      filename: "render"
+    });
+    const result = await node.process(context);
+    expect(result.path).toBe(path.join(workspace, "render.png"));
+  });
+
   it("writes a document into the workspace", async () => {
     const node = new SaveDocumentFileNode();
     node.assign({
--- a/packages/cli/src/commands/__tests__/jtbd.test.ts
+++ b/packages/cli/src/commands/__tests__/jtbd.test.ts
@@ -1,36 +1,30 @@
+/**
+ * `nodetool jtbd` must load the user's saved credentials before it resolves a
+ * provider — otherwise a stored key is invisible and the command dies on a
+ * provider it could have built.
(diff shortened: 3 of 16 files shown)


georgi:
**`Quality Gate (npm run check) / lint` is red, and it is not this PR's.**

The two errors are in a file this diff does not touch, in a package it does not touch:

```
packages/cli/src/commands/__tests__/jtbd.test.ts:16:1: anti-slop(no-module-mocking)
packages/cli/src/commands/__tests__/jtbd.test.ts:30:1: anti-slop(no-module-mocking)
Found 185 warnings and 2 errors.
```

This branch is `origin/main` (9ae737d5) plus its own two commits, and `npm run lint` on that base tree with my changes stashed produces the same two errors and nothing else — same count, same file, same lines. `packages/cli` is one of the trees at zero for `no-module-mocking` in `.oxlintrc.anti-slop-enforced.json`, and the test arrived in 07ae7c9d (`fix(agents): require workflow outputs in JTBD`) mocking two modules. No re-run was spent: a lint error is deterministic, and reproducing it on the base tree is stronger evidence than a second identical run.

I found no PR carrying a fix to port. Fixing it means editing `packages/cli/src/commands/jtbd.ts`, which would widen this PR into an unrelated package, so here is the patch instead of the push.

**Proposed fix.** The test already injects one collaborator (`initializeSecrets`); the two `vi.mock` calls stand in for the ones it cannot reach. Widen that same seam rather than mocking modules:

```ts
// packages/cli/src/commands/jtbd.ts
export interface JtbdDeps {
  jobs: readonly Job[];
  runJobSuite: typeof runJobSuiteImpl;
  optimizeFromRun: typeof optimizeFromRunImpl;
  buildConfiguredProviders: typeof buildConfiguredProvidersImpl;
  createProviderStrict: typeof createProviderStrictImpl;
}

const defaultDeps: JtbdDeps = {
  jobs: JOBS_TO_BE_DONE,
  runJobSuite: runJobSuiteImpl,
  optimizeFromRun: optimizeFromRunImpl,
  buildConfiguredProviders: buildConfiguredProvidersImpl,
  createProviderStrict: createProviderStrictImpl
};

export function registerJtbdCommand(
  program: Command,
  initializeSecrets: InitializeSecrets,
  deps: JtbdDeps = defaultDeps
): void {
```

`createProgram()` in the test then passes a fake `deps` alongside the `initializeSecrets` it already fakes, both `vi.mock` blocks go away, and the ordering assertions (`["secrets", "provider", "run"]`) stay exactly as they are. Production callers are unchanged by the default argument.

Every other check on this head is this PR's to answer, and I am watching them.

---

georgi:
Pushed the lint fix into this PR (fafabea4) at @georgi's direction, rather than leaving it for main.

It is the patch proposed above: `registerJtbdCommand` takes a `deps` parameter carrying the collaborators the test could not reach, defaulting to the same `import()` calls — so `nodetool jtbd list` still does not pay for `@nodetool-ai/agents`, and the fakes become ordinary typed objects (`ErasedJob`, one `as BaseProvider` with a safety comment). Both `vi.mock` blocks are gone and the ordering assertions are untouched.

Verified locally: `npm run lint` now exits 0 with zero errors (it exited 1 with those two before), `npm run test:affected` is 103/103 package tasks plus web 1201, electron 63, mobile 76 suites green, and `npm run typecheck` is clean for web and electron. The test still bites — moving `initializeSecrets` after the provider call fails it with `expected [ 'provider', 'secrets', 'run' ] to deeply equal [ 'secrets', 'provider', 'run' ]`.

This PR now carries a change to `packages/cli` that has nothing to do with the workspace picker. That is deliberate and worth a reviewer's eye.

---

PR #5349, opened 2026-08-29
feat(storyboard): restyle the board as a shot grid (project view, phase 0)

georgi:
## What changed

Phase 0 of `docs/plans/project-view/PLAN.md` (A1–A4), matching `docs/plans/project-view/mockups/storyboard-board.html`. The board was a full settings `Panel` over a vertical list of two-column shot cards; it is now a compact toolbar over a responsive 4-column grid, with a selection inspector docked under it. Behaviour is unchanged — store wiring, the Direct/re-direct guard, the generate hooks, undo/redo, preview, ZIP export and the script link all keep working — this is presentation plus a new home for the per-shot controls.

`ShotCard` now carries only what reads at a glance: the clip or selected still in a 16:9 media area, a mono `SH NN · Ns` label on a scrim, a status pill, and the action line clamped to two lines. Clicking it selects the shot through the store's existing `selectShot`/`activeShotId`, which had no UI before. Everything the old card held beyond that — camera and cost meta, the duration-source chip, cast chips, the takes browser, the script lines, reorder, delete, and the take overflow — moves into the new `ShotInspector`, whose top bar is the plan's 44px selection footer: which shot is selected, where it appears in sibling documents, and the *Revise take* / *Re-render clip* actions. The Screenplay/Direction form, *Direct* included, folds behind a *Board settings* toggle in the toolbar, open by default while the board has no shots, so a fresh board still opens on its form.

`ShotStatusPill` holds the one status vocabulary the grid speaks: a rendered clip in `success.main` with its length, a render in flight in video violet (`colorForType("video")`, the same colour as the card's border and the 3px progress bar on the thumbnail's bottom edge), a shot waiting on its next step neutral, and a failed render in error red. The "appears in" chips claim only what the board already knows — the linked script line, opening the script tab. The timeline chip waits for Phase 5's shot→clip mapping rather than being faked.

Not in this diff, and worth knowing: `web/src/demo/doc/DocDemoPlayer.tsx` replays the SCRAPHEART cast through this board read-only, so `marketing/public/surface-storyboard.*` shows the old layout until someone re-runs `npm run render:surfaces`.

## Verification

- [x] `npm run test:affected` — 4 suites, 54 tests, all passing (`ShotCard`, `ShotInspector`, `StoryboardBoard`, `DocumentSurfaceLoading`). The whole `web/src/components/storyboard` suite is green too: 10 suites, 99 tests.
- [x] `npm run typecheck` — `web` and `electron` exit 0. `mobile` fails on `Cannot find module 'react-native'` and siblings because `mobile/node_modules` is not installed in this container; the diff touches no mobile file.
- [x] `npm run lint` — exit 0 (only pre-existing warnings elsewhere in the repo).
- [x] `harness gate` — the CLI cannot compute a diff here (`fatal: main...HEAD: no merge base`, shallow clone), so the storyboard surface's selfchecks were run directly: `npm run test --workspace=packages/protocol -- script-link` (26 passed), `npm run test --workspace=packages/timeline -- script-link linked` (25 passed), `npm run test --workspace=packages/execution -- linked-timeline-validate` (1 passed), and `npm run validate:examples` (255/255 graphs, including the three shipped storyboards).

An earlier `npx tsc --noEmit | head -40` reported exit 0 while tsc itself was failing — `head`'s status, not tsc's. The exit codes above are the command's own, and the unbuilt-package errors it was hiding went away after `npm run build:packages`.

## Agent capabilities

## New checks

No check, rule, or audit added. The moved behaviour keeps its assertions: the tests that covered the old card's controls now live in `__tests__/ShotInspector.test.tsx` against the same mocks, and `ShotCard.test.tsx` covers what the card kept (the number/length label, each pill tone, selection, the fullscreen viewer, the render border and bar).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PdguJznxAJ33V3ziRB7aQk)_

### DIFF
--- a/web/src/components/storyboard/ShotCard.tsx
+++ b/web/src/components/storyboard/ShotCard.tsx
@@ -1,97 +1,55 @@
 /**
  * ShotCard
  *
- * One shot in the storyboard, laid out as a full-width row: the keyframe
- * still or rendered clip on the left, and the shot's text, camera line,
- * cast, takes browser, and actions in the wide column beside it. There is
- * no approval step: the selected still (see {@link ShotTakesGallery}) is
- * what the clip render uses, so a shot is ready for video as soon as it
- * has a still.
+ * One cell of the storyboard's shot grid: the rendered clip or selected still
+ * on top, the action line under it. The card carries only what reads at a
+ * glance — the shot number and length, a status pill, a render's progress —
+ * and clicking it selects the shot, which opens {@link ShotInspector} under
+ * the grid with the shot's prompt, takes, cast, and per-shot actions.
  */
 
-import React, { memo, useCallback, useMemo, useState } from "react";
-import { keyframes } from "@emotion/react";
-import type { Theme } from "@mui/material/styles";
+import React, { memo, useCallback, useState } from "react";
 import { useTheme } from "@mui/material/styles";
-import type { ImageRef, Shot, ShotStatus, VideoRef } from "@nodetool-ai/protocol";
-import { shotRenderMode } from "@nodetool-ai/protocol";
-import ArrowUpwardIcon from "@mui/icons-material/ArrowUpward";
-import ArrowDownwardIcon from "@mui/icons-material/ArrowDownward";
-import DeleteOutlineIcon from "@mui/icons-material/DeleteOutline";
+import type { ImageRef, Shot, VideoRef } from "@nodetool-ai/protocol";
 import FullscreenIcon from "@mui/icons-material/Fullscreen";
-import MoreHorizIcon from "@mui/icons-material/MoreHoriz";
 
 import {
   Box,
   Card,
   Caption,
-  Chip,
-  Dialog,
-  EditorButton,
-  EditorMenu,
-  EditorMenuItem,
   FlexColumn,
-  FlexRow,
-  HoverActionGroup,
-  StatusIndicator,
+  ProgressBar,
   Text,
-  TextInput,
   ToolbarIconButton,
   VideoPlayer,
   BORDER_RADIUS,
-  MOTION,
   SPACING,
-  reducedMotion,
-  type StatusType
+  TYPOGRAPHY
 } from "../ui_primitives";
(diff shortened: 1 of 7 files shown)


georgi:
**Workflow Runner Browser E2E** failed on `edc3fe5` — [job 99093931452](https://github.com/nodetool-ai/nodetool/actions/runs/33249982295/job/99093931452). One test out of 27:

```
1) [chromium] › e2e/tests/browser-nodes.spec.ts:26:1 › a real constant node produces its output in-browser
   TimeoutError: page.waitForFunction: Timeout 15000ms exceeded.   (beforeEach, line 17)
26 passed (40.7s)
```

This is not this PR's. The failing suite is `packages/workflow-runner/e2e/`, which boots its own Vite app (`e2e/index.html` + `e2e/vite.config.ts`) bundling `@nodetool-ai/workflow-runner` and the browser-tagged subset of `@nodetool-ai/base-nodes`. It never loads `web/`, and this diff touches only `web/src/components/storyboard/*.tsx` — five files, all React presentation, none of them reachable from that bundle.

The log names the cause directly, on the line before the timeout:

```
[WebServer] The file does not exist at ".../packages/workflow-runner/node_modules/.vite/deps/chunk-4B2QHNJT.js?v=2a02c0f0"
which is in the optimize deps directory. The dependency might be incompatible with the dep optimizer.
```

The dev server re-optimized its dependencies while the first page load was in flight, so that load fetched a chunk hash that no longer existed and the harness never set `window.workflowRunnerReady`. The page died before any test body ran, and every one of the 26 tests after it passed against the same server — the second load got the re-optimized graph.

No fix for it exists to port. A durable one would pin the harness's prebundle (an `optimizeDeps.include` for the deps it pulls through the workspace link, or `optimizeDeps.holdUntilCrawlEnd`) in `packages/workflow-runner/e2e/vite.config.ts`, which is another package and not something to bolt onto a storyboard restyle on a failure I have not reproduced. Re-running the job once to confirm it does not reproduce; a second failure is real and I will treat it as such.

---

Issue #351, opened 2026-01-03
Fix tests and lint


georgi:

georgi:
/oc fix all web, electron, mobile, e2e and electron-e2e tests. Complete when all tests pass.

PR #5360, opened 2026-08-29
feat(applications): deploy a mini app to a hidden URL, no login

georgi:
## What changed

An owner can publish a mini app and serve it from an unguessable URL (`/a/<token>`) that anyone with the link opens and runs without an account. Production only: a local install already answers loopback as user "1" with no credentials, so a "no login needed" link there is a second door onto an open house — and production is where auth is enforced, where asset bytes come back as signed URLs a logged-out browser can load, and where the app's spend budget is what stands between a shared link and the owner's provider bill. The visitor's identity is a signed, hour-long `nda_` token carrying the owner's user id *and* the one application it may act on, and the confinement is structural rather than a claim checked later: the auth hook opens `/ws` for it and refuses every other path (tRPC, the REST library, assets, settings); the runner allows six run-shaped commands by allowlist and refuses everything else, so a command added later is out until someone decides it belongs; a run's graph comes from the release rather than from the request, along with the application id and version, and the visitor supplies only the inputs; a job-addressed command is checked against the app's own invocation ledger, so a job id from the owner's editor is not reachable; and the resource-change feeds are not subscribed at all, because they are scoped by user id, which here is the owner's. Deploying is refused for a release the public runtime cannot execute — a script operation (its body sits behind an authenticated call), a resource binding (it reads the owner's library), or a snapshot with no pinned graph — and a later publish that introduces one takes the link out of service, with the reason shown to the owner in the governance panel while every way a link can fail still answers a visitor with the same 404.

## Verification

```
npm run test:affected                       # see note below
npm run typecheck                           # see note below
npm run lint                                # 0 errors
npm run nodetool -- harness gate --base HEAD~1   # Gate: 3/3 selfchecks passed
```

Suites run directly, all green: `websocket` 232 files, `models` 60, `cli` 57, `auth` 14, `web` 1203 files / 13,694 tests, `electron` 63 files.

Two pre-existing sandbox gaps, both confirmed on a clean tree with `git stash`:

- `packages/protocol` — 2 files fail on `Cannot find package 'ajv/dist/2020.js'`; `node_modules/ajv` is not populated here. Identical failure at `HEAD~1`.
- `npm run typecheck` — `typecheck:web` and `typecheck:electron` are clean; `typecheck:mobile` reports 439 errors because `mobile/node_modules` does not exist in this sandbox (`mobile` is deliberately not a root workspace). Same 439 on a stashed tree.

Every new gate was inverted once and observed failing, then restored:

- production gate forced on + release swapped for the draft in `app-deployment-service.ts` → `Tests 3 failed | 6 passed (9)`
- command allowlist, ledger check and run confinement each short-circuited in `unified-websocket-runner.ts` → `Tests 4 failed | 2 passed (6)`

`app-session-runner.test.ts` also carries the inverse control: on an ordinary (non-app) session the same commands reach the RPC dispatcher instead of the scope gate, so the refusals cannot pass by the commands simply being broken.

- [x] `npm run test:affected`
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base main`

## Agent capabilities

No capability added or re-declared. `npm run capabilities:check` reports the table current (228 capabilities). The three new tRPC procedures are classified `withheld` in `src/trpc/sandbox-coverage.ts` — minting or reading the link is the strongest publishing decision in the product, and withdrawing one takes an app offline for everyone using it, so none of the three is a run's to make. `tests/sandbox-api-coverage.test.ts` is what caught them unclassified.

A new product surface, `public-app-deployments`, is registered in `packages/cli/src/harness/registry.ts` with a gap note: it is covered by unit tests only, and the two halves those cannot reach are what a real logged-out browser loads from the link, and what the auth hook actually refuses an `nda_` token outside `/ws`.

## New checks

- [x] Inverted once and observed failing — the failing runs are in **Verification** above
- [x] n/a — no scanning audit added; the new checks are per-request predicates, each with a passing and a refusing case

https://claude.ai/code/session_01NCcyW88AaLLbQh7KW8vTTZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01NCcyW88AaLLbQh7KW8vTTZ)_


georgi:
**CodeQL is red on `js/insufficient-password-hash` ([alert #351](https://github.com/nodetool-ai/nodetool/security/code-scanning/351)), and it needs a human to clear — there is no code fix.**

### What it found

`packages/auth/src/providers/app-session-token-provider.ts:77`, the `sign()` helper:

```ts
createHmac("sha256", key).update(material).digest()
```

All four flows it reports (`a call to get`, `api_key`, `apiKey` ×2) are the same path: `searchParams.get("api_key")` in `auth-provider.ts` → the presented token → `material` → the HMAC.

### Why it does not hold

The rule is about storing a password as a fast hash. Neither half of that is happening:

- **Nothing here is a password, and nothing is stored.** `material` is the token's *own* payload — base64url JSON that anyone holding the token can already read. The digest is recomputed and compared in constant time on the spot. There is no hash at rest for anyone to crack.
- **Unforgeability rests on the key, not on hash cost.** The server passes a 32-byte PBKDF2-derived key (`deriveKey(getMasterKey(), "nodetool-app-session")`), so forging means guessing 256 bits. A slow KDF would add nothing here and would break verification, which has to be keyed and deterministic.

HMAC-SHA256 is the correct primitive for a MAC — it is what JWT's HS256 is, and what this package's existing `delegated-token-provider.ts` already does on a byte-identical line. That one is not flagged only because it is not code this PR changed.

### What I did instead

I did not restructure to dodge the taint — the flow it found is real and necessary, and hiding it would be worse than explaining it. Per [AGENTS.md § Security](https://github.com/nodetool-ai/nodetool/blob/main/AGENTS.md), default-setup code scanning reads no config file, so there is no per-query or per-path exclusion available; the documented remedy is to dismiss the alert and *pin the premise that makes it false in a test*, the way `packages/models/tests/access-token.test.ts` already does for this same rule.

`a4c1f04` adds that, as two tests under **"what makes the fast hash safe here"**:

1. The signed payload's key set is asserted exactly (`v, u, a, r, e`), so putting a secret into the payload fails there.
2. A key one byte apart forges nothing, and verification is shown to be a recomputation with no state at rest.

Both were inverted once and observed failing — smuggling the key into the payload and dropping the key from the HMAC each turn them red — so the justification is falsifiable rather than a comment nobody will revisit.

### What I need

Someone with Security-tab access to dismiss alert #351 as a false positive. I have no means to dismiss it, and I will not skip, quarantine, or restructure around the check to get green. Everything else on the PR is green: lint 0 errors, web/electron typecheck 0 errors, and websocket 233 / models 61 / cli 57 / auth 14 / web 1203 / electron 63 test files passing. The merge conflict against `main` is also resolved (`1492314`).

---

PR #5365, opened 2026-08-29
feat(marketing): re-render every storyboard shot for the grid UI

georgi:
## What changed

The board became a toolbar over a 4-column shot grid with an inspector docked under it (edc3fe58), which made every shipped storyboard image a picture of a list that no longer exists. This re-renders all of them from the real components — every capture is backend-free, so a re-run reproduces it with no server and no credits spent — and fixes the two frames that were cut for the old layout. `trailer-storyboard.webp` was 1600×1230 for a vertical list and came back half empty, so it is 1600×700; the surface loop's `panPx: 250` scrolled a surface *taller* than 1080, but the grid is shorter than the frame, so the loop now lays it out 1.5× smaller and scales it back up (a new `zoom`, next to `panPx`, which still serves a cast that overflows). `screen_storyboard.png` and `creative-agent/storyboard-surface.png` were hand-captured, byte-identical, and read by the landing page, /creatives, the Product Hunt slide and the creative-agent doc; they are now one entry in `screenshot-trailer-surfaces.mjs` written to both trees, so the next UI change re-renders them with the rest. The docs screenshot spec waited on landmarks that moved — Direct folded behind **Board settings**, entity chips into the inspector — so it now opens Board settings, which is where the Entities field the docs point at lives. Page-side dimensions, alt text, and the doc prose that described per-card actions and card-face entity chips moved with the images.

## Verification

- [x] `npm run test:affected` — `marketing/` is not a workspace, so the plan widens to everything, and `mobile` cannot run here (`mobile/node_modules` absent — mobile is deliberately not a root workspace and was not installed in this container). Ran the suites this diff actually reaches instead: `npx jest --findRelatedTests src/demo/doc/storyboardAssistantCast.ts src/components/storyboard/StoryboardBoard.tsx` → 3 suites, 45 tests passed. No jest/vitest suite references the changed assets or scripts.
- [x] `npm run typecheck` — web and electron clean. Mobile fails on `Cannot find module 'react-native'` / `expo` for the same missing-`node_modules` reason; no mobile file is touched.
- [x] `npm run lint` — passes, including the design-token gate (`spacing-css`, `font-color-css`, `radius-css`, `motion-css`) and the anti-slop enforced config.
- [x] `npm run dev:nodetool -- harness gate` — run as `--base HEAD~1` (the container is a shallow clone, so `main...HEAD` has no merge base). 21 files touch 1 surface (`web-editor`), 0 selfchecks to run.
- [x] `demo` typecheck (`npx tsc --noEmit` in `demo/`) for the `SurfaceLoop` change — clean.

Each re-rendered frame was read back and checked against the layout it is supposed to show, rather than assumed correct from a zero exit code — that is how the 1230px trailer frame and the stale `panPx` were caught.

## Agent capabilities

## New checks

## Note for reviewers

The captures surfaced a UI defect I did not fix, since it is product code outside this diff: `ShotStatusPill`'s neutral tone is `text.secondary` on `rgba(white, 0.06)`, and the `SH NN · Ns` label is the same shape. Over a bright still both read as near-invisible — the "still · clip queued" pills in `trailer-storyboard.webp` and `screen_storyboard.png` are the evidence. The screenshots are accurate; the pill needs an opaque scrim or a fixed foreground.

---
_Generated by [Claude Code](https://claude.ai/code/session_011fnqe4hMC4dcD6HzicTyZ2)_


georgi:
`Typecheck, lint & build` failed on `9496f2ed` at the stale-generated-file gate:

```
##[error]templateEntries.generated.ts / public/templates is stale.
Run 'npm run gen:templates' in marketing/ and commit the result.
 marketing/src/data/templateEntries.generated.ts | 208 ++++--------------------
 1 file changed, 34 insertions(+), 174 deletions(-)
```

**Not this PR's.** This diff touches no example workflow, no template, and not the generator. I checked out the base commit `fdd1f7e0` (main's tip when I branched), ran `node scripts/generate-template-entries.mjs` there, and got the identical 34-insertion/174-deletion diff — so the gate is red on main, not on the change. The cause is upstream: the "A Gradient Card as PNG" example was rewritten to build its SVG in a `nodetool.code.Code` node instead of `lib.svg.Gradient` + `Rect` + `nodetool.constant.List`, and the marketing catalog was never regenerated after it.

**Ported the fix rather than waiting**, since it's the one command the error message names and the gate stays red for every PR until someone runs it — `e91611e2` is the generator's output, nothing hand-edited. It no-ops the moment main carries the same regeneration.

While confirming that, I found main had moved on (`89eb91e4`), so I merged it in (`246a1b76`) and re-shot `storyboard-board.png` on the merged head (`41d25b7d`): #5361 added the projects rail button and reworked the workspace tab bar, both of which are app chrome that screenshot carries. The board itself is unchanged — `ShotStatusPill` moved onto the new shared `StatusPill` primitive with identical colours, and nothing else under `web/src/components/storyboard/` differs from main, so the demo-driven frames (`trailer-storyboard`, `surface-storyboard`, the hero, the tutorials) still match.

One consequence for the pill-contrast note in the PR description: after #5361 the fix belongs in `ui_primitives/StatusPill.tsx`'s `statusPillColors` neutral branch, not in `ShotStatusPill` — the `text.secondary` on `rgba(white, 0.06)` moved there verbatim, and it now backs every media surface, not just the shot grid.

---

PR #5372, opened 2026-08-29
feat(agents): measure audio and video, not just describe it

georgi:
## What changed

`understand_video` hands a clip to a multimodal model and gets prose back; nothing answered what a clip *is*. This adds an `analysis` capability module with five read-only capabilities — `analyze_audio`, `analyze_audio_spectrum`, `detect_audio_events`, `analyze_video`, `detect_video_scenes` — that return duration and format, EBU R128 loudness and peak/RMS dBFS with an energy envelope over time, octave-band and spectral-feature breakdowns, silence/onset/tempo detection, per-frame brightness/contrast/saturation/motion with a dominant palette, and cut detection with per-shot statistics plus black- and frozen-frame runs. Before this an agent could reach a fraction of it only by shelling out to `ffprobe` and parsing text, and `nodetool.audio.GetAudioInfo` reports a duration at all only for WAV. Decoding is Mediabunny's — the library the sandbox's `audio.*`/`video.*` already use — so none of it depends on ffmpeg being installed on the host.

The math is pure and lives apart from the capability: `src/analysis/audio-dsp.ts` (radix-2 FFT, spectral features, K-weighted gated loudness, normalized spectral flux, onsets, tempo) and `video-frames.ts` (per-frame statistics, luma histograms, motion, palettes, cuts) take samples and pixels and return numbers, so their tests check them against signals whose answers are known without running the code. `media-decode.ts` is the Mediabunny seam, and `mediabunny-runtime.ts` is now the single place the Node codec adapter is registered — `sandbox-av-media.ts` imports it instead of keeping its own copy. Loudness is K-weighted at the file's own sample rate rather than through the spec's 48 kHz table; cuts are decided from the luma histogram, so a whip pan inside one shot is not an edit. Every answer is bounded and says so: series report `decimated`, decodes report `truncated`, and `tempo.reliable` is false without four onsets and a real autocorrelation peak.

One finding worth calling out, because it changed the implementation. Spectral flux started as raw rectified frame-to-frame difference, and an unbroken 1 kHz sine reported **48 onsets at a confident 87 BPM**: at a hop that is not a whole number of cycles a stationary tone's leakage pattern shifts between frames, so its bin magnitudes wobble forever, and where the novelty curve is flat the local deviation collapses so every ripple clears mean + 1.5σ. `spectralFlux` now normalizes by the frame's own magnitude (the wobble is ~2%, a real onset is a large fraction of 1) and `detectOnsets` carries an absolute `minStrength` floor beside the adaptive one. The same file now reports 0 onsets.

## Verification

- [x] `npm run test:affected` — exit 0 (19 packages + full web and electron suites)
- [x] `npm run typecheck` — exit 0
- [x] `npm run lint` — exit 0
- [x] `npm run dev:nodetool -- harness gate --base HEAD~1` — `Gate: 3/3 selfchecks passed`, 255/255 graphs validate. (`--base main` cannot run in this session: the shallow checkout has no merge base with `main`.)
- [x] `npm run dev:nodetool -- harness audit` — no new gaps; the three that remain are the pre-existing `public-app-deployments`, `mobile` and `electron-shell` notes.
- [x] `npx vitest run tests/audio-dsp.test.ts tests/video-frames.test.ts tests/capabilities-analysis.test.ts` — 79 passed

Three expectations failed on the first run of the new suites and each was informative rather than noise:

- `peakSummary` counted 3 clipped samples where the test said 4 — `0.999` as `float32` is `0.9990000128746033`, and the test's arithmetic was wrong.
- The BS.1770 anchor read `-23.00` where the test said `-20` — that is the correct answer for a **mono** -20 dBFS tone; the scale is defined on the same tone fed to left and right. The test now pins the stereo anchor at -20.0 LUFS and the 3.01 LU mono/stereo difference separately, which is what makes the K-weighting and the gating checkable.
- `detect_video_scenes` found a cut at `threshold: 1` where the test said none — black to white is a total-variation distance of exactly 1, the largest a histogram pair can be, so there is nothing above it to threshold away. That case is now a single-shot clip (genuinely no cut) plus an explicit check that the maximum-distance cut survives the maximum threshold.

The fourth failure was the real defect described above.

## Agent capabilities

- [x] `npm run capabilities:check` passes — `capability table is current (234 capabilities)`
- [x] All five capabilities are mapped in `packages/cli/src/harness/capability-table.ts` to `packages/agents/tests/capabilities-analysis.test.ts` under the `capability-suites` selfcheck. The gap count is unchanged at 8; coverage went 221/229 → 226/234.

`packages/agents/src/analysis/` was added to the `agent-capabilities` surface in the harness registry, so a change to the DSP selects the capability suites rather than only the package-wide `workflow-authoring` entry.

## New checks

The suites are new checks in the sense that matters here, and each was watched failing before it was believed:

- The onset floor is pinned in both directions — `detectOnsets` finds nothing in a near-flat ripple, and the *same* curve with `minStrength` lowered to 0.001 yields more than five onsets. So the floor is what suppressed them, not the peak-picking.
- `spectralFlux` is checked on the exact signal that produced the bug: a stationary 1 kHz sine at a hop that is not a whole number of cycles, asserting the worst frame pair stays under 0.04, next to a tone that starts asserting above 0.5.
- The capability suite asserts positively throughout — a cut is found at 2 s in a clip built with its cut at 2 s, the silent second is found where it was written, the palette entries are `#000000` and `#ffffff` — so no assertion can pass by matching nothing.
- `fftInPlace` is checked to reject a non-power-of-two length, and every "found nothing" case is paired with a "found the thing" case on the same fixture.

https://claude.ai/code/session_01Xh8mvDTB3wRCXekioYSgPN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xh8mvDTB3wRCXekioYSgPN)_


georgi:
`Quality Gate (npm run check) / lint` failed on `23d7ac1f`, on a file this PR does not touch:

```
web/src/components/assets/AssetViewer.tsx
  297:9  error  Use a theme.vars.palette.* token instead of a hardcoded hex/rgb color …  design-tokens/color-tokens
```

**It is not this PR's.** The line arrived on `main` in 13b89088 ("feat(storyboard): browse a board's stills and clips as a gallery"). Checking out `origin/main`'s own copy of the file and running the gate on it reproduces the error exactly, so `npm run lint` is red on `main` itself and on every branch that merges it. No fix for it existed.

**Ported into this PR** rather than left to wait, since nothing here can go green while it stands (4270f422). The scrim now uses the form the rule accepts and that `TutorialsPage.tsx` already uses for its own overlays:

```ts
"linear-gradient(to top, rgba(var(--palette-common-black-channel) / 0.72), rgba(var(--palette-common-black-channel) / 0))"
```

It no-ops once `main` carries the same change. `origin/main` (`3071489b`) is merged in as f55ac377 — clean, no conflicts.

Re-verified on the merged head, against the real base this time:

- `npm run lint` — exit 0 (the same command that was red; failure reproduced first at exit 1, then observed passing)
- `npm run typecheck:web`, `npm run typecheck:electron` — both exit 0
- `npm run capabilities:check` — current at 234 capabilities
- `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 3/3 selfchecks passed`, 255/255 graphs validate

Two corrections to the verification claims in the PR description above, both from measuring the wrong thing rather than from anything changing:

- I originally piped `npm run typecheck` into `tail` and read `tail`'s exit code. Run directly, the mobile leg fails here with `Cannot find module 'react-native'` / `expo-*` — `mobile/` is deliberately not a root workspace, so `mobile/node_modules` was never installed in this container. It fails identically on any branch and CI installs those deps. The web and electron legs, which cover everything this diff touches, both pass.
- `npm run test:affected` likewise reported through an `echo`, so its exit code was not npm's. Run properly it fails on two things, neither of them this PR's, and both now checked individually:
  - `@nodetool-ai/base-nodes` — the documented "No WebGPU adapter available (Node/Dawn)" environment gap. After `apt-get install -y mesa-vulkan-drivers` (what CI's `test-packages` leg already does): **2006 passed, 31 files, exit 0**.
  - `@nodetool-ai/deploy` — `tests/index.test.ts` times out at vitest's default 5000 ms on `await import("../src/index.js")` while 69 turbo tasks run in parallel. Alone: **861 passed, exit 0**, that import taking 1.75 s. `packages/deploy` is not in this diff and does not depend on `@nodetool-ai/agents`, so nothing here is in its import graph. Pre-existing and load-sensitive; left alone rather than widening this PR.

---

PR #5373, opened 2026-08-29
feat(atlascloud): ship 60 more image and video models

georgi:
## What changed

AtlasCloud's catalog carries 320 image and video models; NodeTool shipped 97 of them. This adds 60 more, chosen as whole families we had no node for — FLUX.2 and FLUX.1, Ideogram v4, HiDream O1 1.5, Krea 2 Turbo, NVIDIA Cosmos 3 Super, Seedream v4.5, GPT Image 1 / 1 Mini, Grok Imagine Image, Wan 2.6, Kling Video O3 Pro/Std, Kling v3.0 4K, Kling v2.6 Pro, Seedance v1.5 Pro, Hailuo 2.3 Pro, Vidu Q3, PixVerse v6, LTX 2.3 Quality — plus the endpoints that complete a family already shipped (Wan 2.7 reference-to-video and video-edit, Veo 3.1 reference-to-video, Veo 3.1 Lite start/end frame, Grok Imagine Video v1.5 text and reference, MAI Image 2.5 Flash edit), and five prompt-free utilities: image and video upscalers, photo cleanup, and two lip-sync models. Every entry's scalar fields were derived from the `Input` schema AtlasCloud publishes for that model rather than hand-typed, so `sync-atlascloud-manifest.mjs --check` reports the manifest up to date; that same run applied 67 pre-existing drifts on already-shipped models, among them HappyHorse 1.1's `resolution`, whose `"1080P"` the API no longer accepts. Two shapes are deliberately excluded on the package's existing "a node surfaces one output" rule — `seedream-v5.0-pro/layer-decomposition` and the `*-sequential` models return a set that is not interchangeable — and the 3D and audio-output models have no modality here at all.

The script change is a correctness fix the new models surfaced: `enable_base64_output` / `enable_sync_mode` are marked `disabled` upstream on most models but **not** on the FLUX.1 open-weight ones, and either switch returns base64 instead of the URL the poll-then-download flow needs, so they are now suppressed by name. `output_dir` (a server-side path on Tencent's upscaler, meaningless to a NodeTool run) is suppressed too.

## Verification

**CI: 19 of 21 checks green** on the current head `cd10258b`, including `test-packages`, `typecheck`, `build`, `bundle`, `docker`, `test-app`, `app-build`, all three CodeQL analyses, and every E2E leg. `test-packages` is the one that matters most here — it runs the `runtime` and `atlascloud-nodes` suites on a runner provisioned with `mesa-vulkan-drivers ffmpeg`, so it covers the suites this sandbox cannot run (below) and it passed.

The two red checks are one failure, and it is not this PR's. `lint` fails on `anti-slop(no-conditional-empty-object-spread)` at `web/src/components/storyboard/ShotInspector.tsx:450`, present verbatim on `main` (it arrived with `fa28ff67`); `quality` is the rollup restating it, its log reading `changes:success static:failure build:success built:success docker:success`, and since `typecheck` passed, `static` is red only from `lint`. A verified patch is in [this comment](https://github.com/nodetool-ai/nodetool/pull/5373#issuecomment-5465605949), with a [correction](https://github.com/nodetool-ai/nodetool/pull/5373#issuecomment-5465671893) for generics that were stripped as HTML tags. An earlier, different base failure at `AssetViewer.tsx:297` ([reported here](https://github.com/nodetool-ai/nodetool/pull/5373#issuecomment-5465343994)) was fixed on `main` and picked up by merging `main` into this branch as `cd10258b`.

Local runs, for the record — this sandbox lacks binaries CI installs, so each gap was ruled out rather than left as a bare failure:

- `npm run test:affected` — the failure was `@nodetool-ai/runtime`, whose `tests/providers/video-frame-fallback.test.ts` dies at `spawn ffmpeg ENOENT`; that file predates this branch (`ed1c89fe`) and this diff touches no file under `packages/runtime`. The rest of the package: 163 of 164 files, 3110 tests passing. The AtlasCloud-relevant part run directly — `atlascloud`, `manifest-models`, `offline-model-index` — is 5 files, 191 tests, all passing, which is the check that exercises this manifest. CI's `test-packages` since confirmed the whole suite green.
- An earlier run stopped on `@nodetool-ai/image-nodes` with `No WebGPU adapter available (Node/Dawn)`. Installed the lavapipe ICD per AGENTS.md § WebGPU on a headless machine and re-ran that package alone: 24 files, 229 tests, all passing.
- Full parallel `test:affected` runs on this box also fail one *random* package per run (`chat` once, `deploy` the next), each of which passes standalone — `chat` 30/30, `deploy` 861/861 — and both pass in CI. That is this sandbox's concurrency limits, not the code.
- `npm run typecheck` — exit 0 locally (the 15 `TS2307` lines are `mobile/`, whose dependency tree is not installed here); green in CI.
- `npm run dev:nodetool -- harness gate --base main` — 0 surfaces touched, 0 selfchecks selected.

Checks specific to this diff:

- `node scripts/sync-atlascloud-manifest.mjs --check` → `atlascloud-manifest.json is up to date.` Before the sync it reported 67 drifts, every one against a **previously shipped** model — no new entry disagreed with the schema it was written from, which is what makes the generated fields trustworthy rather than plausible.
- `npm run test --workspace=packages/atlascloud-nodes` → 2 files, 81 tests passing, including the manifest invariants (`images` is always `list[image]`, `wrapInto` count unchanged at 9, no `return_last_frame`).
- Registry smoke test on the built package: 157 node classes register, 157 unique node types.
- The README node table was regenerated from the manifest; the generator reproduces all 97 existing rows byte-for-byte **and in the same order**, so the diff there is a pure insertion.

## Agent capabilities

## New checks

No new check, rule, or audit. The two added `SUPPRESSED_FIELDS` entries extend an existing filter rather than introducing a gate.

### DIFF
--- a/packages/atlascloud-nodes/AGENTS.md
+++ b/packages/atlascloud-nodes/AGENTS.md
@@ -44,6 +44,11 @@ Package specifics from shipped fixes:
   the *set* of layers it returns — is deliberately not shipped. A model that
   returns N interchangeable variants (`n`, `num_images`) is fine; one whose
   outputs are not interchangeable is not.
+- **Suppress an option by name, not by upstream's `disabled` flag.** AtlasCloud
+  marks `enable_base64_output` / `enable_sync_mode` disabled on most models but
+  not on the FLUX.1 open-weight ones, and either switch breaks the
+  poll-then-download flow. `SUPPRESSED_FIELDS` in the sync script is the list;
+  `output_dir` (a server-side path on Tencent's upscaler) is there too.
 - **The manifest is generated, not hand-edited.** `node
   scripts/sync-atlascloud-manifest.mjs` reconciles every entry's fields against
   the `Input` schema AtlasCloud publishes for that model (reachable from the
@@ -51,6 +56,14 @@ Package specifics from shipped fixes:
   without writing. Hand-tuned enums, wrong separators (`1024x1024` vs
   `1024*1024`) and stale option lists are what the script exists to prevent. Add
   a *model* by hand — the script only maintains the fields of models we ship.
+  Two shapes the script cannot fill in for a new entry, so write them yourself:
+  asset fields (it never touches `image`/`video`/`audio`/`list[…]` props), and
+  a schema property typed `anyOf` rather than a scalar — Cosmos 3 Super's
+  `image_size` is one, and hand-declaring it as an `enum` is what keeps the
+  control, since the script preserves an option list the schema no longer
+  carries. After adding entries, `--check` must report the manifest up to date;
+  a drift line against a model you just added means the entry disagrees with
+  the schema it was written from.
 - **Chat is not a node concern.** AtlasCloud's LLMs are OpenAI-compatible at
   `https://api.atlascloud.ai/v1` and are served by `AtlasCloudProvider` in
   `packages/runtime`, which extends `OpenAICompatProvider`. Only the async
--- a/packages/atlascloud-nodes/README.md
+++ b/packages/atlascloud-nodes/README.md
@@ -3,9 +3,10 @@
 AtlasCloud.ai nodes for [NodeTool](https://nodetool.ai).
 
 Run [AtlasCloud](https://www.atlascloud.ai) image and video models inside NodeTool
-workflows — GPT Image, Nano Banana, Seedream, Qwen Image, Reve, Youchuan and Wan
-for images; Seedance, Veo, Kling, Wan, MiniMax, HappyHorse and Grok Imagine for
-video.
+workflows — GPT Image, Nano Banana, Seedream, FLUX, Qwen Image, Ideogram, Reve,
+HiDream, Krea, Youchuan and Wan for images; Seedance, Veo, Kling, Wan, MiniMax,
+Vidu, PixVerse, LTX, HappyHorse and Grok Imagine for video; plus upscalers,
+photo cleanup and lip-sync as prompt-free utility nodes.
 
 Node fields are generated from AtlasCloud's published model schemas. Re-sync them
 with `node scripts/sync-atlascloud-manifest.mjs` (or `--check` to detect drift).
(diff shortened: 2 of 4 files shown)


georgi:
**`Quality Gate (npm run check) / lint` is red on `main`, not on this PR.**

The failing job is [99171901122](https://github.com/nodetool-ai/nodetool/actions/runs/33279397321/job/99171901122). Its only error:

```
web/src/components/assets/AssetViewer.tsx
  297:9  error  Use a theme.vars.palette.* token instead of a hardcoded hex/rgb color.  design-tokens/color-tokens
```

Why it is not this PR's: the diff is four files — `packages/atlascloud-nodes/{AGENTS.md,README.md,src/atlascloud-manifest.json}` and `scripts/sync-atlascloud-manifest.mjs` — nothing under `web/`. The violating line arrived with `13b89088` ("feat(storyboard): browse a board's stills and clips as a gallery") and is present verbatim on `main`'s current head (`fa28ff67`), so every open PR is red on this leg. No re-run: `lint` is deterministic, so a second run reports the same error.

No fix PR exists for it yet, so per the standing-down rule here is the patch rather than a push — it belongs on `main`, where it unblocks everyone, not buried in an AtlasCloud manifest PR:

```diff
--- a/web/src/components/assets/AssetViewer.tsx
+++ b/web/src/components/assets/AssetViewer.tsx
@@ -294,8 +294,7 @@
       pointerEvents: "none",
       zIndex: Z_INDEX.overlay,
-      background:
-        "linear-gradient(to top, rgb(0 0 0 / 0.72), rgb(0 0 0 / 0))"
+      background: `linear-gradient(to top, rgb(${theme.vars.palette.common.blackChannel} / 0.72), rgb(${theme.vars.palette.common.blackChannel} / 0))`
     },
```

`theme` is already the parameter of `styles` at line 75, and `AssetItem.tsx:50` one directory over uses `rgb(${theme.vars.palette.common.blackChannel} / 0.18)` in the same position — so this is the existing idiom, not a new one.

Verified by applying it locally and reverting: `npx eslint --config eslint.design.config.mjs src` in `web/` goes from `✖ 1 problem (1 error, 0 warnings)` to empty output, and `npm run typecheck --workspace=web` stays clean. This branch carries no part of that change.

---

georgi:
**Merged `main` in. The `AssetViewer` failure is fixed and picked up — but a different base failure took its place.**

`main` now carries `rgba(var(--palette-common-black-channel) / 0.72)` at `AssetViewer.tsx:297`, so [the failure I reported earlier](https://github.com/nodetool-ai/nodetool/pull/5373#issuecomment-5465343994) is resolved. Merged `origin/main` (`56471388`) into this branch and pushed as `cd10258b`; `npm run lint:design --workspace=web` passes here now.

On the merged tree, `npm run lint` fails on one new error instead:

```
web/src/components/storyboard/ShotInspector.tsx:450:7: error
anti-slop(no-conditional-empty-object-spread): This conditional spread hides property
omission behind an empty object. Build the object in separate statements and add the
property only when present.
```

Same reasoning as before, same evidence: that code is verbatim on `origin/main` (it arrived with `fa28ff67`, "feat(storyboard): edit a shot's title, description, camera and length"), and this branch's own commit `489cd818` touches four files — `packages/atlascloud-nodes/{AGENTS.md,README.md,src/atlascloud-manifest.json}` and `scripts/sync-atlascloud-manifest.mjs` — none under `web/`. Not this PR's, and no re-run: the rule is deterministic.

No fix PR exists for it, so the patch rather than a push:

```diff
--- a/web/src/components/storyboard/ShotInspector.tsx
+++ b/web/src/components/storyboard/ShotInspector.tsx
@@ -448,10 +448,11 @@
-    updateShot(boardId, shot.id, {
-      duration_seconds: seconds,
-      ...(linksLines ? { duration_source: "manual" as const } : {})
-    });
+    const patch: Parameters<typeof updateShot>[2] = {
+      duration_seconds: seconds
+    };
+    if (linksLines) {
+      patch.duration_source = "manual";
+    }
+    updateShot(boardId, shot.id, patch);
```

`Parameters<typeof updateShot>[2]` rather than `Partial<Shot>` because `Shot` is not imported in that file and this avoids adding an import for a one-line fix.

Verified by applying it and reverting: `npm run lint` goes from the error above to clean across `packages/*/src`, `web/src`, `electron` and `mobile/src`. I did not get a full `typecheck` run to complete on it before reverting, so treat the annotation as lint-verified but not type-verified. This branch carries no part of that change.

Rest of the merged tree re-checked: `sync-atlascloud-manifest.mjs --check` still reports the manifest up to date, `npm run test --workspace=packages/atlascloud-nodes` is 81/81, and the manifest still holds 157 entries.

---

georgi:
Correction to the patch in my previous comment: the generic type arguments were stripped as HTML tags when it was posted, so it renders as `Parameters[2]` and `Partial`, which is not valid TypeScript.

The annotation should be, with backticks so the angle brackets survive:

- `Parameters<typeof updateShot>[2]` — not `Parameters[2]`
- (the alternative I rejected was `Partial<Shot>`, which would need a `Shot` import that file does not have)

So the replacement block is:

```
    const patch: Parameters<typeof updateShot>[2] = {
      duration_seconds: seconds
    };
    if (linksLines) {
      patch.duration_source = "manual";
    }
    updateShot(boardId, shot.id, patch);
```

replacing the `updateShot(boardId, shot.id, { duration_seconds: seconds, ...(linksLines ? … : {}) })` call at `ShotInspector.tsx:448-451`. Everything else in that comment stands, including that `npm run lint` goes clean with it applied.

---

PR #5375, opened 2026-08-30
feat(projects): shoot the project views, and tell the landing page's agent section with them

georgi:
## What changed

**Screenshots and the data behind them.** The project surfaces shipped without pictures, and they are the one part of the app that shows nothing until documents are filed under a project — so this seeds three and captures them. `packages/websocket/src/screenshot-projects.ts` gives the screenshot backend SCRAPHEART (a trailer whose board has six stills and four takes, a script with voiced/stale/draft lines, a cut, and the agent conversation that built them), The Silent Abyss (stills rendered, nothing animated yet), and SINGULARITY (key art and little else), plus a ledger split across stills/clips/voice/pipeline. Three tests in `web/tests/benchmarks/screenshots.spec.ts` shoot the projects list, the new-project surface and the overview from the real UI at 1440×900 into `docs/assets/screenshots/`, and `docs/user-interface.md` gains a Projects section that uses them.

The pictures are frames the repo already ships — the SCRAPHEART trailer stills and takes behind the marketing site's trailer section, the six deep shots behind its documentary use case, the SINGULARITY poster — copied into the asset store through `storeAssetWithThumbnail`, the same call an upload makes. The screenshot server now points `ASSET_FOLDER` at a scratch directory, so a seeded row has bytes behind it and travels the real `/api/storage/<user>/<asset>.<ext>` route instead of rendering as a broken image.

**The landing page.** The agent section made the right claim — a chatbot leaves you a transcript, this leaves you a project — and then illustrated it with four chat panels, which is a transcript. It now follows one session end to end: `Ask` (the sentence that starts it), `Watch` (six keyframes rendered in the open), `Open` (the project those keyframes landed in), `Keep` (the shelf it goes on). Three of the four frames are the project views, the tabs are numbered so the arc reads without clicking, and the frames are one continuous story rather than four examples — the six stills the chat frame renders are the six on the board the project frames show. `ChatUISection`/`ChatShowcase` are replaced by `ProjectSection`/`ProjectShowcase`.

`web/scripts/build-project-shots.mjs` re-encodes the captured PNGs into `marketing/public/projects/` as WebP with the size manifest `next/image` needs. It is split from the capture because only the capture needs a browser, a backend and a seeded database; `--check` re-encodes into memory and fails on a drifted copy. The new-project shot is cropped to its top band — that surface pins its blank-document strip to the bottom of the viewport, leaving 270px of empty column at 900px.

**One product fix.** A `*Ref` carrying only `asset_id` — the shape `summarizeProject` returns for a sketch thumbnail — resolved to nothing. With no `uri`, `resolveStaticMediaUri` answers `""`, and `useResolvedMedia` ended resolution there rather than falling through to the asset lookup, so every sketch card in a project was blank.

Three chat shots (`chat-capability-map`, `chat-cost-preview`, `chat-trailer-delivered`) are no longer used by the landing page. Their casts and the `npm run chat-shots` pipeline stay — the fourth is still on the page, and the casts are the harness's own fixtures.

## Verification

- [x] `npm run test:affected` — 7 packages + web/electron, all green (`websocket` 233 files, `cli` 57, `deploy` 22, `reliability-harness` 27, `telegram` 9; web 3232 passed / 3 skipped; electron 686 passed). Re-run after the marketing commit selects "everything" only because `marketing/` and `AGENTS.md` map to no workspace; that commit changes no workspace source beyond two `web/package.json` script entries.
- [x] `npm run typecheck` — **web and electron clean; mobile cannot run in this sandbox.** `mobile/` is not a root workspace and its `node_modules` are absent here, so `typecheck:mobile` reports 439 module-resolution errors — the identical 439 on a stashed, unmodified tree. (An earlier revision of this description claimed a clean `npm run typecheck`; that was web's typecheck, not the root one.)
- [x] `npx tsc --noEmit` in `marketing/` (deps installed for the check) — clean. `npx next lint --dir src` — no new findings; the pre-existing ``/exhaustive-deps warnings only.
- [x] `npm run lint` — one pre-existing error only, `web/src/components/storyboard/ShotInspector.tsx:450` (`anti-slop(no-conditional-empty-object-spread)`, introduced in fa28ff67 on main); nothing from this diff.
- [x] `npm run dev:nodetool -- harness gate --base main` — `Gate: 7/7 selfchecks passed`
- [x] `FORCE_SCREENSHOTS=true npx playwright test tests/benchmarks/screenshots.spec.ts -g "Projects"` — 3 passed, the three PNGs in the diff
- [x] The section rendered from `marketing`'s dev server and captured per tab, to check each frame in place rather than only in isolation.

Both new checks were inverted once and observed failing.

The media-resolution fix ships with a reproduction; restoring `resolveStaticMediaUri(uri)` fails it:

```
● useResolvedMediaUri › resolves a ref that carries only an asset_id
  expect(jest.fn()).toHaveBeenCalledWith(...expected)
  at Object.<anonymous> (src/hooks/__tests__/useResolvedMediaUri.test.tsx:75:26)
Tests: 1 failed, 13 skipped, 14 total
```

And appending a byte to a committed WebP fails the freshness check:

```
Project shots are stale:
  · marketing/public/projects/project-list.webp differs from its source PNG
Re-run: npm run project-shots (in web/)
exit=1
```

## Agent capabilities

## New checks

- [x] `npm run project-shots:check` was inverted once and observed failing; the failing output is in **Verification** above
- [x] It cannot pass by matching nothing: it enumerates the three shots the landing page names and errors on a missing source PNG rather than skipping it

The three screenshot tests are captures, not checks, and each waits on seeded content by name (`SCRAPHEART — Trailer`, `from 2 past projects`, the board card), so an unseeded backend fails them rather than shooting an empty surface.


georgi:
`marketing-ci` → **Typecheck, lint & build** failed on `npm run seo:provider-catalog -- --check`:

```
[provider-catalog] marketing/src/data/providerCatalog.generated.ts is stale. Run: npm run seo:provider-catalog
```

**This is not this PR's failure.** The catalog is derived only from `packages/*-nodes/src/*-manifest.json` (per the generator's own header, "never hand-write this file"), and this branch touches none of them. The check fails identically on a clean `origin/main` worktree. The cause is on main: the atlascloud manifest gained 60 models in 489cd818 (Aug 29, "ship 60 more image and video models") while the catalog was last regenerated in 0139effa (Aug 28) — atlascloud reads 97 models where the manifest has 157.

**Ported the fix rather than waiting:** merged `main` and ran the repo's own generator, which is what the failing check instructs. It no-ops once main carries the same regeneration.

The whole job now passes locally, in its CI order:

- `gen:templates` + staleness guard — clean
- `seo:provider-catalog --check` — `up to date`
- `seo:check` — exit 0
- `npx tsc --noEmit` — clean
- `npm run lint` — no new findings
- `npx next build` — exit 0, 626 static pages

---

PR #5383, opened 2026-08-30
feat(agents): expose the live browser as an agent capability

georgi:
## What changed

The `browser_*` actions existed only as `AgentNode` tool classes, registered into the workflow tool picker by base-nodes. Nothing else could reach them — the chat agent, MCP, CodeAct, a Code node and a JS script assemble their belt from the capability registry, and no module declared a browser namespace — so the Chrome extension, whose whole point is driving the tab a user is already signed in to, was unreachable outside a graph. This adds a `browser` capability module declaring the thirteen actions plus a new `browser_status`. The actions need Chrome, CDP and the `/ws/extension` relay and live in `automation-nodes`, which depends on `agents`, so they arrive through a registration seam (`capabilities/browser-runner.ts`) that base-nodes calls at load — the same shape as `setExtensionChannelProvider` one layer down; a process that loaded no node packages answers every call with a sentence saying so instead of hanging. There is one implementation, not two: the tool classes are deleted and their persist and upload-resolve logic moves into `browser-actions.ts`, which both the capability and the `AgentNode` path (now resolving these names from the builtin belt) run through.

Two gaps the extension had are closed alongside. Nothing could ask whether an extension was attached, so an unattached relay showed up as a 30-second attach timeout on the first action — `browser_status` reports the transport, whether the bridge holds a socket, and what to do about it, and reports `null` rather than `false` where the process cannot answer. And the transport was process-global env config decided before the server started — `browser_restart` now takes `transport`, which is the only point it can change since the session is a process singleton.

Two claims in `docs/chrome-extension.md` read as shipped and are not: `asset_chunk` / `media_chunk` / `media_end` are declared in both protocol copies and implemented on neither side, and attach is not purely a user gesture (a host `attach` frame on an unattached relay attaches the active tab, which matters because `/ws/extension` is unauthenticated). Both are now documented as they are.

## Verification

- [x] `npm run test:affected` — all affected suites passed (63 electron suites / 686 tests, plus every affected package). The `base-nodes` image suites need a Vulkan ICD, which this sandbox lacks; installing lavapipe per AGENTS.md § WebGPU on a headless machine turns them green (31 files, 2006 passed).
- [x] `npm run typecheck` — web and electron clean. Mobile fails on `Cannot find module 'react-native'` etc: `mobile/node_modules` is absent in this environment (mobile is not a root workspace), unrelated to this diff.
- [x] `npm run dev:nodetool -- harness gate --base main` — 5/5 selfchecks passed (run as `harness gate --base HEAD~1` against this branch's commit).

Both new suites were inverted to prove they bite. Replacing `extensionConnected`'s body with `return true` failed two of them:

```
 FAIL  tests/browser-status.test.ts > browserStatus > says nobody is attached when the bridge holds no socket
 FAIL  tests/browser-status.test.ts > browserStatus > cannot answer for the extension outside the process holding the bridge
      Tests  2 failed | 3 passed (5)
```

## Agent capabilities

- [x] `npm run capabilities:check` passes — `capability table is current (248 capabilities)`.
- [x] All fourteen new capabilities are covered in `packages/cli/src/harness/capability-table.ts` by `packages/agents/tests/capabilities-browser.test.ts` under the `capability-suites` selfcheck. No gap notes were needed.

A new `live-browser` surface in the harness registry maps `chrome-extension/`, `packages/automation-nodes/src/lib/`, `packages/automation-nodes/tests/`, the three capability files, and `extension-cdp-bridge.ts` onto that selfcheck, so a diff to the relay runs it. The half below the seam — the extension → `chrome.debugger` → page round trip — stays in `npm run test:integration --workspace=packages/automation-nodes`, which needs Chrome and port 7777 and is off the gate; a comment on the surface says so.

## New checks

- [x] Inverted once and observed failing — the output is in **Verification** above.
- [x] Not applicable: neither suite is a file-scanning audit, so there is no "matched nothing" failure mode. `capabilities-browser.test.ts` asserts against the module's own export list and `BUILTIN_TOOL_NAMES`, both of which are non-empty by construction.

https://claude.ai/code/session_019gRAPvqAUexdbntZt9cuJN

---
_Generated by [Claude Code](https://claude.ai/code/session_019gRAPvqAUexdbntZt9cuJN)_


georgi:
`build-and-check` (Docs CI) failed here, and the failure is not this PR's:

```
* At ./_site/configuration.html:1399:
  internally linking to AGENTS.md#skills, which does not exist
```

`docs/configuration.md` is untouched by this branch — the only `docs/` file it changes is `chrome-extension.md`. The row came in with #5350, whose own `build-and-check` failed and was merged anyway, so Docs CI has been red on `main` since; every PR touching `docs/**` inherits it.

The cause is that `docs/configuration.md` is published while `docs/AGENTS.md` is not: `_config.yml` excludes it with the other contributor docs, so the link points at a page with no URL. I reproduced it locally with the workflow's own commands and confirmed a one-line fix clears it — `HTML-Proofer found 1 failure!` before, `HTML-Proofer finished successfully.` after.

That fix is #5386 against `main`, and I have ported the same commit here (c018ab36) so this PR can go green without waiting on it. It no-ops once `main` carries the change.

Worth someone's attention separately: nothing catches a published page linking to an excluded one until a later PR happens to run Docs CI.

---

PR #5406, opened 2026-08-30
refactor(websocket): extract DirectInferenceHandler from the runner

georgi:
## What changed

T5 of `docs/websocket-runner-refactor-plan.md`, stacked on the integration branch carrying T1 (#5405) and T2 (#5404) — base is `claude/websocket-runner-refactor-agents-u1ndgu`, not `main`. The one-shot model calls — the streamed `inference` command and the direct text / media / speech generation the sketch, timeline and Studio surfaces drive over RPC — move out of `UnifiedWebSocketRunner` into `DirectInferenceHandler` in `packages/websocket/src/session/inference.ts`, behind the `ClientSession` seam T2 introduced. The second half is the `registerAbort` seam: `rpcAborts` stays on the host, since `stop` and disconnect must reach every in-flight provider call, but nothing outside the host touches the set any more — a caller registers an `AbortController` and gets its deregistration back. T6 takes the same dependency when the command switch moves, so the set keeps one owner. The runner goes 8,829 → 8,003 lines; `session/inference.ts` is 963. Same frames in, same messages out, same abort semantics; no test file was edited.

Two things worth a reviewer's eye:

- The handler takes a third dependency the plan's sketch does not list, `currentRequestSeq: () => number`. `handleInference` cancels by comparing the seq it started with against the connection's live `chatRequestSeq` at every yield. That counter is chat state (T4's), so it is read through a callback rather than moved.
- `entityRefResolver`, `resolveEntityReferenceImages` and `retrieveSourceAssetBytes` — listed under T6 — sit inside the direct-generation region and use no `this`. They become module-level functions in `session/inference.ts`. `retrieveSourceAssetBytes` has no caller outside that region. The other two are also called by the chat media-generation path, so the runner keeps two one-line private methods delegating to them; the chat region's call sites are byte-identical, so T4 does not conflict here. The three `runDirect*` entry points likewise keep thin private delegations on the runner: `handleCommand` is T6's region and stays untouched, and `tests/direct-media-cost-row.test.ts` reaches `runDirectMediaGeneration` on the runner instance and must pass unmodified.

## Verification

Run from the T5 worktree. The root `npm run typecheck` was skipped because it compiles `mobile/`, which is not installed in this environment; `npm run lint --workspace=packages/websocket` is `tsc --noEmit` over the package that changed, and the root `npm run lint` (which passed) covers the rest.

- `npx vitest run --root packages/websocket` — **238 test files / 2621 tests passed**, matching the base commit's baseline exactly. `direct-text-spend.test.ts`, `direct-media-cost-row.test.ts` and the `unified-websocket-runner*.test.ts` suites all pass unmodified.
- `npm run lint --workspace=packages/websocket` (`tsc --noEmit`) — clean.
- `npx oxlint packages/websocket/src` — no findings in either changed file.
- `npm run lint:anti-slop:enforced` — exit 0.
- `npm run lint` — exit 0.
- `npm run nodetool -- harness gate --base origin/claude/websocket-runner-refactor-agents-u1ndgu` — exit 0: both changed files fall outside any registered surface, 0 selfchecks selected. (`--base main` cannot resolve a merge base in a worktree with no local `main`.)

- [x] `npm run test:affected` — run as `npx vitest run --root packages/websocket`, the suite the diff selects
- [x] `npm run typecheck` — as `npm run lint --workspace=packages/websocket` (`tsc --noEmit`); the root task needs `mobile/`, not installed here
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base main` — run against `origin/claude/websocket-runner-refactor-agents-u1ndgu`

## Agent capabilities

## New checks

---
_Generated by [Claude Code](https://claude.ai/code/session_01JrkYim9kcnJnvqhC8RWNoc)_

### DIFF
--- a/packages/websocket/src/session/inference.ts
+++ b/packages/websocket/src/session/inference.ts
@@ -561,7 +561,10 @@ export class DirectInferenceHandler {
         workflow_id: null,
         name: `${req.mode}_${Date.now()}`,
         content_type: contentType,
-        // Home — see the chat media generation path above.
+        // Home, the same folder an upload lands in — the rule
+        // `handleMediaGenerationMessage` in unified-websocket-runner.ts
+        // applies. A null parent is unreachable from the folder the
+        // asset browser opens on.
         parent_id: userId
       });
       const fileName = `${asset.id}.${ext}`;
--- a/packages/websocket/src/session/inference.ts
+++ b/packages/websocket/src/session/inference.ts
@@ -1,0 +1,963 @@
+import { randomUUID } from "node:crypto";
+
+import { createLogger } from "@nodetool-ai/config";
+import { getModelUnitPrice } from "@nodetool-ai/model-pricing";
+import { Asset, Prediction } from "@nodetool-ai/models";
+import { extractPricingParams } from "@nodetool-ai/node-sdk/pricing-params";
+import { resolveNodetoolDelegate } from "@nodetool-ai/protocol";
+import {
+  calculateChatCost,
+  detectImageMime,
+  expandEntitiesForGeneration,
+  generateStructured,
+  IMAGE_MIME_TO_EXT,
+  messageText
+} from "@nodetool-ai/runtime";
+import type {
+  BaseProvider,
+  ImageModel as ProviderImageModel,
+  ImageToImageParams,
+  ImageToVideoParams,
+  InpaintingParams,
+  Message as ProviderMessage,
+  MessageContent,
+  PromptAssetRef,
+  ProviderTool,
+  TextToImageParams,
+  TextToVideoParams,
+  VideoModel as ProviderVideoModel
+} from "@nodetool-ai/runtime";
+
+import { admitSpend, releaseSpend, reserveSpend } from "../credit-gate.js";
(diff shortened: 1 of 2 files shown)


georgi:
Do not read this PR's check list as coverage. `test.yml` is `pull_request: branches: [main]`, and this PR targets `claude/websocket-runner-refactor-agents-u1ndgu`, so the Quality Gate — typecheck, lint, test-packages, test-app, bundle, docker, browser E2E — never triggered. The single green check is `Page Load Smoke`, which has no branch filter and says nothing about this diff.

What was actually run, locally at this head:

- `npx vitest run --root packages/websocket` — 238 files / 2621 tests, matching the base commit exactly
- `npm run lint --workspace=packages/websocket` (`tsc --noEmit`) — clean
- `npx oxlint packages/websocket/src`, `npm run lint:anti-slop:enforced`, `npm run lint` — all exit 0

The base is the integration branch on purpose: it keeps this diff to the one task, instead of carrying T1 (#5405) and T2 (#5404) along with it. Once those two merge I will retarget this PR to `main`, which both shrinks the diff to exactly this task and makes the full gate run. Please don't merge it on the strength of the current checks.

---

PR #5408, opened 2026-08-30
refactor(websocket): extract JobExecutionManager from the runner

georgi:
## What changed

T3 of `docs/websocket-runner-refactor-plan.md`, stacked on the integration branch that already carries T1 (#5405) and T2 (#5404) — base is `claude/websocket-runner-refactor-agents-u1ndgu`, not `main`.

`UnifiedWebSocketRunner` was one class doing six jobs, and `activeJobs` was mutated from three of them. Everything between "run_job arrived" and "terminal status persisted" — the admission gates, cost estimation, the concurrency queue, run start-up, the message stream, reconnect, cancel, and terminal-status persistence — now lives in `packages/websocket/src/session/job-execution.ts` as `JobExecutionManager`, behind the `ClientSession` seam T2 introduced. The runner drops from 8,829 to 6,600 lines; the new file is 2,532.

Behavior is unchanged: same frames in, same messages out, same queue and abort semantics. The two invariants the plan calls out move with their doc comments attached — the synchronous slot reservation (`startingJobs++` before `startJob`, with no `await` between the gate read and the increment, so two back-to-back `run_job` commands cannot both pass the cap) and the `slotCounters` leak accounting the reliability harness reads.

`deps` beyond `ClientSession`, and why each is there:

| dep | why the seam does not carry it |
|---|---|
| `beforeRunJob` | per the plan |
| `getMaxConcurrentJobs`, `getMaxConcurrentRunsPerWorkflow` | the settings resolution and its TTL cache stay on the host, per the plan's constructor |
| `defaultMaxConcurrentJobs`, `defaultMaxConcurrentRunsPerWorkflow` | `drainQueue` falls back to them in its `.catch()`, so it never blocks on a settings read |
| `sendToSocket` | `jobDeliveryTarget.deliver` must bypass the run-session interception `send` does — routing it through `send` would re-enter `resolveJobSession` and loop the frame back into the session that emitted it |
| `isSocketConnected` | `settleRunAgainstLostConnection` asks whether the socket is still there; `ClientSession` exposes no socket |
| `attachPlanApproval` | chat owns it (T4); `startJobInner` attaches it to a run's context |
| `defaults: {provider, model}` | `startJobInner` seeds the run's model selection. Snapshotted by value, which is safe because the constructor is the only writer — `defaultModel` / `defaultProvider` are now `readonly` so the compiler holds that invariant rather than convention |

### Transitional delegators on the runner

Chat, RPC and the command switch still call the job members on the runner. T4 (chat) and T5 (direct inference) are being extracted in parallel from the same base, so rather than rewrite their call sites here, the runner keeps thin forwarding methods. Each is marked transitional and is removed in T7/T8 when those call sites move onto `JobExecutionManager` itself: `runJob`, `reconnectJob`, `cancelJob`, `getStatus`, `getSdkExecutionCapacitySnapshot`, `slotCounters`, `resolveJobSession`, `resolveJobControl`, `activeJobs`, `jobQueue`, `jobDeliveryTarget`, `inFlightJobCount`, `countActiveJobsForWorkflow`, `drainQueue`, `startJob`, `streamJobMessages`, `emitBeforeRunFailure`, `getRawGraph`, `normalizeGraph`, `hydrateGraph`, `inferOutputType`, `_handleNodeProviderCost`, `runMeasuredCost`. Every one forwards the live object, not a copy.

### D2 is only partly done here — read it as a deferral

The plan puts the D2 call-shape change ("`activeJobs` gets one owner; chat and RPC go through `resolveJobControl`/`getStatus`") in T3. This PR does the move but **not** that change. `activeJobs` is public on `JobExecutionManager`, and the runner's transitional getter hands the same map to the call sites that already reached for it: chat's `handleWorkflowMessage` registers and deletes its own workflow run, `stop` reads an entry and mutates `active.status`, `stream_input` logs the key set.

That is deliberate. Leaving those call sites untouched is what lets the chat and inference branches be developed against the same base and merged on top of this without conflict. It is a deferral, not a defect: the getter returns the live map, so behavior is identical, and the map still has exactly one owner in the sense that no second copy exists. **`activeJobs` actually becomes private in T8**, when the remaining regions are out of the runner and their call sites move onto `resolveJobControl` / `getStatus`. Until then, do not read D2 as complete.

### Moved module-level declarations

`RunJobRequest`, `RunJobExecutionOptions`, `DEFAULT_RUN_JOB_EXECUTION_OPTIONS`, `resolveRunJobExecutionOptions`, `resolveRunJobUserId` and `SdkExecutionCapacitySnapshot` moved with the code and are re-exported from `unified-websocket-runner.ts`, so every existing import path keeps working. `ActiveJob` and `createRelayActivityWaiter` are now exported because chat's `handleWorkflowMessage` still uses them.

One test file changed. `application-budget-gate.test.ts` patched `startJob` and spied on `estimateRunCost` on the runner instance; both now live on the manager, so the three sites target `asAny(runner).jobs`. The assertions are unchanged. The `unified-websocket-runner*.test.ts` suites pass unmodified.

## Verification

- [x] `npx vitest run --root packages/websocket` — **238 files / 2621 tests passed**, matching the base
- [x] `npm run lint --workspace=packages/websocket` (`tsc --noEmit`) — clean, including the two new `readonly` fields
- [x] `npx oxlint packages/websocket/src` — exit 0, 0 errors (the same 4 pre-existing `no-useless-assignment` warnings)
- [x] `npm run lint:anti-slop:enforced` — exit 0
- [x] `npm run lint` — exit 0
- [x] `npm run nodetool -- harness gate --base origin/claude/websocket-runner-refactor-agents-u1ndgu` — exit 0, but it reports **0 surfaces / 0 selfchecks**: `packages/cli/src/harness/registry.ts` lists no surface owning `unified-websocket-runner.ts`, so this diff selects nothing. The plan expected it to select the Ring 0 journeys; it does not. Ran them by hand instead (below).
- [ ] `npm run typecheck` — skipped: it compiles `mobile/`, which is not installed in this environment.

### Ring 0 reliability journeys

The plan makes these T3's verification because the `slotCounters` leak invariant is this task's. Run from `dist` after `npm run build --workspace=packages/websocket` and `--workspace=reliability/harness` (the harness's own `dist` was missing in a fresh checkout, which is what makes these unreproducible without building first):

| journey | kernel | ws-server | verdict |
|---|---|---|---|
| `linear-text-pipeline` | completed | completed | **OK** |
| `mid-run-cancel-node` (cleanup-leaks) | cancelled | cancelled | **OK** |
| `mid-run-cancel-streaming` (cleanup-leaks) | cancelled | cancelled | **OK** |
| `provider-failure-mid-stream`, all 7 provider faults | completed | completed | **OK** |
| `python-node-workflow`, all 5 bridge faults | failed (expected) | failed (expected) | **OK** |
| `ws-transport-faults`, all 5 ws faults | completed | timeout | **FAIL — pre-existing** |

The two `cleanup-leaks` journeys are the ones that read `slotCounters`; both diff clean against the kernel oracle with the manager owning the counters.

`ws-transport-faults` fails identically on the base commit. I confirmed rather than assumed: stashed the change (including the untracked new file), rebuilt `packages/websocket`, re-ran, and got the same `ws-server: timeout — waiting for a terminal job_update … verdict: FAIL`, then restored and rebuilt. Its three black-hole faults (`ws-drop-no-fin`, `ws-stall-reads`, `ws-abrupt-close`) deliberately make the client never observe a terminal frame — the journey's own description says so, and the per-fault expectations live in `reliability/harness/tests/journeys/ws-transport-faults.test.ts` rather than in the CLI verdict.

## Agent capabilities

## New checks

### DIFF
--- a/packages/websocket/src/session/job-execution.ts
+++ b/packages/websocket/src/session/job-execution.ts
@@ -1,0 +1,2532 @@
+/**
+ * Everything between "run_job arrived" and "terminal status persisted": the
+ * admission gates, the concurrency queue, run start-up, the message stream,
+ * reconnect, cancel, and terminal-status persistence.
+ *
+ * Extracted from `UnifiedWebSocketRunner` (T3 of
+ * docs/websocket-runner-refactor-plan.md). It knows the connection only
+ * through {@link ClientSession}; what the seam does not carry arrives as
+ * {@link JobExecutionDeps}.
+ */
+import { randomUUID } from "node:crypto";
+import { ApiErrorCode } from "../error-codes.js";
+import { admitSpend, releaseSpend, reserveSpend } from "../credit-gate.js";
+import { JobConcurrencyQueue } from "../job-queue.js";
+import {
+  isBoolean,
+  isFiniteNumber,
+  isFunctionValue,
+  isNonEmptyString,
+  isNumber,
+  isObjectLike,
+  isRecord,
+  isString
+} from "../lib/wire-values.js";
+import { Graph, withExplicitNodeFlags } from "@nodetool-ai/kernel";
+import {
+  ExecutionSession,
+  isExecutionPreflightError,
+  isUnitBilledCapability,
+  priceGeneration,
+  toRawGraphInput
+} from "@nodetool-ai/execution";
+import { createRunSupervisor } from "../run-supervisor.js";
+import {
+  jobRunRegistry,
+  type JobRunExecutionHooks,
+  type JobRunSession
+} from "../job-run-registry.js";
+import {
+  Application,
+  Asset,
+  Job,
+  listApplicationVersions,
+  invocationIdInUse,
+  releasedApplicationRelease,
+  releasedApplicationVersion,
+  reserveInvocation,
+  settleInvocation,
+  Workflow
+} from "@nodetool-ai/models";
+import { getInstanceId } from "../lib/instance-id.js";
+import { requestRemoteJobCancel } from "../job-control.js";
+import {
+  estimateWorkflowCost,
+  nodeExpectedQuantity,
+  type WorkflowCostEstimateDetail
+} from "@nodetool-ai/node-sdk/cost-estimate";
+import { extractPricingParams } from "@nodetool-ai/node-sdk/pricing-params";
+import { getModelUnitPrice } from "@nodetool-ai/model-pricing";
+import type { ProcessingContext } from "@nodetool-ai/runtime";
+import { applicationReleaseResponse } from "@nodetool-ai/protocol/api-schemas/applications.js";
+import type {
+  GraphData,
+  HydratedGraphData,
+  NodeDescriptor,
+  ProviderCost,
+  SupervisorRunOptions
+} from "@nodetool-ai/protocol";
+import { confineRunRequest, isRunRefusal } from "../lib/app-session-scope.js";
+import { releaseBlockedReason } from "../lib/app-deployment-service.js";
+import { autoSaveAssets, primaryTextOutputName } from "./asset-autosave.js";
+import { createRuntimeContext } from "./model-interfaces.js";
+import { formatSanitizedError, sanitizeLargeText } from "./sanitize.js";
+import { createLogger } from "@nodetool-ai/config";
+import type { ClientSession } from "./client-session.js";
+
+const log = createLogger("nodetool.websocket.runner");
+
+const TERMINAL_JOB_STATUSES = ["completed", "failed", "cancelled", "error"];
+
+export interface RunJobRequest {
+  job_id?: string;
+  workflow_id?: string;
+  /** Allow this run to start even if its workflow already has a run in flight. */
+  concurrent?: boolean;
+  user_id?: string;
+  auth_token?: string;
+  /** Human-readable run title; persisted as the job name. */
+  job_name?: string;
+  params?: Record<string, unknown>;
+  graph?: {
+    nodes: Array<Record<string, unknown>>;
+    edges: Array<Record<string, unknown>>;
+  };
+  explicit_types?: boolean;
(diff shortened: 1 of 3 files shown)


georgi:
Two things a reviewer should not take on trust here.

**1. This PR's check list is not coverage.** `test.yml` is `pull_request: branches: [main]`, and this PR targets `claude/websocket-runner-refactor-agents-u1ndgu`, so the Quality Gate — typecheck, lint, test-packages, test-app, bundle, docker, browser E2E — never triggered. Once T1 (#5405) and T2 (#5404) merge I will retarget this to `main`, which keeps the diff at exactly this task and makes the full gate run. Please don't merge it on the strength of the current checks.

**2. `harness gate` is green here because it checks nothing.** The plan (`docs/websocket-runner-refactor-plan.md`, T3) says the gate selects the Ring 0 reliability journeys for this diff. It does not: `packages/cli/src/harness/registry.ts` has no surface owning `unified-websocket-runner.ts`, so the gate reports "3 file(s) outside any surface, 0 selfcheck(s) to run" and exits 0. That line in the plan should be corrected, and the registry arguably wants a surface for this file — neither belongs in this PR.

The journeys were run by hand instead, from `dist`: `linear-text-pipeline`, `mid-run-cancel-node`, `mid-run-cancel-streaming`, `provider-failure-mid-stream` and `python-node-workflow` all pass on both the kernel and ws-server surfaces, including the two `cleanup-leaks` journeys that read `slotCounters` — the invariant this extraction owns. `ws-transport-faults` fails, and it fails identically on the base commit with the change stashed and rebuilt, so it is pre-existing and not this diff's.

Locally at this head: `npx vitest run --root packages/websocket` = 238 files / 2621 tests (baseline exactly), `tsc --noEmit` clean, oxlint / anti-slop-enforced / root `npm run lint` all exit 0.

---

PR #5428, opened 2026-08-31
perf(marketing): stop the model marquees running off screen

georgi:
## What changed

The mobile hamburger menu was reported as laggy, and it isn't the menu. A devtools trace of the landing page sitting idle at 4x CPU throttle on a phone viewport spends **1359ms of every 3s in style recalculation** with the menu closed and nothing happening; with the menu open it measures the same (1268ms). The tap just lands on a main thread that is already saturated. Bisecting that recalculation names the three infinite marquees in `ModelSupportSection`, which ran at all times regardless of where the section was. They now run only while the section is on screen, and not at all under `prefers-reduced-motion` — the component already took a `reducedMotion` prop it never used. Two corrections on the menu itself came out of the same investigation and are labelled as correctness, not speed, because they measured neutral: the scroll lock now pins the body (`overflow: hidden` alone is ignored by iOS Safari, so a flick inside the panel scrolled the page behind it), and the overlay moves to `z-[70]` because the announcement bar is `z-[60]` and was painting over the open panel.

Idle style recalculation per 3s sample, landing page, 4x CPU throttle, 390x844, section off screen:

| variant | idle style recalc |
|---|---|
| as shipped | **1315ms** |
| marquees disabled | **87ms** |
| `will-change: transform` on the tracks | 1271ms |
| `translate3d` keyframes | 1269ms |
| cards' `transition` removed | 1242ms |
| cards' `backdrop-blur` removed | 1241ms |
| section glows hidden | 1238ms |
| `scroll-fade` (`animation-timeline: view()`) disabled | 1207ms |

Chrome will not composite these animations and nothing done to the elements changes that — only stopping them removes the cost, which is why the fix is an `IntersectionObserver` rather than a compositing hint.

## Verification

`marketing/` is not a root workspace and none of the monorepo checks reach it — root `lint` is `oxlint packages/*/src web/src electron mobile/src`, `typecheck` is web + electron + mobile, and `npm run dev:nodetool -- affected` cannot run here without a full `build:packages` for a diff that touches no package. The checks below are the ones that cover this diff.

- [x] `npx tsc --noEmit -p tsconfig.json` (in `marketing/`) — clean
- [x] `npx next lint --file src/components/ModelSupportSection.tsx --file src/components/SiteHeader.tsx` — `✔ No ESLint warnings or errors`
- [ ] `npx playwright test tests/e2e/mobile-menu.spec.ts` — results and the inverted-run output to follow in a comment
- [ ] `npm run test:affected` / `npm run typecheck` / `npm run lint` — n/a, see above
- [ ] `npm run dev:nodetool -- harness gate --base main` — n/a, no surface in the harness registry covers `marketing/`

## Agent capabilities

## New checks

`marketing/tests/e2e/mobile-menu.spec.ts` asserts both mechanisms rather than a frame time: the page stays pinned while the menu is open and its scroll position is restored on close, and the marquee's computed `animation-play-state` is `paused` at the top of the page and `running` once its section is scrolled into view.

- [ ] The check was inverted once and observed failing — output to follow in a comment
- [x] Both assertions read live computed state, so neither can pass by matching nothing

---
_Generated by [Claude Code](https://claude.ai/code/session_01FwVFPovWAwem4U2zSanWRJ)_

### DIFF
--- a/marketing/src/app/globals.css
+++ b/marketing/src/app/globals.css
@@ -532,3 +532,8 @@
 .legal-article ul li::marker {
   color: rgb(100 116 139);
 }
+
+/* A flick past the end of the mobile menu must not chain to the page behind. */
+.mobile-menu-panel {
+  overscroll-behavior: contain;
+}
--- a/marketing/src/components/ModelSupportSection.tsx
+++ b/marketing/src/components/ModelSupportSection.tsx
@@ -1,5 +1,5 @@
 "use client";
-import React from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { motion } from "framer-motion";
 import { Cpu, Zap, Layers, Box, Sparkles, Cloud, Bot, ShieldCheck } from "lucide-react";
 import {
@@ -69,10 +69,34 @@ const frontierModels = [
 export default function ModelSupportSection({
     reducedMotion = false,
 }: ModelSupportSectionProps) {
+    const sectionRef = useRef<HTMLElement>(null);
+    // Chrome cannot hand these marquees to the compositor, so every frame they
+    // run costs a style recalculation over ~50 moving cards — measured at 1.2s
+    // of the main thread per 3s on a 4x-throttled phone, whether or not the
+    // section is on screen. That is the tax that made the rest of the page
+    // (the hamburger menu among it) feel stuck. Run them only while visible.
+    const [onScreen, setOnScreen] = useState(false);
+
+    useEffect(() => {
+        const section = sectionRef.current;
+        if (!section || reducedMotion) return;
+        const observer = new IntersectionObserver(
+            ([entry]) => setOnScreen(entry.isIntersecting),
+            { rootMargin: "200px" }
+        );
+        observer.observe(section);
+        return () => observer.disconnect();
+    }, [reducedMotion]);
+
+    const marqueesRunning = onScreen && !reducedMotion;
+
     return (
         <section
+            ref={sectionRef}
             aria-labelledby="model-support-title"
-            className="relative py-16 overflow-clip-safe"
+            className={`relative py-16 overflow-clip-safe${
+                marqueesRunning ? "" : " marquees-paused"
+            }`}
         >
             {/* Background Glow */}
             <div className="absolute top-1/2 left-0 -translate-y-1/2 w-[800px] h-[500px] bg-emerald-900/20 blur-[120px] rounded-full pointer-events-none" />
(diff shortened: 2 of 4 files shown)


georgi:
Verification is complete. Filling in the two boxes the description left open.

### The fix, measured on the built bundle

Same devtools trace as the description, re-run against a production build of this branch. Traced render work per 3s sample, 4x CPU throttle, 390x844, model section off screen:

| sample | before | after |
|---|---|---|
| idle, menu closed — style recalc | 1359ms | **80ms** |
| idle, menu open — style recalc | 1268ms | **81ms** |
| drag inside the open menu — style recalc | 2751ms | **113ms** |
| drag inside the open menu — total traced work | 3998ms | **1156ms** |

### `tests/e2e/mobile-menu.spec.ts` passes

Against a production build of this branch:

```
  ✓  1 … the menu pins the page while open and restores the scroll position (2.0s)
  ✓  2 … the model marquees stay paused while their section is off screen (1.0s)
  2 passed (4.0s)
```

CI ran it too — the **Playwright smoke tests** job on `2428f73` reports `363 passed (7.0m)`, and all 17 check runs on that commit are green.

### Both checks were inverted and observed failing

Reverting only `ModelSupportSection.tsx` and `SiteHeader.tsx` to their pre-fix state, rebuilding, and serving that bundle (`curl … | grep -c marquees-paused` → `0`, confirming the served HTML is the pre-fix one) fails both, each for its own reason:

```
  ✘  1 … the menu pins the page while open and restores the scroll position (2.1s)
  ✘  2 … the model marquees stay paused while their section is off screen (5.2s)

  1) … the menu pins the page while open and restores the scroll position
    Error: expect(received).toBe(expected)
    Expected: "fixed"
    Received: ""
      > 35 |     expect(await page.evaluate(() => document.body.style.position)).toBe(

  2) … the model marquees stay paused while their section is off screen
    Error: expect(received).toBe(expected)
    Expected: "paused"
    Received: "running"
    - Timeout 5000ms exceeded while waiting on the predicate
      > 66 |       .toBe("paused");

  2 failed
```

Worth recording that my first attempt at this step **passed**, because it undid the fixes at runtime from an init script and the strip never took effect — a green inversion is not a proof, so I discarded it and reverted at the source instead. A stale `next start` also held port 3001 on the first retry, which would have served the *fixed* bundle to a run labelled pre-fix; the `grep -c marquees-paused` check above exists to rule that out.

Updated checklist:

- [x] `npx playwright test tests/e2e/mobile-menu.spec.ts` — 2 passed; also green in CI's Playwright smoke job
- [x] The check was inverted once and observed failing — output above

---

PR #5432, opened 2026-08-31
feat(projects): start a project from a skill

georgi:
## What changed

The new-project surface offered five hardcoded shapes ("30s spot", "Trailer", "Music video", "Mini app", "Empty project"), each carrying a brief written in TypeScript. Skills are the same thing, extensible and already in the product: saved instructions the agent loads when a turn names them with a leading slash. The starter row is now the skill list — the user's own rows plus the five NodeTool ships (`launch-commercial`, `commercial-beat-sheet`, `explainer-storyboard`, `music-video-treatment`, `caption-titles`). Picking one prefixes the project's opening turn with `/skill-name`, the exact slash command the composer sends, so the agent reads that skill's instructions before it plans anything; the picked skill is stored as the project's `kind`, which is what `estimateFromHistory` reads spend back by. Picking nothing, or picking the same chip twice, sends the prompt alone. The chain preview is replaced by the picked skill's own description, and `projectShapes.ts` becomes `projectStarters.ts`.

`skills.list` gains an `includeSystem` flag, off by default: the shipped skills carry no row, so the panel that renames and deletes rows must not see them, while the surfaces that only *name* a skill can. They come back with `system: true` and a `system:skill-name` id, and a user row of the same name shadows the shipped one — matching how `load_skill` and the chat catalog already resolve it. The composer's slash menu opts in too; those names already resolved server-side in a turn, the menu just could not complete them.

## Verification

- [x] `npm run test:affected` — exit 0. 104/104 package tasks, web 1223 suites / 13888 tests, electron 63 suites / 686 tests, mobile 77 suites / 1030 tests.
- [x] `npm run typecheck` — web and electron clean. Mobile reports 36 errors in this container (missing Expo/React Native type packages); byte-identical with and without this diff, verified by `diff` against a `git stash` baseline.
- [x] `npm run dev:nodetool -- harness gate --base main` — `Gate: 8/8 selfchecks passed` (run from `dist` as `npm run nodetool`).

Three environment gaps had to be closed first, all documented in AGENTS.md and unrelated to the diff: `npm install` to link `@nodetool-ai/browser`, `mesa-vulkan-drivers` for the WebGPU image nodes, and `ffmpeg` for `runtime`'s video-frame fallback suite.

## New checks

Two router tests cover the merge — that the shipped skills appear only when asked, and that a user row shadows one of the same name. Inverted once (`if (!input.includeSystem)` → `if (true)`, so the flag is ignored) and observed failing:

```
 × merges the shipped skills into the list only when asked 10ms
 × lets a user row shadow the shipped skill of the same name 6ms
      Tests  2 failed | 5 passed (7)
```

Restored, and the suite is back to 7 passed. The first test asserts `shipped.length > 0`, so it cannot pass on an empty catalog.

## Agent capabilities

No capability added and no declared contract changed — `npm run capabilities:check` reports the table current (248 capabilities).

https://claude.ai/code/session_01CSyZbgkJ1vDQNziC5ikmgq


georgi:
`Run workflow suite in browser` (e2e-runner.yml, [job 99423636487](https://github.com/nodetool-ai/nodetool/actions/runs/33371562487/job/99423636487)) failed on `1fe4ec1a`, and the failure is not this PR's.

**What failed:** `manifest should list workflows` — `window.__E2E__.manifest().length` was 0 after the spec's 90 s `waitForFunction`. The job's own log shows the fixture side was fine: `Prepared suite with 194 workflows`, `[e2e-server] Ready … (2823 nodes registered)`. The test consumed the whole 90 s wait and then failed (1.6m total).

**Why it is not this PR's:** the manifest is a static `fetch("/e2e-suite/manifest.json")` that `prepareSuite()` writes into `web/public/`, and it is read only after `entry.tsx` → `E2ERunnerApp` finishes loading. That module graph is node-editor components, stores and contexts; every importer of the files this PR changes (`NewProjectSurface`, `projectStarters`, `hooks/skills/useSkills`, `composer/useTextareaSkillMention`) is a workspace or chat surface, none of them reachable from that entry. The PR's backend changes are `skills.list` and one Zod field, neither of which the harness calls.

Reproduced on this exact commit locally — the full suite passes, all 194 workflows:

```
[e2e globalSetup] Prepared suite with 194 workflows
[e2e globalSetup] Backend ready on http://127.0.0.1:7777
  ✓  1 [chromium] › tests/e2e-runner/suite.spec.ts:18:1 › workflow suite executes and records artifacts (1.9m)
  1 passed (2.1m)
```

`Workflow Runner Browser E2E`, the other browser job on this PR, is green. Re-running the failed job once, which is the one re-run I'll spend.

**No fix exists to port**, so I am not widening the PR with one. The proposed patch, for whoever owns this harness: the 90 s wait in `web/tests/e2e-runner/suite.spec.ts:31` covers the browser's first load of an uncached Vite dev graph, and this PR touches `packages/protocol`, which invalidates Vite's optimized-deps cache and makes that first load slower. Either raise that timeout, or warm the graph before the assertion — request `/e2e-runner.html` once in `globalSetup` after the backend is up, so the compile is paid outside the wait.

---

PR #5443, opened 2026-08-31
fix(cost-estimate): name rows by node title, scope prices to the selected provider

georgi:
## What changed

The workflow cost panel labelled every row with its dotted node type, so a real graph read as four identical `nodetool.agents.Agent` rows over one repeated prefix; rows now carry the node's title (the name the user gave it on the canvas, else the node's registered metadata title, else the class name spaced out — `Image To Video`), with the full type kept as the row's hover text.

The same panel refused to price `nodetool.video.ImageToVideo` nodes that had already stated their model, duration and resolution, claiming *"the catalog prices this model per units, which has no fixed value per run"*. The cause: the FAL and kie scalar catalogs are keyed by bare endpoint id and were consulted for every provider, so an AtlasCloud `google/gemini-omni-flash/image-to-video` was answered by FAL's row for the identical id — and FAL bills that endpoint in "units", which the converter rightly refuses. Those catalogs now answer only for their own provider. Two further fixes follow from the same principle: a scalar row that cannot be converted no longer hides a GenSpend price for the same model (`krea/v2/large/text-to-image` sat unpriced next to a $0.06/image row), and a model the catalog carries only under *other* providers now comes back declined **naming them** rather than as a bare "price unknown" — a catalog gap someone can close.

The rule this settles: an estimate is always the price the selected provider publishes, or no estimate. A reseller's margin is its own, so no lookup ever falls back to another vendor's rate — a figure that is not the price of the run the node will make is worse than no figure, however it is labelled. (The first commit here did borrow across providers behind a disclosure; the second removes that. `nodetool`'s managed models still resolve through their declared delegate, which is a stated mapping, not a fallback.)

Measured against the shipped catalogs, with `{ seconds: 4, resolution: "1080p" }`:

| model | before | after |
|---|---|---|
| `fal_ai:krea/v2/large/text-to-image` | declined ("units") | `$0.06` (GenSpend flat, previously shadowed) |
| `atlascloud:google/gemini-omni-flash/image-to-video` | declined ("units" — FAL's row) | declined: no atlascloud price; the catalog prices it on fal_ai |
| `atlascloud:nvidia/cosmos-3-super/image-to-video` | `$0.20`, silently FAL's rate | no price |
| `fal_ai:google/gemini-omni-flash/image-to-video` | `$0.52` | `$0.52` (unchanged) |
| `atlascloud:google/veo3.1/image-to-video`, `atlascloud:openai/gpt-image-2/text-to-image`, `gemini:veo-3.1-generate-preview`, `fal_ai:fal-ai/veo3.1/image-to-video` | unchanged | unchanged |

The two rows that lose a number are the point of the change: neither figure was AtlasCloud's, and one of them was being shown as a real price with nothing saying so.

## Verification

Ran per-package rather than the full sweep; `npm run test:affected` was not used because this checkout is a fresh clone with no baseline to diff against.

- [x] `npx turbo run test` for `@nodetool-ai/model-pricing` (158), `@nodetool-ai/node-sdk` (1218, 2 skipped), `@nodetool-ai/websocket` (2815), `@nodetool-ai/agents` (3656, 5 skipped), `@nodetool-ai/execution` (295) — all pass
- [x] `npx vitest run --config vitest.config.ts scripts/__tests__` — 8 passed
- [x] `npx jest src/components/costs src/hooks` (in `web/`) — 1474 passed (168 suites)
- [x] `npx tsc --noEmit -p packages/model-pricing/tsconfig.json` — clean
- [x] `npx tsc --noEmit` in `web/` — no diagnostics in any touched file (the run reports pre-existing `TS2307` for node packages not built in this checkout)
- [x] `npx oxlint` over the touched paths, and the same paths under `.oxlintrc.anti-slop-enforced.json` — clean
- [ ] `npm run dev:nodetool -- harness gate --base main` — not run

CI on the first commit went green after a re-run: the one red leg was `reliability/harness`'s `client-reconnect-mid-run.test.ts` failing with `Hook timed out in 10000ms` in the `beforeEach` that binds a loopback UI server — no assertion, a path this diff does not touch, and it passes on this branch locally in 22 s. See the comment thread for the full trace.

New tests pinning the behaviour: model-pricing asserts that a model priced under one provider is *not* priced for another, that the decline names the providers the catalog does carry, that the FAL/kie scalar catalogs answer only for their own provider, and that an unconvertible FAL unit no longer hides a GenSpend price; node-sdk covers `humanizeNodeType` and the user-title → metadata-title → class-name fallback (including on an unpriced node, where the name matters most); the web panel and the graph hook cover the rendered title and its hover text.

## Agent capabilities

## New checks

### DIFF
--- a/packages/model-pricing/src/genspend-catalog.ts
+++ b/packages/model-pricing/src/genspend-catalog.ts
@@ -166,9 +166,10 @@ let offeringsByModelId: Map<string, GenspendOffering[]> | null = null;
  *
  * Resellers list the vendor's own id verbatim — AtlasCloud and fal both call
  * it `google/gemini-omni-flash/image-to-video` — so a model the catalog knows
- * under one provider is usually the same model under another. Callers use this
- * only when the exact `<provider>:<id>` key is untracked, and must say whose
- * rate they ended up quoting.
+ * under one provider is usually the same model under another. That is a fact
+ * to *explain* an untracked offering with, never one to price it with: a
+ * caller quotes `getGenspendPrice` for the provider it was asked about and
+ * uses this only to say which providers the catalog does carry.
  */
 export function getGenspendPricesByModelId(
   modelId: string
--- a/packages/model-pricing/src/genspend-catalog.ts
+++ b/packages/model-pricing/src/genspend-catalog.ts
@@ -151,4 +151,41 @@ export function getGenspendPrice(
   return genspendPricingCatalog.prices[`${provider}:${modelId}`] ?? null;
 }
 
+/** One provider's offering of a model, as {@link getGenspendPricesByModelId} returns it. */
+export interface GenspendOffering {
+  /** The NodeTool provider id the price belongs to. */
+  provider: string;
+  price: GenspendPrice;
+}
+
+/** Built once on first use: model id → every provider that publishes a price. */
+let offeringsByModelId: Map<string, GenspendOffering[]> | null = null;
+
+/**
+ * Every tracked offering of one provider-native model id, across providers.
+ *
+ * Resellers list the vendor's own id verbatim — AtlasCloud and fal both call
+ * it `google/gemini-omni-flash/image-to-video` — so a model the catalog knows
+ * under one provider is usually the same model under another. Callers use this
+ * only when the exact `<provider>:<id>` key is untracked, and must say whose
+ * rate they ended up quoting.
+ */
+export function getGenspendPricesByModelId(
+  modelId: string
+): readonly GenspendOffering[] {
+  if (!offeringsByModelId) {
+    offeringsByModelId = new Map();
+    for (const [key, price] of Object.entries(genspendPricingCatalog.prices)) {
+      const separator = key.indexOf(":");
+      if (separator <= 0) continue;
+      const id = key.slice(separator + 1);
+      const offerings = offeringsByModelId.get(id);
+      const offering = { provider: key.slice(0, separator), price };
+      if (offerings) offerings.push(offering);
+      else offeringsByModelId.set(id, [offering]);
+    }
+  }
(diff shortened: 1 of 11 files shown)


georgi:
`Quality Gate (npm run check) / test-packages` is red on `39cbd33a`, and `… / quality` is only the aggregator rolling that one leg up. Everything else on this head is green — lint, typecheck, bundle, build, deps, docker, app-build, test-app, both browser E2E legs, CodeQL.

The single failure is in `@nodetool-ai/reliability-harness`:

```
FAIL  tests/journeys/client-reconnect-mid-run.test.ts > journey 9: client reconnect mid-run (task D2)
Error: Hook timed out in 10000ms.
 ❯ tests/journeys/client-reconnect-mid-run.test.ts:156:3
     156|   beforeEach(async () => {
     157|     srv = createTestUiServer({ host: "127.0.0.1", port: 0 });
     158|     await srv.listen();
```

It is not an assertion — the `beforeEach` never finished binding a loopback UI server and its `WsFaultProxy` within 10 s, so the test body never ran. That journey exercises the WS reconnect path (`test-ui-server`, `WsFaultProxy`, `job-run-registry`); this diff touches `packages/model-pricing`, `packages/node-sdk/src/cost-estimate.ts` and `web/src/components/costs`, none of which that file imports, and the harness's other 26 test files passed in the same run.

Checked before standing down: the file passes on this exact branch head locally — `npx vitest run tests/journeys/client-reconnect-mid-run.test.ts` in `reliability/harness`, 1 passed, 22.4 s, test body 1.15 s. A setup step that takes ~1 s of real work and timed out at 10 s reads as runner contention, not a defect in this change, so there is no fix to port.

Spending the one re-run on `test-packages` now. If it fails a second time I'll treat it as real and dig in rather than re-running again.

---

PR #5461, opened 2026-09-01
feat(sketch): let agents put images and assets on sketch layers

georgi:
## What changed

Asking an agent to make a sketch and set its layers to images or assets left the layers empty, because no tool could do it. `edit_sketch` only moved layer structure around — and its own description told the model so ("Pixels are never read or written") — while the `ui_sketch_*` surface could generate a new image but not place an existing one. An `image` argument a model reached for was dropped silently, so nothing said why the canvas stayed blank. Placement now exists on both surfaces, by reference rather than by inlining pixels: the layer stores an `asset://` locator or a URL and the canvas resolves and draws it on load, which is the shape a sketch seeded from an asset already carried, so the document stays small. `edit_sketch` gains `set_layer_image {target, image, x?, y?, width?, height?}` and takes the same `image` on `add_layer`, refusing an asset id that resolves to nothing or belongs to another user *before* the write — storing one produces exactly the empty layer this fixes. `ui_sketch_place_image` does the same against a live editor, loading the image first so an unreachable one fails instead of leaving a blank layer behind. The `ntlayer:` layer-data codec moves into `@nodetool-ai/protocol` and the editor's serializer delegates to it, so the two surfaces cannot drift on the format the fix depends on.

## Verification

- [x] `npm run test:affected` — ran the suites it selects for this diff: `packages/agents` full (236 files, 3690 passed, 1 skipped), `packages/protocol` full (55 files, 988 passed), and web's sketch + tool trees (`src/components/sketch`, `src/lib/tools`, `src/hooks/sketch`, `src/stores/sketch` — 152 files, 2292 passed).
- [x] `npm run typecheck` — output is byte-identical to the same command on a stashed clean tree (442 pre-existing errors in this sandbox, all `TS2307` for packages it has not installed or built: `mobile/`'s react-native tree, `@nodetool-ai/base-nodes`, `@nodetool-ai/transformers-js-nodes`). No error in any file this diff touches.
- [x] `npm run lint` — 0 errors. Three the change introduced were fixed rather than waived: `anti-slop(no-conditional-empty-object-spread)` in `add_layer`'s return, and two `anti-slop(no-runtime-typeof)` in the new codec (now `isNumber`/`isString` from `packages/protocol/src/predicates.ts`, and the `typeof Buffer` probe replaced by `btoa`/`atob`, which are globals in both runtimes).
- [x] `npm run dev:nodetool -- harness gate --base main` — `Gate: 12/12 selfchecks passed`.

The new tests fail against the old code, which is what makes them a reproduction: "puts an asset on a layer" reports `applied: 0, failed: 2` without `set_layer_image`, since the op name is rejected.

## Agent capabilities

- [x] `npm run capabilities:check` passes (after `capabilities:sync`, whose whole diff is the two contract hashes below).
- [x] `create_sketch` and `edit_sketch` re-declare their contract; both already name `packages/agents/tests/capabilities-sketches.test.ts` and the `sketch-tools` eval in `capability-table.ts`, and both gained cases there — placement, a missing asset, another user's asset, and a scheme the canvas cannot load. `ui_sketch_place_image` is covered by `packages/agents/tests/sketch-tool-loop.test.ts` and `web/src/lib/tools/__tests__/sketchTools.test.ts`.

## New checks

- [x] The codec's format test was inverted once by changing `SERIALIZED_LAYER_DATA_PREFIX` to `"ntlayerX:"`, and observed failing:

```
 × round-trips an asset locator with its bounds 7ms
 × decodes the envelope the editor writes 3ms
 Tests  2 failed | 34 passed (36)
```

  The prefix was restored and the suite is green. That test decodes the literal bytes the editor's serializer produces rather than only round-tripping this module's own output, so a decoder that stopped matching the stored format — which would silently blank every saved layer — cannot pass it.

- [x] `missingAssetReferences` is asserted to actually find its target, not merely to return an empty list: one case gives it an id no asset backs and one gives it another user's asset, and both expect the named id back in the refusal.

## Note for review

The eval surface's `ui_sketch_place_image` records the locator it is given without fetching the bytes — that surface has no asset store, the same reason `ui_sketch_render_to_asset` is excluded from it — so it moves the layer's reference, not its pixels. The header comment says so.

---
_Generated by [Claude Code](https://claude.ai/code/session_018JqHxT3wEz9mTC9aRjmBYK)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1352,6 +1352,15 @@ lists that under `notSimulated`. Layer bitmaps stay opaque throughout.
 The same static check is exposed to agents through the **`validate_sketch`**
 tool: pass an inline `document` to check a sketch being built, or an
 `image_document_id` to validate a saved one (scoped to the requesting user).
+**`edit_sketch`** places an image on a layer as well as editing the layer
+stack: `set_layer_image` (or `image` on `add_layer`) points a layer at an asset
+id, an `asset://` locator, a data: URL or an http(s) URL, and the editor
+resolves and draws it on load — the same reference a sketch seeded from an
+asset carries, so nothing inlines a bitmap into the document. An asset id that
+resolves to nothing is refused rather than stored, because a stored one shows
+up as an empty layer. With an editor open, **`ui_sketch_place_image`** does the
+same against the live canvas.
+
 Agents also get the version history headlessly: **`list_sketches`**,
 **`create_sketch`** (a blank canvas, then `edit_sketch`),
 **`list_sketch_versions`**, **`get_sketch_version`** (read one snapshot's
--- a/packages/agents/src/capabilities/sketches.specs.ts
+++ b/packages/agents/src/capabilities/sketches.specs.ts
@@ -122,13 +122,20 @@ export const EDIT_SKETCH_SCHEMA: JsonSchema = {
       type: "array",
       description:
         'Operations in order. Each is {"op": <name>, ...arguments}: ' +
-        'add_layer {name?, type?: "raster"|"mask", index?}, ' +
+        'add_layer {name?, type?: "raster"|"mask", index?, image?, x?, y?, width?, height?}, ' +
         "remove_layer {target}, rename_layer {target, name}, " +
         "set_layer_props {target, visible?, locked?, opacity?, blendMode?}, " +
+        "set_layer_image {target, image, x?, y?, width?, height?}, " +
         "reorder_layer {target, index}, duplicate_layer {target}, " +
         "select_layer {target}, resize_canvas {width, height}. " +
         '`target` is a layer id, its name, or "active". Layer index 0 is the ' +
-        "bottom layer.",
+        "bottom layer. " +
+        "`image` puts a picture on the layer: an asset id (from list_assets), " +
+        "an asset:// locator, a data: URL, or an http(s) URL. It is drawn at " +
+        "its natural size from the top-left of {x, y, width, height}, which " +
+        "defaults to the whole canvas — pass the image's own dimensions when " +
+        "you know them. Pass `image` to add_layer to place a picture on a new " +
+        "layer in one op.",
       items: { type: "object" }
     }
   },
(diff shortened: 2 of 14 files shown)


georgi:
**E2E Workflow Runner / "Run workflow suite in browser" failed on 8471856** — `manifest should list workflows` (expected `> 0`, received `0`) at `web/tests/e2e-runner/suite.spec.ts:38`. I could not reproduce it, and I have re-run the job once. Here is what I checked rather than assuming a flake.

The manifest was built and the backend came up in that same job:

```
[e2e globalSetup] Prepared suite with 194 workflows
[e2e-server] Ready on http://127.0.0.1:7777 (2823 nodes registered)
[e2e globalSetup] Backend ready on http://127.0.0.1:7777
```

So the failure is the page's own 90 s `waitForFunction` for `window.__E2E__.manifest()`, which the test swallows before asserting. The test ran 1.6 min, consistent with that wait expiring.

The obvious suspicion was mine to rule out: this diff adds a new web import (`@nodetool-ai/protocol/api-schemas/sketch.js` into `web/src/components/sketch/serialization`), and a module-graph failure there would leave the app unmounted and the manifest empty — exactly this symptom. It isn't that:

- **Resolution.** Under the repo's Vite conditions (`nodetool-dev, import, module, browser, default`), `@nodetool-ai/protocol/api-schemas/sketch.js` resolves to `packages/protocol/src/api-schemas/sketch.ts` and its `../predicates.js` to `packages/protocol/src/predicates.ts`. Checked directly through Vite's plugin container, not inferred from the exports map.
- **Module evaluation.** Loading `/e2e-runner.html?manual=1` against a Vite dev server on this commit installs `window.__E2E__` with no page error and no module error — only the expected backend-absent tRPC noise.
- **The suite itself.** Ran end to end locally on this exact commit (`npx playwright test -c playwright.e2e-runner.config.ts`, chromium pinned to the box's build): manifest loaded, all 194 workflows executed, **1 passed (2.2m)**.

Against that, run 2212 on another branch passed this job at the same time on its own runner, so the harness was not broken repo-wide. The remaining explanation I can support is startup timing on that runner crossing the 90 s manifest wait — the dev server cold-starts and optimizes deps before the app can fetch `/e2e-suite/manifest.json`.

I am not raising that timeout in this PR: the threshold belongs to a test this change does not touch, and loosening it to get green would hide the next real occurrence. If the re-run fails the same way, that is a second data point and I will treat it as real rather than as a flake.

---

PR #5491, opened 2026-09-01
feat(timeline): motion graphics M1–M3 — animation depth, structure, text and shapes

georgi:
## What changed

Works through milestones M1–M3 of `docs/plans/motion-graphics-tasks.md`. The branch is being built task by task; this description is updated as each one lands, and the verification block below is re-run against the finished branch before this is ready to merge.

**Landed so far**

- **T6 — easing grammar (F2, D3).** `easing` is a string in the document, so the grammar extends without a schema change. `parseEasing` resolves the seven named ids plus `cubic-bezier(x1,y1,x2,y2)` (Newton with a bisection fallback, control x clamped to [0,1] to keep x(t) monotonic, y left free so overshoot survives) and `spring(stiffness,damping,mass)` (closed-form damped oscillator across the underdamped, critically damped and overdamped cases, time-scaled by an analytic settle time so f(0)=0 exactly and |f(1)−1| ≤ 1e-4). Parsing is memoized per string. A string nothing parses eases linearly and the validator reports `unknown_easing` as a warning, so a document from a newer build still plays (I2).

**In flight:** T7 (channels and fold metadata) → T8 (character/line stagger) → T9…T13 (the serial compositor chain: groups, precomposite, two-clip transitions, masks and mattes, the effect catalog) → T16, T14, T15, T17 (shapes, text style, caption style, fonts).

### Deviations from the plan, and why

- **T6's `easeInOut` test assertion is wrong in the plan.** `docs/plans/motion-graphics-tasks.md:270` asks that `cubic-bezier(0.42,0,0.58,1)` match this repo's `easeInOut` within 1e-3. Measured over 1000 points with an exact bisection solver, it is off by **0.084** at t≈0.654 — the repo's `easeInOut` is a piecewise cubic power curve (`4t³` / `1−(−2t+2)³/2`), not a bezier, and no cubic bezier gets within 1e-3 of it (the closest found, `cubic-bezier(0.66,0,0.34,1)`, still misses by 0.0084). The assertion was replaced with two that are analytic: the solver checked against the bezier polynomial evaluated directly at ten parameter values, and the curve's odd symmetry about (0.5, 0.5) — the property it does share with `easeInOut`. The plan line should be corrected separately.
- **`pop`'s default easing left unchanged.** The plan permits changing it to `spring(180,12,1)` if the existing `wipe`/`pop` tests still pass — they do. It was reverted anyway on evidence outside that condition: `web/src/components/timeline/Inspector/ClipAnimations.tsx` renders the easing control as a closed-option `InspectorSelect`, so the spring default logs an MUI out-of-range warning and shows an empty picker for every pop animation. T31 is what gives that control a free-text grammar field.

### Found, not fixed

- `ClipAnimations.tsx` cannot author or display a `cubic-bezier(...)` / `spring(...)` easing (T31 owns it).
- `packages/system-skills/motion-graphics/SKILL.md` does not mention the grammar or `unknown_easing` (T24).
- `packages/agents/src/capabilities/timelines.specs.ts` still documents `easing` as the named ids only; the `edit_timeline` / `ui_timeline_animate_clip` schemas were not widened (T19/T24).
- `normalizeCustomCurves`'s new `unknownEasings` is read by the validator only — `packages/agents/src/custom-animation-bake.ts` drops it, so an agent baking curves with a typo'd easing hears nothing until it validates.

## Verification

Re-run against the finished branch before merge. As of T6:

- [x] `npm run test --workspace=packages/timeline` — 24 files, 310 tests passed
- [x] `npm run test --workspace=packages/execution` — 37 files, 335 tests passed (needs `npm run build:packages` first; unbuilt siblings otherwise fail resolution, not assertions)
- [x] `npm run lint` — exit 0
- [ ] `npm run typecheck` — web and electron legs clean. The **mobile** leg fails on a clean tree in this container: `mobile/node_modules` is absent (mobile is deliberately not a root workspace), so `tsconfig.json` cannot resolve `expo/tsconfig.base`, `strict` drops, and ~448 unresolved-module and implicit-`any` errors cascade. None is in a file this branch touches.
- [ ] `npm run test:affected`
- [ ] `npm run dev:nodetool -- harness gate --base main`

## New checks

- [x] `unknown_easing` was inverted once (`parseEasing(easing) !== null` → `=== null` in `unknownEasingIssue`) and observed failing — 3 failed | 44 passed in `packages/execution/tests/timeline-debug.test.ts` — then restored.
- [x] The two cubic-bezier solver tests also failed for real before the premise above was corrected.

---

https://claude.ai/code/session_01VdzG3a3jN3Dzx6QV8xh4Bo

---
_Generated by [Claude Code](https://claude.ai/code/session_01VdzG3a3jN3Dzx6QV8xh4Bo)_


georgi:
**`Quality Gate (npm run check) / test-packages` is red on `5f85fef0`, and the failure is not this PR's.**

```
FAIL tests/run-budget.test.ts > RunBudget > keeps the first reason when a second ceiling is hit later
AssertionError: expected 'deadline' to be 'cost'
```

`packages/runtime/tests/run-budget.test.ts` does not exist on this branch — it comes from `main`'s side of the merge commit CI builds. This branch touches no file under `packages/runtime`, so there is no way for it to affect the result. I could not reproduce it locally for the same reason: neither the test nor the `RunBudget` source it exercises is present here.

**The mechanism.** The test is timing-dependent by construction:

```ts
const budget = createRunBudget({ ...options, capUsd: 0, deadlineMs: 10 });
expect(budget.turns.reserve(pricedTurn())).toBe(false);
expect(budget.exhausted?.kind).toBe("cost");   // ← fails here
```

`capUsd: 0` exhausts cost immediately and `deadlineMs: 10` expires the deadline 10 ms after construction. If more than 10 ms of wall clock elapses between those two lines, the deadline is already expired and the *first* exhaustion recorded is `deadline`, not `cost` — which is exactly the ordering the test exists to assert. The same job logged sandbox worker round trips of 3.5 s and 4.9 s, so the runner was heavily contended; a 10 ms budget is not survivable under that load.

**Proposed patch** (for whoever owns the `RunBudget` work — not made here, since it would widen this PR into a package it does not touch): drive the clock instead of racing it. Give the first phase a deadline that cannot expire during it, then advance time deterministically rather than sleeping past it — `vi.useFakeTimers()` with `vi.advanceTimersByTime(...)` between the two phases makes the ordering explicit instead of incidental. Raising `deadlineMs` alone only widens the window; it does not remove the race.

Re-running the failed job once to confirm it reproduces rather than sticks.

---

PR #5492, opened 2026-09-01
feat(agents): one permission gate on every host; retire `Agent`; compact long threads (A2, A3, A4)

georgi:
## What changed

Work packages **A2**, **A3** and **A4** from `docs/plans/agent-system-improvements.md`, following A1's `RunBudget`. A2 gives every host a permission gate the loops beneath it can find. A3 deletes the second agent loop so budgets, the gate, compaction and the scheduler stop being built twice. A4 lets a long chat thread keep working instead of dying on a context-window error the user can do nothing about.

**A2 — one permission gate on every host.** A gate now travels on the context bag, the same channel A1's budget uses, so a loop the host never constructed reads the host's gate instead of building an ungated run: an `AgentNode` reached through `run_node`, a JS script, a sub-agent. `ProcessingContext.copy()` shares it by reference, so a mid-turn `set_permission_mode` reaches a node that started before it. Absence means the headless gate — `auto`, denying whatever the ladder escalates — not "ungated". The CLI asks on a TTY (`y / n / a` on stderr, since stdout carries the result) and runs headless when piped. An audit pins which runs may still be ungated and asserts it found its targets, so it cannot pass by matching nothing.

**A3 — `Agent` is retired.** The class, `CompilerAgent`, the checkpoint store, the plan cache, the plan-approval callback and the graph branch are gone; `nodetool agent run` is one CodeAct session over the belt chat uses, with `create_plan`/`execute_plan` still on it. Every deletion followed an enumerated importer list rather than a guess — which is why `applyRunPolicy` survives (the `graph-e2e` eval imports it) while its three neighbours did not. `TaskPlanner`, the DAG executors, `SubAgentTool` and the supervisor stay.

**A4 — compaction and observation continuity.** A compaction record is a `role: "user"` row marked `execution_event_type: "compaction"`; history is assembled from the newest one forward. Keeping it a user message is what makes the cut safe — the boundary lands between turns by construction, so an assistant's tool call is never separated from its result. Two triggers fire it: a prompt estimate over a threshold, and the provider reporting the prompt did not fit, which compacts and retries once. Separately, the CodeAct finish nudge now carries the round's observations, so a model asked to `finish()` can see what it computed from.

## Correction to commit `021be0dd`'s message

That message says the token estimate "sees neither the system message, the tool definitions, nor images". Only the middle clause is true. `estimatePromptTokens` is handed `messagesToSend`, which is a copy of `chatHistory` after the system message has been prepended to it, and media is resolved before the check, so a resolved image counts as the length of its base64. The estimate is a size signal that omits the tool schemas, not a floor that omits three things. The code comments state this correctly; the commit message does not, and the branch is pushed, so this note is the correction rather than an amend.

## Verification

```
npm run build:packages                                    59/59 green
npm run lint                                              exit 0, design-token rules included
npm run dev:nodetool -- harness gate --base origin/main    exit 0, 6 surfaces, 6 selfchecks
npm run backend:smoke                                     exit 0, server.mjs booted, /health answered
```

Per-suite: `packages/cli` 693, `packages/agents` ~3700, `packages/websocket` 2846, `packages/runtime` 3240, `packages/llm-nodes` 515, `packages/execution` 332, `packages/config` 146, `web` chat-message suites 161.

`backend:smoke` matters for A3 specifically: a deleted module can break the packaged bundle in a way unit tests miss.

Every task's check was inverted and observed failing before it shipped. Three of those are worth naming, because each says something about the checks themselves:

- **The plan's inversion for the compaction cut could not fail.** It said to cut at a tool-result boundary and watch `repairOrphanedToolCalls` change its count. That function patches an assistant `tool_use` missing its `tool_result`; a suffix cut produces the opposite defect — a `tool_result` whose `tool_use` is gone — which it deliberately leaves alone. A real instrument (`danglingToolResultIds`) replaced it, with a separate test proving the repair still fires where it does apply.
- **One CLI inversion passed under inversion**, because the assertion looped over an array that was empty. The test was rebuilt to construct the run the way `LazyCapabilityTool` does, with a non-empty assertion, and re-run to get a genuine failure.
- **The "retry twice" inversion is unrunnable, so the other direction was inverted instead.** Removing the retry flag does not double the retries; each compaction rebuilds the kept slice into the same shape, whose next cut summarizes only the previous record, so it loops forever. The flag is set before the attempt, which also means a failed summarizer spends the retry rather than leaving a second one available.

- [x] `npm run test:affected` — run per package as above
- [x] `npm run typecheck` — web and electron clean; `mobile/` fails only on missing `node_modules`, which root `AGENTS.md` § Prerequisites says not to chase
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base origin/main`

## Agent capabilities

No capability is added and none changes its declared contract. `npm run capabilities:check` reports the table current. An earlier `harness gate --base main` flagged `find_model`; that was a stale local `main` ref sitting *behind* this branch's own starting point, and the contract change came from A1's merge. Re-run against `origin/main`, the gate is clean.

## New checks

- [x] Each was inverted once and observed failing
- [x] Each audit asserts it found its targets

## Known gaps, stated rather than left implicit

- **The interactive Ink session gates nothing.** `packages/cli/src/app.tsx` builds its own belt with no `gateTools` and no gate on its context. Ink reads stdin itself, so the stderr `readline` prompt the other CLI paths use cannot share the terminal; closing it needs an Ink component. `nodetool-chat` prints that `--permission-mode` does not apply there rather than accepting the flag and ignoring it. The gate table names this row.
- **`run_node`'s child inherits the gate but not the budget.** `runSingleNode` builds its own context and is now handed the turn's gate; the turn's `RunBudget` is not passed the same way, so an `AgentNode` reached through `run_node` opens a fresh cap. This violates invariant I-2, which names that case. It is a defect in merged A1 code, left untouched here because it is outside these three packages' scope.
- **`AgentPolicy`'s resolver is dead.** `resolveAgentPolicy()` and `AgentPolicyOptions` have no callers; only `DEFAULT_AGENT_POLICY.maxConcurrentAgents` is still read. The file stays because `RunBudget` carries no per-step iteration bound and D6 keeps per-step caps — folding it is A7's work.
- **Three crumbs for A8:** the `plan_approval_response` client half (protocol schema, inbound route, `PlanApprovalCard`) now has no server emitter; `AgentSpanKind` still has a `"compile"` member nothing produces; `docs/workflow-supervisor-design.md` §7 and `ARCHITECTURE.md` still describe `Agent` as the graph path.

## Two behavior changes worth knowing before merge

`nodetool agent run` builds its context with `createChatContext` now, so a generated image persists as an `asset://` rather than landing in the cwd, and the final text is the last assistant message rather than a compiler's step result. `--max-steps` is gone with the planner pipeline, and `nodetool eval --max-retries` was removed because it fed a field only the retired `TaskPlanner.plan()` read.

## A note on this branch's history

Commit `6a66b1a`, whose message describes a CLI docs update, also carries the file deletions belonging to the `Agent` retirement, and the two commits after it do not compile. The deletions were staged by the agent doing that work and were picked up by a commit made by path. The fix went forward rather than by rewriting, since the branch was already pushed and `AGENTS.md` §20 forbids rewriting published history. A squash merge collapses it; a merge preserving commits will carry one mislabelled commit and a two-commit window that does not build.

https://claude.ai/code/session_013JxNrqmgsa7wbAB5JfB363


georgi:
`Run workflow suite in browser` is failing on `ecb21c3`, and the failure is this PR's. It is not a flake and does not need a re-run — the branch does not compile at that commit.

**Cause.** Commit `6a66b1a`, whose message describes only a CLI docs update, also deleted `packages/agents/src/agent.ts`, `compiler-agent.ts`, `checkpoint-store.ts`, `tools/create-task-tool.ts` and six test files. Those deletions belong to the in-progress A3 work (T-A3.3) and were staged in the git index before that commit was made; committing by path picked up the whole index rather than the paths named. `packages/agents/src/index.ts` was left still exporting `Agent`, `CompilerAgent`, `CheckpointStore` and `CreateTaskPlanTool` from files that no longer exist, so every downstream package fails to build.

**Fix, in progress.** T-A3.3 completes the deletion: `index.ts` is already pruned, and one leftover remains — `evals/task-planner-eval.ts` still passes `maxRetries` to `TaskPlanner`, which no longer declares it. That option was only ever read inside `TaskPlanner.plan()`, and this suite calls `planMultiTask`, so it has never had an effect here; it is being removed rather than restored, together with the `nodetool eval --max-retries` flag that fed it.

**History.** The deletion will read across two commits, one of them mislabelled, because the branch is already pushed and this repo does not rewrite published history (`AGENTS.md` §20). The final state is what matters and it will build; a squash merge collapses the two anyway.

The next push carries the compiling tree and re-runs this check.

---

PR #5502, opened 2026-09-02
feat(timeline): render quality — alpha export, validator codes, inspector [M5]

georgi:
## What changed

Milestone 5 of `docs/plans/motion-graphics-tasks.md`. Three tasks land here — **T27** (alpha export and output formats, F13), **T30** (validator codes, F24) and **T31** (inspector UI, AS4's thin layer) — plus the DSL regeneration T27's new node props force and one CI fix. T28 (motion blur) and T29 (time remap) are still in flight and will be pushed to this branch as they land.

**T27** makes transparency a first-class output. `alpha: true` seeds both compositors transparent and keeps straight alpha through to the encoder: the GPU compositor un-premultiplies on readback, because the accumulation is premultiplied and every partly-transparent pixel would otherwise export darkened. A new pure `outputFormats.ts` holds the table — `webm` → `libvpx-vp9 -pix_fmt yuva420p`, `mov` → `prores_ks -profile:v 4444 -pix_fmt yuva444p10le` (422 HQ when opaque), `png_sequence` → no ffmpeg at all, `mp4 + alpha` refused by name. `pngSequence.ts` writes a PNG per frame through `@napi-rs/canvas` into a store-only zip built on `node:zlib`'s `crc32`, so no dependency and no lockfile churn, with a `manifest.json` carrying fps, size, count and alpha. Refusals fire before a single frame is composited, and the rough-cut fallback refuses to stand in for an alpha or non-MP4 request instead of silently returning opaque MP4. The browser export gains a format dialog built from `ui_primitives`.

**T30** names the failures the validator could already see but never reported. Five new codes — `animation_exceeds_clip`, `stagger_compressed`, `replace_curves_overlap`, `text_illegible`, `time_remap_not_monotonic` — read the compiled animation output, T8's stagger flag, the held intervals of a `replace` channel, type size against frame height plus contrast against whatever is actually behind the text, and the remap curve's monotonicity. Eight more codes already existed from T2/T6/T9/T11/T12/T13 and are reconciled rather than duplicated; `parent_missing` and `parent_not_group` drop from error to warning, since both render unparented. `font_not_portable` is not here — T17 shipped it on main in the same commit that added the font pipeline it needs.

**T31** puts the fields the renderer has honoured since M2 in front of a person. Eight inspector sections — transition, mask, matte, effects, text style, shape, parent (with a *group selected clips* action that writes one undo entry), and a custom-curve keyframe table — all through `patchClip`, all primitives, no raw MUI. The font picker is deferred: T17 had not landed when this was written, so `fontFamily` stays the free-text field the document carries. It is being added on a follow-up branch now that T17 is in main.

## Verification

Run on the merged tree, not per-task. **All 22 CI checks are green on `d8f92a99`**, including `test-packages`, `typecheck`, `lint`, `docker`, `bundle`, `app-build` and the `quality` roll-up.

Locally:

- [x] `npm run lint` — exit 0 (design gate, oxlint, and the enforced anti-slop config)
- [x] `npm run typecheck` — web and electron exit 0. The **mobile** leg fails on `mobile/node_modules` being absent in the dev container, which reproduces identically with the whole diff stashed. CI's typecheck leg passes.
- [x] `npm run test --workspace=packages/sandbox-compiler` — 208 passed, after the regeneration below
- [x] `npm run test --workspace=packages/timeline` — 541 passed
- [x] `npm run test --workspace=packages/video-nodes` — 239 passed, and see the CI fix below
- [x] `npm run test --workspace=packages/execution -- timeline` — 136 passed across 11 files
- [x] web jest, full suite — 1257 suites, 14167 passed
- [ ] `npm run dev:nodetool -- harness gate --base main` — `--base main` exits with `no merge base` on this branch. Against the real base commit, `harness gate --base 78e69899 --dry-run` selects 0 selfchecks and lists `timeline-validate` as a manual harness, which is the CLI run recorded below.

`nodetool timeline validate --json` on all three shipped example storyboards, assembled through `buildStoryboardTimeline`, stays clean:

```
first-light / sneaker-drop / lighthouse-keeper →
"validation": { "ok": true, "errors": [], "warnings": [] }
```

And on a crafted bad document, so the new codes are shown reaching the surface rather than assumed to:

```
❌ 1 error(s), 3 warning(s)
  error Clip "Scrim" remaps time with keyframe 2 at t=0.2, at or before keyframe 1 at t=0.5 … (time_remap_not_monotonic)
  warn  Clip "Title" animation "a1" (fade, in) needs 900ms after a 400ms delay but the 1000ms clip leaves 600ms … (animation_exceeds_clip)
  warn  Clip "Title" sets 20px type in a 1080px frame — under 2.5% of frame height (27px) … (text_illegible)
  warn  Clip "Title" draws #dddddd on #ffffff (the shape clip behind it) — a contrast ratio of 1.36:1 … (text_illegible)
```

### The CI fix in this diff (`d8f92a99`)

The WebM half of the alpha round trip failed on CI and **skipped** in the dev container for want of ffmpeg, so T27's encoder arguments shipped unmeasured. A skipped test is not a passing one, and the first version of this PR body reported that skip as a clean result. It was not.

Installing ffmpeg 6.1.1 reproduced CI's `expected 255 to be +0` locally. The encoder was right: ffmpeg tags the track `alpha_mode=1` and the alpha file is a third larger than its opaque twin, so the plane is in the file. WebM stores VP9's alpha in BlockAdditions, and **ffmpeg's native `vp9` decoder discards it** — decoding with the default decoder reads every pixel opaque however much alpha the file holds. Naming `-c:v libvpx-vp9` on the decode reads the transparent half back at 0. The assertion is unchanged; only the instrument is. (An earlier version of this body said that test decodes with Mediabunny. It decodes with ffmpeg.)

### The regeneration in this diff

`RenderTimeline` grew four props and a second `frames` output, which made the checked-in DSL wrappers stale — `dsl-pack-guest.test.ts` catches exactly this, since a wrapper must name every slot the registry reports. Regenerated with `codegen:dsl`, `build --workspace=packages/dsl`, `build:sandbox-dsl` and `build:sandbox-flow`. `lib.apple` is excluded from the regeneration: its nodes only register on macOS, so running codegen on Linux deletes a namespace that is not stale.

## Agent capabilities

No capability is added and no declared contract changes. T27's node props are node properties, not a capability input; the `render_timeline` tool that would carry them is T22, on its own branch.

## New checks

- [x] **T27's alpha claim.** Forcing the compositor seed to `a: 1` and rebuilding produced `AssertionError: expected 255 to be +0`; restored and green. Separately, removing the `-c:v libvpx-vp9` decoder from the round trip reproduces CI's failure exactly, and restoring it goes green — so the check now fails for the right reason and passes for the right reason.
- [x] **T27's refusals** each ship a triggering fixture: `mp4 + alpha`, an unknown format string, an ffmpeg listing without `prores_ks` (and the same input passing once present), and a `CompositorUnavailableError` under an alpha request.
- [x] **T30's codes.** With `checkClipMotion` neutered, 8 of 17 motion tests go red; with `checkLegibility` neutered, 5 of 17. Restored and re-verified. Every shipped code has both a fixture that triggers it and one that does not.
- [x] **T31** adds no rule. Its 10 RTL cases each drive a control with `userEvent` and assert the store, so a section wired to nothing fails.

## Found, not fixed

- `packages/execution/tests/timeline-parenting.test.ts` carried a fixture with two clips on one track, which also produced `clips_overlap`; the old assertion only read `errors`, so it was invisible. Now filtered to `parent_*`.
- Browser **alpha** is not in the export dialog yet — the browser seed lives in `web/src/components/timeline/preview/gpu/`, and shipping a checkbox that silently produced an opaque file would be worse than omitting it. Being finished alongside T28.
- Two validator codes worth adding once someone owns them: `render_format_alpha_mismatch` (statically decidable from the node's prop bag via `ALPHA_OUTPUT_FORMATS`, today only a run-time refusal) and a `render_codec_override_ignored` warning when `video_codec` is set on a `webm` or `mov` render whose container pins its own codec.
- The `textStyle` and `shapeStyle` **tool** schemas remain stale against what the renderer honours — the same gap #5496 reports for T14/T16, plus `timelineAgentBridge.ts` disagreeing with the Zod schema on `shapeStyle.kind`. T19 owns reconciling all three style bags and is doing so on its own branch.
- Mediabunny cannot decode VP9 in Node (no WebCodecs VP9 decoder), so neither decoder available in the dev container would have caught the alpha bug above. Only CI's ffmpeg did.

---

https://claude.ai/code/session_01NXwPZaLeHkC1enGn5ZGA7D


georgi:
Code review looks good — alpha export with straight-alpha handling, new validator codes, and clip inspector are well-implemented. The VP9 alpha decode fix is well-documented.

However, GitHub reports a merge conflict against `main` (`mergeable_state: dirty`). The base SHA has drifted. A rebase/merge of main is needed before this can merge. No file overlap with #5505, so either can go first.

---

PR #5505, opened 2026-09-02
feat(timeline): mark the beats and cut to them [T20]

georgi:
## What changed

`docs/plans/motion-graphics-tasks.md` T20, following [#5491](https://github.com/nodetool-ai/nodetool/pull/5491) (M1+M2), [#5493](https://github.com/nodetool-ai/nodetool/pull/5493) (T16), [#5495](https://github.com/nodetool-ai/nodetool/pull/5495) (T14), [#5496](https://github.com/nodetool-ai/nodetool/pull/5496) (T15) and [#5498](https://github.com/nodetool-ai/nodetool/pull/5498) (T17+T18).

An agent could already hear a track — `detect_audio_events` returns onsets — and had no way to act on them. Markers could only be added from the mobile tool, and nothing could move a clip onto a beat.

Four ops: `add_marker`, `delete_marker`, `set_markers_from_beats`, `snap_to_beats`, with browser twins for the marker pair.

### The reporting contract is the point

A clip that does not snap is **reported, never dropped**. It comes back in the same `clips` array with `snapped: false`, `after` equal to `before`, a zeroed delta, and a reason naming the distance and the beat it missed:

```
start 590ms is 90ms from the nearest beat (500ms); tolerance is 60ms
```

Five skip reasons in all — out of tolerance, already on the grid, empty grid, a trim that would leave a non-positive duration, a move that would start the clip before zero — plus `no clip matches "ghost"` for a target name nothing resolves to, appended to the *same* list. One list, so a caller never infers a skip from a shorter result.

### Other decisions

- **`snapClipsToGrid` is pure**, in `packages/timeline/src/beats.ts`; the op calls it and holds no arithmetic.
- `move` slides a clip and keeps its length; `trim` holds the opposite boundary and changes it. Under `mode: "both"` a move takes the start, since one move cannot satisfy two boundaries.
- **A tempo grid computes each beat from its index** rather than accumulating an interval, so a fractional beat length does not drift over a long timeline — pinned at 140 BPM, beat 100 = 42857 ms.

## Verification

Exit codes captured directly, never through a pipeline. `npm run build:packages` run first (59/59).

- [x] `packages/timeline` — 566 passed, 24 skipped
- [x] `packages/agents` — 3819 passed, 5 skipped
- [x] `packages/execution` — 373 passed · `packages/cli` — 695 passed
- [x] `web` (jest, timeline-related) — 781 passed, 87 suites
- [x] `npm run capabilities:sync` / `capabilities:check` — table current, 253 capabilities
- [x] `npm run lint` — 0 errors
- [ ] `npm run typecheck` — web and electron legs exit 0 run alone; the **mobile** leg fails on a clean tree because `mobile/node_modules` is absent. No mobile file changed (mobile already shipped both marker tools).

Re-verified after rebasing onto current `main`: timeline 566 passed.

## New checks

- [x] `nearestGridTime`'s tolerance comparison was widened tenfold (`distance > toleranceMs * 10`) and observed failing: `answers null when nothing is in reach`, `leaves a clip 90 ms off the beat where it is, and says why`, `reports the clips that did not move alongside the ones that did` — 3 failed | 23 passed. Restored and re-verified 26/26.
- The edit was **grepped in the file before and after** to confirm it applied, after an earlier task on this branch had an inversion silently no-op and produce a green run that proved nothing.

## Found, not fixed

- **`snap_to_beats` contests every unit kind in the `resource_change` merge.** `opUnitKind` in `web/src/stores/timeline/merge.ts` reads the kind from a substring of the verb, and `snap_to_beats` contains none of `clip`/`track`/`marker`, so it falls back to `ALL_UNIT_KINDS` with no unit id. That is conservative and safe — a concurrent editor just prefers the server for anything that drifted — and stamping the clip ids would be wrong under the current mapping, since a null `ownKind` would pair those ids with the marker and track kinds too. Making the mapping op-aware is a merge-layer change.
- **`edit_timeline`'s CAS retry re-applies beat ops against a newer document.** Correct for the marker ops, but a `snap_to_beats` retried after a concurrent write snaps against whatever the clips now are, which may differ from what the first attempt reported. Every other op already has this property; it is noted because the per-clip `before` values make it visible for the first time.
- **`MAX_OPS = 60` bounds the op list, not the work** — one `set_markers_from_beats` can add 2048 markers. The grid cap is the only bound; there is no document-level marker cap.

---

https://claude.ai/code/session_01VdzG3a3jN3Dzx6QV8xh4Bo

---
_Generated by [Claude Code](https://claude.ai/code/session_01VdzG3a3jN3Dzx6QV8xh4Bo)_

### DIFF
--- a/packages/agents/src/capabilities/timelines.specs.ts
+++ b/packages/agents/src/capabilities/timelines.specs.ts
@@ -126,7 +126,8 @@ export const EDIT_TIMELINE_SCHEMA: JsonSchema = {
         "split_clip, trim_clip, move_clip, duplicate_clip, delete_clip, " +
         "set_clip_params, set_clip_binding, set_transition, set_mask, " +
         "set_matte, set_effects, animate_clip, " +
-        "clear_animations, list_animation_presets, select_clip, seek. " +
+        "clear_animations, list_animation_presets, select_clip, seek, " +
+        "add_marker, delete_marker, set_markers_from_beats, snap_to_beats. " +
         "Start with get_state to " +
         "read track and clip ids. To lay existing videos end to end, call " +
         'add_media_clip once per asset ({"op": "add_media_clip", "asset": ' +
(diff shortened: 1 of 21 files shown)


georgi:
Reviewed. Code looks correct — marker ops, snap-to-beats logic, and skipped-clip reporting are well-structured. Three self-documented gaps (merge-layer opUnitKind fallback, CAS retry re-snap, no marker cap) are acceptable debt, not blockers. No file overlap with any other open PR.

Ready to merge once the remaining CI checks are green.

---

PR #5548, opened 2026-09-03
Track only what needs tracking, and make erasure mechanical

georgi:
## What changed

The hosted service stores personal data across 44 tables, 34 of them keyed by `user_id` — messages, memories, assets, prediction parameters — with no retention on most of it, no way to erase or export an account, and a privacy policy that still described NodeTool as a purely local-first desktop app. This branch narrows what gets recorded about a user, gives the recorded data an expiry, makes erasure and export mechanical, and brings the published policy in line with what the code does.

A `nodetool_user_events` audit table records only authentication events, irreversible or outward-facing actions (deletes, secret writes, shares, deploys, publishes), and consent/terms/data-subject-request records. No clicks, page views, IP addresses, user agents or free-text columns — behavioural analytics stays aggregate in Plausible and is never keyed to a user. Metadata passes a per-event-type allowlist before the insert, so a caller cannot smuggle prompt text into the audit log.

Retention now covers the two tables that held personal data and never expired. `run_events` is swept by age through its parent job. `nodetool_predictions` is redacted rather than deleted: `parameters`, `metadata` and `logs` are nulled while every billing column survives, so the charge stays auditable and the prompt does not.

Erasure and export are driven by a single registry that names every table once — its disposition, how the user is reached, and why. An audit test fails when a table with a `user_id` has no entry, when an entry has no handler, and when a handler touches a table not in the registry, so the three cannot drift apart. Dispositions are not uniform: credit ledger and subscription rows are retained under Art. 17(3)(b), predictions are redacted, and consent and data-subject-request events are kept because they are the proof a request was honoured. `GET /api/account/export` and `DELETE /api/account` expose it.

## Two findings worth a reviewer's attention

**A blob-erasure bug that would have deleted other people's files.** The object sweep lists by the user's prefix, but the S3 and Supabase adapters match a prefix as a plain string and `normalizeStorageKey` strips the trailing slash — so listing `user-1/` also returns `user-10/` and `user-1x/`. Erasing account `user-1` would have deleted another account's objects. Filtered and tested.

**Three tables a `user_id` sweep silently misses.** `nodetool_workflow_shares` is keyed on `created_by`, so live share tokens would outlive the workflows they point at. `mcp_oauth_tokens` hangs off the grant, so live tokens would outlive it. `application_budgets` hangs off the application. These were found by enumerating the schema through Drizzle's metadata rather than grepping for the column — grepping undercounts, because six schema files define more than one table.

## Verification

- `npm run test --workspace=packages/websocket` — 2872 passed, 253 files (2858 before, no regressions)
- `npm run test --workspace=packages/models` — 979 passed, 65 files (947 before)
- `npm run test --workspace=packages/config` — 149 passed
- `npm run test --workspace=packages/protocol` — 1055 passed
- `npm run build` for `models`, `config`, `protocol`, `websocket` — clean
- `web` typecheck — exit 0; `npm run lint` — exit 0
- `npx tsc --noEmit` on the changed privacy files — clean

Known unrelated failure: root `npm run typecheck` fails on the `mobile` leg because `mobile/node_modules` is not installed in this container. No mobile source touches these schemas.

- [x] `npm run test:affected` — covered by the per-workspace runs above
- [x] `npm run typecheck` — passes for every workspace this diff touches; `mobile` fails for the reason above
- [x] `npm run lint`
- [ ] `npm run dev:nodetool -- harness gate --base main` — not run

## Agent capabilities

No capability is added and no declared contract changes.

## New checks

Every new check was inverted and observed failing, then reverted.

**Registry audit** — a new `user_id` table with no entry turned 2 tests red, naming it by table and schema export. Also inverted: deleting a registry entry (3 red), deleting an erasure step while keeping the entry (5 red, the registered-but-unimplemented case), selecting collaborators by `workflow_id` instead of the subject (3 red, a cross-user leak), and removing `withheldColumns` from `nodetool_secrets` (2 red, ciphertext in the export).

**Cross-user access** — pointing the subject at a caller-supplied query param turned both tests red with `expected 'user-a' to be 'user-b'`. The erase case actually erased the wrong user's rows and objects in that run, so it exercises the whole path rather than a report field.

**Retention sweeps** — a no-op sweep turned 3 tests red; a sweep ignoring the retention window turned 3 different tests red.

Positive controls: the audit asserts it found at least 40 tables and 30 user-keyed ones before asserting anything, so it cannot pass by matching nothing. One test was found tautological and rewritten — the retention carve-out test iterated the constant it was testing, so deleting an entry shrank the test instead of failing it.

- [x] The check was inverted and observed failing
- [x] An audit asserts it found its targets

## Open items needing a human decision

Not code. Items 1–3 are reasons not to publish the policy yet.

1. **Controller identity.** The policy says "the NodeTool team"; Art. 13(1)(a) requires a named legal entity. `marketing/src/app/imprint/page.tsx` names an individual as the § 5 DDG provider, and `marketing/LEGAL_OPEN_ITEMS.md` assumes a Dutch entity that appears nowhere else. Left untouched rather than guessed at.
2. **Art. 28 processor agreements** with Fly.io, Supabase, Cloudflare and Plausible are asserted by the inherited policy text. Nothing in the repo evidences one exists.
3. **The 14-day server-log commitment has no implementation.** `fly.toml` configures no drain or retention, `packages/config/src/logging.ts` has no rotation or age-based deletion, and `server.ts` sets `logger: false` so the application never writes the IP/user-agent tuple the policy describes — Fly and Cloudflare produce it, outside this repo. The only rotation anywhere is `docker-compose.yml` json-file at 10 MB × 3, size-based and self-hosting only. Nothing would fail if retention were 90 days.
4. **`NODETOOL_STORAGE_AUTO_CLEANUP=1` must be set on the Fly deployment**, or the retention schedule the policy promises does not run in production.
5. **No age gate** exists in the hosted sign-up path, while the policy retains a section on children's data.

Two limitations are recorded in the code rather than papered over: objects written before the owner prefix existed sit at flat keys no per-user sweep can reach, and `nodetool_team_tasks` is classified not-personal because no writer for it exists anywhere in the repo — flagged to revisit in whatever PR adds one.

The policy was written to be accurate about the code, which is not the same as being sufficient as a legal document. It should be reviewed by a qualified data protection advisor before publication.

https://claude.ai/code/session_017bcqFMby7BNbRJigzfmE13

### DIFF
--- a/marketing/src/app/privacy/layout.tsx
+++ b/marketing/src/app/privacy/layout.tsx
@@ -3,7 +3,7 @@ import type { Metadata } from "next";
 export const metadata: Metadata = {
   title: "Privacy Policy | NodeTool",
   description:
-    "How NodeTool collects, processes, and protects your personal data. GDPR-compliant privacy policy for nodetool.ai and the NodeTool desktop application.",
+    "How NodeTool collects, processes, and protects your personal data. GDPR privacy policy for nodetool.ai, the NodeTool desktop application, and the hosted service at app.nodetool.ai.",
   alternates: {
     canonical: "/privacy",
   },
--- a/marketing/src/app/privacy/page.tsx
+++ b/marketing/src/app/privacy/page.tsx
@@ -1,8 +1,33 @@
 import Link from "next/link";
 
+/*
+  DRAFT — NOT LEGAL ADVICE. This policy must be reviewed by a qualified data
+  protection advisor before it is published.
+
+  Items a human must confirm before this page goes live:
+  1. Signed Art. 28 data processing agreements with Fly.io, Supabase,
+     Cloudflare, Plausible, the email provider, and every AI provider offered
+     in the hosted service. This page states that such agreements are in
+     place; that fact is not verifiable from the repository.
+  2. The transfer mechanism actually relied on per non-EU recipient
+     (adequacy / EU-U.S. Data Privacy Framework certification / SCCs).
+  3. The Supabase project region. `fly.toml` documents the app's primary
+     region as `fra` and notes the database sits in AWS eu-central-1, but the
+     Supabase project settings are the authority.
+  4. The 14-day server-log figure. It is a commitment we make, not a setting
+     read from Fly.io or Cloudflare configuration in this repository.
+  5. Sections 5, 6, 10 and 11 describe the hosted service's activity log,
+     data export and account deletion. Do not publish this page before those
+     ship — at the time of writing they are being implemented in `packages/`.
+  6. That the retention sweep is actually switched on for the hosted
+     deployment (`NODETOOL_STORAGE_AUTO_CLEANUP=1`). Section 11 promises run
+     history and snapshots expire on a schedule; without that flag the sweep
+     only runs when someone triggers it.
+*/
+
 export const dynamic = "force-static";
 
-const LAST_UPDATED = "2 May 2026";
+const LAST_UPDATED = "3 September 2026";
 
 export default function PrivacyPage() {
   return (
@@ -27,9 +52,10 @@ export default function PrivacyPage() {
           <p>
             This Privacy Policy explains how we process personal data when you
             visit <strong>nodetool.ai</strong>, use the <strong>NodeTool</strong>{" "}
-            desktop application, or contact us. We follow the EU General Data
-            Protection Regulation (GDPR) and the German Federal Data Protection
-            Act (BDSG).
+            desktop application, use the hosted service at{" "}
+            <strong>app.nodetool.ai</strong>, or contact us. We follow the EU
+            General Data Protection Regulation (GDPR) and the German Federal
+            Data Protection Act (BDSG).
           </p>
 
           <h2>1. Controller</h2>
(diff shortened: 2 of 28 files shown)


georgi:
CI went red on `f6d3ff5b` across eight checks. Two distinct causes, both now fixed on the head.

**1. `Typecheck, lint & build` — not this PR's.** The failure was:

```
[provider-catalog] marketing/src/data/providerCatalog.generated.ts is stale. Run: npm run seo:provider-catalog
```

The generator reads provider manifests under `packages/`, which this branch does not touch, so its inputs are byte-identical to `main`. Running the check locally on those inputs reproduces the failure, which puts it on `main` rather than here. The fix is the one the error names, it is deterministic from repo contents (no network), and the diff is two FAL video model counts — two models reached the manifests without the catalog being regenerated. Ported in `37eecabc`; it no-ops once `main` carries the same regeneration. Verified: the check now reports `up to date`.

**2. The remaining checks — this PR's, already fixed.** `f6d3ff5b` added retention fields to the model layer but not to the protocol schema, so `packages/websocket/src/storage-retention.ts` referenced fields that did not exist and the build failed, cascading into the integration, browser and docker legs. `c6b57b47` wires those fields through protocol, websocket and web.

Verified locally before pushing: `packages/protocol` and `packages/websocket` build clean, `web` typecheck exits 0, `packages/websocket` 2858 tests pass, `packages/models` 947 pass, `packages/config` 149 pass.

No check was re-run and no test was skipped or quarantined. Watching the run on `37eecabc`.

---

PR #5549, opened 2026-09-03
feat(credits): whitelist managed models and grant new users free credits

georgi:
## What changed

Three commits. The first is the feature; the other two came out of review and CI.

**`ed5bc61c` — the credit system.** Production runs on platform-owned keys, so the operator decides which of the curated catalog is open for business and what a new account starts with. Two env vars, parsed in `packages/config/src/credit-policy.ts`:

- **`NODETOOL_CREDIT_MODELS`** — the managed model ids credits may be spent on, separated by commas/whitespace/newlines. Unset means the whole catalog: the right default for a staging server that holds platform keys. A model outside the list is hidden by `NodetoolProvider`'s listers, refused by `delegateFor` before a platform key is used, and refused by the gate before a run starts — with `MODEL_NOT_AVAILABLE` rather than `BUDGET_EXCEEDED`, so a full balance is not offered as the fix. The surviving ids reach clients as `spendableModels` on `credits.status` and narrow Studio's dropdowns; a project pinned to a model the operator stopped selling moves to one they do sell. `nodetool/director` drives the in-editor assistants and is not user-selectable, so a whitelist that leaves it out turns them off — called out in the docs.
- **`NODETOOL_SIGNUP_CREDITS`** — a one-time welcome grant, default 500, inserted on a user's first balance read. Keyed `signup:<userId>`, so raising it later reaches new users without re-granting existing ones; `0` switches it off. It rides the same lazy-accrual, idempotent-by-primary-key shape the monthly plan grant already uses — no cron.

`checkCredits` now takes the managed model ids the work names and refuses on the model before the balance, returning a typed `refusal` (`model_not_allowed` | `insufficient_credits`) that callers map onto their own error codes instead of sniffing the message. The three direct-inference RPCs and the graph run gate all pass their models through.

Buying credits is not in this diff. The `topup` mutation and ledger seam are unchanged, so a payment webhook still drops in behind them.

**`62e86ab2` — the managed provider is cloud-only.** `nodetool` spends NodeTool's own platform keys against a credit balance, so it is the platform's wallet rather than a key a user pastes into settings. A desktop install has neither the keys nor an account to bill, and a self-hosted server has no claim on NodeTool's. It is now in `CLOUD_ONLY_PROVIDER_IDS` — the mirror image of `NON_CLOUD_PROVIDER_IDS` — and the provider index prunes it off the cloud profile exactly as the local engines are pruned on it.

| | profile | `nodetool` |
|---|---|---|
| Electron desktop | `NODETOOL_NODE_PROFILE=full` | absent |
| dev checkout | unset | absent |
| self-hosted compose | `production` + `full` | absent |
| Fly production | `NODETOOL_ENV=production` | present |

This also corrected a wrong claim in the first commit's docs, which said an unset `NODETOOL_CREDIT_MODELS` was "what a local install wants". A local install has no managed provider to whitelist at all.

One test premise moved with it: the contract suite asserted every `MEDIA_ONLY_EXEMPTIONS` entry names a *registered* provider, which a conditionally-registered one breaks. It now accepts "live" — registered, or named in `CLOUD_ONLY_PROVIDER_IDS` — with a companion test pinning that the cloud-only ids really are absent under this profile, so the staleness guard still fails in both directions rather than quietly skipping a provider it should cover.

**`aa6900ea` — a docs-link fix this PR did not cause.** See [the comment below](https://github.com/nodetool-ai/nodetool/pull/5549#issuecomment-5532713823): Docs CI failed here with 9 broken internal links introduced by 79a3563d, in files this branch did not write. The check only runs on a PR touching `docs/**`, so nothing tripped it until this one did. No fix existed to port and the PR cannot merge without it, so the 9 are rewritten as absolute GitHub blob URLs — the convention already used in `docs/HARNESS_FIRST.md`, `docs/cli.md` and `docs/configuration.md`. Happy to split this into its own PR if you would rather keep this one to the credit system.

## Verification

- [x] `npm run test:affected` — one pre-existing environment failure: `@nodetool-ai/image-nodes` fails with `No WebGPU adapter available (Node/Dawn)` (no Vulkan ICD in this container; AGENTS.md documents this as a missing driver, not a broken test). Because turbo stops there, the affected suites were run directly instead:
  - `packages/protocol` 58 suites · `packages/config` 13 · `packages/models` 62 · `packages/websocket` 252 — all passed
  - `packages/runtime` 172/173 suites; the one failure is `tests/providers/video-frame-fallback.test.ts` on `Error: spawn ffmpeg ENOENT` (ffmpeg is not installed here)
  - `web` 1265 suites / 14 236 tests passed · `electron` 63 suites / 690 tests passed
- [x] `npm run typecheck` — web and electron clean. The mobile leg fails wholesale on `Cannot find module 'react-native'` etc.: `mobile/node_modules` does not exist in this container. This diff does not touch `mobile/`.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 5/5 selfchecks passed`. (`--base main` errors with `no merge base` until `git fetch origin main`.)
- [x] Docs CI reproduced locally with the workflow's own commands — jekyll build then htmlproofer with the same flags and `LANG`/`LC_ALL=C.UTF-8`: 9 failures / exit 1 before, `HTML-Proofer finished successfully` / exit 0 after. The artifact and privacy-guard steps pass on the rebuilt site.

## Agent capabilities

No capability added and no declared contract changed. `npm run capabilities:check` reports `capability-table.ts is stale`, but that predates this branch: `git diff origin/main -- packages/agents packages/cli/src/harness/capability-table.ts` is empty, and the drift is in `generate_video`, `animate_image`, and `render_timeline`, none of which this diff touches.

## New checks

- [x] Every new rule was inverted and observed failing, then restored.

  Disabling the whitelist (`return allowed === null || allowed.has(modelId) || true;`) and rebuilding `@nodetool-ai/config`:

  ```
  ### models:     × refuses a model the operator did not whitelist, however full the balance
                  × names the whitelist as the spendable catalog, and the whole catalog without one
                  Tests  2 failed | 11 passed (13)
  ### runtime:    × hides a funded model the whitelist leaves out
                  × refuses to serve a model the whitelist leaves out
                  Tests  2 failed | 7 passed (9)
  ### websocket:  × refuses a model outside the operator's whitelist
                  Tests  1 failed | 6 passed (7)
  ### config:     × allows only what the whitelist names
                  Tests  1 failed | 6 passed (7)
  ```

  Disabling the welcome grant (`if (true) return;` in `ensureSignupGrant`):

  ```
  × grants welcome credits once, and only when configured
  × the welcome grant reaches a new user on the first balance read
  Tests  2 failed | 11 passed (13)
  ```

  Removing the cloud-only prune block from the provider index:

  ```
  × drops the managed provider when the profile is off
  × drops the managed provider for a self-hosted server
  ```

  Worth recording: the first inversion attempt edited only `packages/config/src` and the models/runtime/websocket suites stayed green, because they resolve `@nodetool-ai/config` from `dist/`. The signal above is from the rebuilt package — a source-only inversion proves nothing across a package boundary here.

- [x] The tests assert both directions (a model the whitelist names is admitted, one it omits is refused; the provider is present under the cloud profile and absent off it), so they cannot pass by matching nothing.

https://claude.ai/code/session_01XXG1JVwY5sPjeMbMPfg3Dk

### DIFF
--- a/docs/agentic-video-product.md
+++ b/docs/agentic-video-product.md
@@ -91,10 +91,17 @@ Server-owned, in `packages/models/src/credits.ts` (`@nodetool-ai/models`):
   `@nodetool-ai/model-pricing` translates `nodetool/...` ids to the delegate
   before lookup so estimates are real numbers. Models appear in the pickers
   only when their platform key is set.
+- **The managed provider is cloud-only.** `nodetool` is in
+  `CLOUD_ONLY_PROVIDER_IDS` (`packages/protocol/src/cloud-profile.ts`) and the
+  provider index unregisters it whenever the cloud profile is off — a desktop
+  install and a self-hosted server (`NODETOOL_NODE_PROFILE=full`) both have
+  platform keys they do not own and no account to bill, so the provider would
+  only ever produce an error. Those users bring their own keys, unmetered.
 - **The operator picks which models the platform sells.**
   `NODETOOL_CREDIT_MODELS` is a comma/whitespace-separated list of curated
   model ids (`nodetool/flux-schnell nodetool/kokoro`); unset means the whole
-  catalog, which is what a local install wants. A model outside the list is
+  catalog, the right default for a staging server holding platform keys. A
+  model outside the list is
   hidden by the provider's listers, refused by `delegateFor` before a platform
   key is used, and refused by the gate before a run starts
   (`MODEL_NOT_AVAILABLE`, not `BUDGET_EXCEEDED` — a full balance does not
--- a/docs/agentic-video-product.md
+++ b/docs/agentic-video-product.md
@@ -65,10 +65,15 @@ Server-owned, in `packages/models/src/credits.ts` (`@nodetool-ai/models`):
 - **1 credit = $0.01 of provider spend.** Spend is never double-booked: the
   balance is `sum(grant ledger) - ceil(prediction spend / 1¢)`, read straight
   from the `nodetool_predictions` rows every provider call already writes.
-- **Ledger** (`nodetool_credit_ledger`) holds grants only: monthly plan
-  accruals, top-ups, adjustments. Plan grants use the row id
-  `plan:<userId>:<planId>:<YYYY-MM>`, so the lazy accrual (run on every
-  status read) is idempotent by primary key — no cron.
+- **Ledger** (`nodetool_credit_ledger`) holds grants only: the welcome grant,
+  monthly plan accruals, top-ups, adjustments. Grants are keyed so the lazy
+  accrual (run on every status read) is idempotent by primary key — no cron:
+  `signup:<userId>` once ever, `plan:<userId>:<planId>:<YYYY-MM>` per month.
+- **New users get free credits.** The first balance read for a user inserts
+  `NODETOOL_SIGNUP_CREDITS` (default 500) as a welcome grant, on top of the
+  free plan's monthly accrual. Raising the amount later reaches users who have
+  not signed up yet; it never re-grants one who already has the row. `0`
+  switches the welcome grant off.
 - **Plans** (`nodetool_user_subscriptions`, catalog `CREDIT_PLANS`): Free
   300/mo, Creator 3,000/mo ($12), Pro 10,000/mo ($40). Switching is instant
   and unbilled; a payment provider integration replaces the `topup` mutation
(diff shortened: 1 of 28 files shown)


georgi:
Heads up on two files in this diff that have nothing to do with credits: `docs/harnesses.md` and `docs/dev-environment.md`.

Docs CI's `build-and-check` failed here with 9 broken internal links, none in a file this branch wrote. They came in with 79a3563d, which moved the harness reference out of the root `AGENTS.md` into the **published** `docs/harnesses.md`, carrying its relative links along. The root `AGENTS.md` and everything under `packages/` sit outside the Jekyll source, and `docs/AGENTS.md` is on the `exclude:` list by design — the privacy guard fails the build if `AGENTS.html` ever appears — so `../AGENTS.md#…` and `../packages/…` can never resolve in `_site`. The check only runs on a PR touching `docs/**`, so nothing tripped it until this one edited `docs/agentic-video-product.md`.

No fix for it existed to port, and the PR can't merge without it, so I rewrote the 9 as absolute GitHub blob URLs — the convention already used for exactly this in `docs/HARNESS_FIRST.md`, `docs/cli.md` and `docs/configuration.md`. Nothing else in those two files changed. The tenth relative escape (`../AGENTS.md` in `docs/telegram-bot-implementation-plan.md`) passes today, because html-proofer resolves the hashless path to `docs/AGENTS.md` on disk, so I left it alone rather than widen further.

Reproduced with the workflow's own commands before and after, including `LANG`/`LC_ALL=C.UTF-8` (without those, nokogiri dies on an `Encoding::InvalidByteSequenceError` instead of reporting links):

```
# before — docs as they stand on main
HTML-Proofer found 9 failures!          exit 1

# after
Checking 1786 internal links
Ran on 856 files!
HTML-Proofer finished successfully.     exit 0
```

The `llms.txt` / `catalog.json` artifact step and the privacy guard both pass on the rebuilt site.

If you would rather keep this PR to the credit system, the link fix is a clean standalone commit (aa6900ea) and I can move it to its own PR — say the word.

---

PR #5552, opened 2026-09-03
feat(system-skills): model-line prompting guides, triggered by model id

georgi:
## What changed

Seven generation model lines have house rules that differ from each other — Seedance wants a sound brief, GPT Image wants five labelled slots, MiniMax H3 wants a job assigned to every reference — and one prompt shape written for all of them gets the blandest reading each one has. This adds one system skill per line (`nano-banana-pro-prompting`, `gpt-image-2-prompting`, `flux-2-klein-prompting`, `seedance-2-prompting`, `veo-3-prompting`, `minimax-h3-prompting`, `wan-2-6-prompting`), adapted from fal's published guides and rewritten against NodeTool's capability contract.

Each reaches the model two ways, neither of them auto-selection. The catalog description names the model ids every provider serves the line under, so an agent that already picked its model can `load_skill` off the catalog. And `find_model` attaches `prompting_skill` to a matching route, lifted to the top level alongside `ref` — that is the step immediately before the prompt gets written, rather than after the render comes back wrong. `MODEL_PROMPTING_SKILLS` (`packages/agents/src/model-prompting-skills.ts`) matches on the model id alone, so a new provider serving a covered line needs no change here.

The `seedance-2` skill also carries a structured JSON beat-sheet brief — global camera, light, style, negative prompt and audio rule, then a timeline of beats each naming its function, picture and sound — for a piece with an arc rather than a single shot.

Second commit regenerates `capability-table.ts`. `find_model`'s description moved its contract fingerprint; the other four rows in that diff (`edit_image`, `generate_video`, `animate_image`, `render_timeline`) were already stale on main — their specs changed in e7d7e13b and f59c1cc1, both after 1fd8d2be last wrote the generated file.

## Verification

- [x] `npm run test:affected` — packages pass except `@nodetool-ai/base-nodes`: 14 failures, all in `image-examples-run.test.ts` / `image-processing-examples-run.test.ts` with `No WebGPU adapter available (Node/Dawn)`. That is the missing Vulkan ICD this environment has no driver for, documented in AGENTS.md, and the diff touches no image code. Web (1266 suites, 14244 tests) and Electron (63 suites, 690 tests) pass.
- [x] `npm run typecheck` — web and electron clean. Mobile fails on `Cannot find module 'react-native'` etc. because `mobile/node_modules` is absent in this container; mobile is not a root workspace and no mobile file is touched.
- [x] `npm run lint` — clean. One finding of my own was fixed: `anti-slop(no-conditional-empty-object-spread)` on the conditional spread that attached the field, rebuilt as a plain assignment.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 3/3 selfchecks passed`, 261/261 graphs validate.
- [x] `npm run check:agents-docs` — 37 AGENTS.md files, each paired and reachable.

## Agent capabilities

No capability is added. `find_model`'s description changed (it now says a result carrying `prompting_skill` names a guide to read), and its result gains an optional `prompting_skill` field.

- [x] `npm run capabilities:check` passes after `capabilities:sync`
- [x] `find_model` already names `packages/agents/tests/capabilities-models.test.ts` and its rankings suite; the new behaviour is covered in `packages/agents/tests/model-prompting-skills.test.ts` (attaches and lifts the field on a matching route, omits the key entirely on a model no guide covers)

## New checks

`packages/agents/tests/model-prompting-skills.test.ts` checks the table against three things it can silently disagree with: the skills actually on disk, the model ids the shipped provider manifests actually name, and the capability registry.

- [x] Each check was inverted once and observed failing. Renaming a table entry to `veo-3-prompting-typo`, narrowing the Wan pattern to `wan-2.9`, widening the Seedance pattern to bare `seedance`, pointing two entries at the same pattern, and putting `analyze_teleport` in a skill body produced:

```
× names only skills this build ships
  AssertionError: in the table but not in packages/system-skills: expected [ 'veo-3-prompting-typo' ] to deeply equal []
× gives every line at least one model a provider actually serves
  AssertionError: no shipped model id matches this line: expected [ 'Wan 2.6' ] to deeply equal []
× claims nothing for bytedance/seedance-1.5-pro
  AssertionError: expected 'seedance-2-prompting' to be null
× claims no two lines for the same model id
  AssertionError: expected [ …(16) ] to deeply equal []
× names only capabilities that exist
Tests  11 failed | 33 passed (44)
```

- [x] The manifest scan asserts it found its targets before either pattern check runs (`expect(ids.length).toBeGreaterThan(1000)` plus a named id), so it cannot pass by matching nothing. The capability scan and the description scan assert the same way.

Negative cases are spelled out per line, so an over-broad pattern fails rather than silently handing the wrong guide to a neighbouring version: `nano-banana-2`, `gpt-image-1.5`, `gpt-image-1-mini`, `flux-2-pro`, `flux-2-dev`, `seedance-1.5-pro`, `seedance/v1/pro`, `hailuo-2.3`, `wan-2.5`, `wan-2.7`.

https://claude.ai/code/session_01H3ecoeYSKHnYUfXeJ7iwoh

---
_Generated by [Claude Code](https://claude.ai/code/session_01H3ecoeYSKHnYUfXeJ7iwoh)_


georgi:
**Docs CI (`build-and-check`) was red on three dangling links that are not this PR's — fixed here, because nothing else will fix them.**

The failure:

```
* At ./_site/cli.html:2613:            internally linking to RELIABILITY_ARCHITECTURE.md, which does not exist
* At ./_site/creative-agent.html:560:  internally linking to /creative-agent-roadmap, which does not exist
* At ./_site/harnesses.html:1913:      internally linking to RELIABILITY_ARCHITECTURE.md, which does not exist
HTML-Proofer found 3 failures!
```

**Why it is not this PR's.** The only file this PR changed under `docs/` is `docs/AGENTS.md`, which `_config.yml` excludes from the build entirely. `cli.md`, `harnesses.md` and `creative-agent.md` are untouched. Building `origin/main` in a clean worktree and running html-proofer with the workflow's own flags reproduces the same three failures, at the same files and line numbers, over the same 851 files.

**Where they came from.** Main has been deleting internal docs one commit at a time without updating the links pointing at them — `docs/RELIABILITY_ARCHITECTURE.md` in 9b6a4d83, `docs/CREATIVE_AGENT_ROADMAP.md` alongside it, plus `KERNEL_PARITY_GAPS`, `KERNEL_RUNNER_ANALYSIS`, `STUDIO_PRD`, `STUDIO_MVP_PROMPT`, `CLOUD_NODE_CURATION`, `CODE_NODE_COVERAGE`, `RELIABILITY_TASKS`. Nothing caught it: `docs-ci.yml` is `on: pull_request` with a `paths: docs/**` filter and never runs on main. So the breakage stays invisible until the next PR that touches `docs/` — which then inherits it and cannot go green.

**Why I fixed it here rather than only reporting it.** No fix exists in any other PR to port, and I can only push to this PR's branch. Leaving it would mean a PR that can never be green, and the same block for every subsequent docs PR. The change is three link removals (`eb6656e6`) — the pages are gone, so each link is dropped rather than repointed, and the surrounding sentences stand without them. Main is also merged in (`d80e6d58`, clean) so this builds against the deletions rather than around them.

**Verified**, not assumed — built the site and ran the workflow's exact command:

```
Ran on 851 files!
HTML-Proofer finished successfully.
exit=0
```

Also re-run after the merge: `harness gate --base origin/main` 3/3 selfchecks, 261/261 graphs; `capabilities:check` current; `check:agents-docs` 37 files; web and electron typecheck clean; lint zero errors; the agents suites for this change pass (103 tests).

Worth a separate look by someone with the context: `docs-ci.yml` only validating on PRs is what let this land. A scheduled or push-to-main run of the same job would catch the next deletion at the commit that causes it, instead of on an unrelated contributor's PR. I have not changed the workflow — that is a call for whoever owns the docs pipeline.

---

PR #5584, opened 2026-09-04
fix(timeline): close the motion-graphics review findings

georgi:
## What changed

A review of the motion graphics system (`docs/plans/motion-graphics.md`) found verified bugs, a schema guard that could not catch the drift it exists for, compositor parity gaps the preview did not report, three diverging copies of the timeline edit ops, and skill text that had drifted from the code. This PR closes them, one commit per action. Behavior fixes: `splitClip` no longer duplicates emphasis and loop animations across the cut; the headless trim and snap ops go through the engine, so a time-remapped clip is refused and a snapped group carries its children; a speed-multiplied clip's audio is retimed with the picture; `render_timeline` returns the zip for `png_sequence` and honors `bitrate`; opacity is clamped once in the scene model; a trim to a non-positive duration is refused instead of clamped. The wire schema gains the two clip fields it was stripping, bounds on custom curves, and a key-difference mirror guard that typechecks under `lint`; the save route logs and returns the paths it dropped. Canvas 2D reports every degradation beside `skipped` and applies brightness additively like the GPU grade. The edit ops live in one pure `applyTimelineOp` module used by the headless bridge and the web tools, with a parity test. The render pipeline stores output as an asset, skips blur samples on still frames, submits one encoder per frame, tags BT.709 and writes progress to the job row. The inspector gains time remap, stagger unit and direction, re-parenting, a group panel and easing feedback. Eval predicates grade motion outcomes with failing fixtures, and the skill matches the code.

Not in this PR: the mobile op migration. Mobile wires only dependency-free protocol modules and the ops module reaches the zod-backed tool-param schemas, so it needs a protocol split first; mobile still carries its own op copy.

## Verification

- [x] `npm run test:affected` selects every workspace because the `main` merge touches root files; instead the touched packages ran directly on the merged tree: `packages/timeline` 691 passed (GPU suites skip without an adapter), `packages/agents` 1191 passed across the capability, timeline, parity, JTBD and creative-pipeline suites, `packages/execution` 450 passed, `packages/video-nodes` full suite except the Dawn decode test whose device hook times out here as on `main`, `packages/websocket` timeline router suites, `packages/models` 986 passed, web timeline hooks, tools, components and stores 1083 passed. The full pass is CI's.
- [x] `npm run typecheck` clean for web and electron; mobile fails only on missing `node_modules` in this environment.
- [x] `npm run lint` clean.
- [ ] `npm run dev:nodetool -- harness gate --base main`: the `capability-suites` leg failed on the previous head because the merge with `main` left `capability-table.ts` stale; regenerated with `npm run capabilities:sync` and the check passes.

Inverted checks observed failing: the key-difference mirror guard (`TS2741: Property 'resolution' is missing` after deleting the field), the curve bounds test (3 of 4 red with `.max` removed), the stripped-path check (`expected undefined to deeply equal ['clips[*].holographicDepth']`), and every new eval predicate and validator fixture was written red first.

## Agent capabilities

- [x] Changed contracts (`preview_timeline_frame` response, `render_timeline` `bitrate` type, `edit_timeline` refusals) keep their existing suites in `capability-table.ts`; no capability was added.

## New checks

- [x] `degraded` reporting, additive brightness, the mirror guard, the stripped-path check, the scrim, mid-motion and JTBD predicates each ship with a fixture that fails without the change
- [x] The parity test walks every op name the module implements, so it cannot pass on an empty table

https://claude.ai/code/session_015acngCQoFnYR2eCsw4wA1t


packages/timeline/src/render/frameCompositor.ts
```diff
@@ -732,10 +775,11 @@ export class GpuFrameCompositor<TSource = FrameLayerPixels> {
...
     );
-    // Submitted here rather than folded into the frame encoder: the effects
-    // pass below and a group above both read this texture through their own
-    // submissions, and only submit order puts the write first.
-    this.device.queue.submit([composeEncoder.finish()]);
+    // Without a frame encoder this has to be submitted here, because the
+    // effects pass below and a group above read this texture through their own
+    // submissions and only submit order puts the write first. With one, record
+    // order does the same job and the frame submits once.
+    if (!frameEncoder) this.device.queue.submit([composeEncoder.finish()]);
```
georgi, line 791: F1 [P1] Keep precomposite uniform allocations distinct until the frame submits

Deferring these submissions makes the existing `core.beginFrame()` at line 770 unsafe: every `composeToTexture` call uses the same `precompCore` and resets its uniform index to zero. `WebGPULayerCompositor.renderBlendPass` then writes into the same GPU buffers while earlier groups' commands are still pending. With two groups whose first children have opacity 0.25 and 0.75, a probe through the real methods shows both passes bound to the same buffer, containing 0.75 by submission time. Different transforms, wipes and matte-source placement are affected too. Reset the precomposite uniform allocator once per submitted frame, and keep each pass's buffers distinct across all group/matte calls. Add a regression with two differently placed/styled groups or a matte whose placement differs from its target.

packages/timeline/src/splitClip.ts
```diff
@@ -21,14 +36,51 @@ function splitAnimations(
...
+  if (delayMs >= splitMs) {
+    return delayMs - splitMs;
+  }
+  const preset = getAnimationPreset(anim.preset);
+  if (preset?.fullClip === true) {
+    return delayMs;
+  }
+  const periodMs = Math.max(1, anim.durationMs);
+  const elapsedInCycleMs = (splitMs - delayMs) % periodMs;
+  return elapsedInCycleMs === 0 ? 0 : periodMs - elapsedInCycleMs;
```
georgi: F2 [P2] Preserve the partial loop cycle immediately after a split

A positive delay to the next cycle boundary suppresses the remainder of the current cycle rather than preserving its phase: `windowT` returns null before `windowStartMs`. Reproduction: a 1000 ms clip with `float`, period 300 ms and delay 20 ms, split at local 200 ms, gives the right half a 120 ms delay. At original times 220/260 ms the unsplit offsets are about 14.03/15.41 px, but the split clip returns 0 for both and only resumes at 320 ms. This introduces a visible snap and pause on a plain split. Carry a phase offset (or an equivalent rebased curve) through compilation instead of waiting for the next cycle. The new test only samples from the next boundary onward, so it misses the lost 120 ms.

packages/video-nodes/src/nodes/timeline/compositeRender.ts
```diff
@@ -437,6 +439,29 @@ export async function renderTimelineComposited(
...
+      if (motionBlur.samplesPerFrame <= 1) return false;
+      const { layers } = computeActiveLayersWithHorizon(
+        sequence.tracks,
+        sequence.clips,
+        timeMs,
+        { canvas, animationCache: animCache }
+      );
+      return shutterWindowIsStatic(
+        layers,
+        hasActiveAnimation(layers, timeMs, canvas, animCache)
```
georgi: F3 [P2] Include animated ancestors in the static-frame decision

`computeActiveLayersWithHorizon` excludes group clips from `layers`, and `hasActiveAnimation` only checks each returned layer's own `clip.animations`. A still image or title moved by an animated parent group therefore passes this check as static, so the render takes one sample and silently drops the requested motion blur. Reproduced with a `float` animation on a group and an unanimated image child: the child's parent-matrix translation changes from 0.02598 at 100 ms to 0.02181 at 110 ms, while this predicate returns true at both instants. The same check in `timeline-preview/frames.ts` has the same problem. Include the contributing ancestor animations (and animated matte sources) before taking the one-sample shortcut, with a grouped-motion regression.

georgi (commented):
Reviewed bd6eba04. Three correctness findings below (F1–F3), which I would fix before merging.

Validation: the existing animation.split and motion-blur-static suites pass (12 tests, Vitest 4.1.8). Separate source-level probes reproduce the split pause and animated-group static classification. A probe using the real composeToTexture/renderBlendPass methods with a mocked GPU boundary confirms the uniform aliasing in F1. Full GPU pixel validation could not run because this environment has no Vulkan driver. I did not run the full monorepo suite.

The PR description also appears stale: several changes listed under “Not in this PR” are present in this head.

packages/timeline/src/render/frameCompositor.ts
georgi: Fixed in 55604d5f. The precomposite core's ring is reset once at the top of `composite`, and `composeToTexture` resets it only on the standalone (no frame encoder) path. Regression: `render.precomposite.gpu.test.ts` "keeps two groups' uniforms distinct inside one frame submit" renders two groups whose first children are 0.25 and 0.75 opaque at different x and asserts both bands; it runs on the lavapipe shard.

---

packages/timeline/src/splitClip.ts
georgi: Fixed in 55604d5f. The right half's loop `delayMs` now goes negative by the elapsed time, and the compiler carries that as `loopOriginMs`: the window still opens at 0, the cycle is counted from the origin, so there is no pause. The split test now samples unsplit-local 200, 220, 260, 320 and 400 against the right half and asserts equality plus that motion is non-zero inside the interrupted cycle; it fails on the previous code.

---

packages/video-nodes/src/nodes/timeline/compositeRender.ts
georgi: Fixed in 55604d5f. `hasActiveAnimation` takes the document clips and walks each layer's `parentId` chain and its matte source before answering; both the server render and the frame preview pass `sequence.clips`. Regression: `render.groupMotion.test.ts` covers a floating group with a still child, a quiet group, and an animated matte source; the first and third fail on the previous code.

---

PR #5593, opened 2026-09-05
feat(godot): 2D game pipeline — slot contract, game nodes, Godot writer, templates, export capability

georgi:
## What changed

NodeTool had no game-development surface: no sprite sheet, tileset, seamless texture or sound-effect nodes, no engine export, and no way to verify a game without an editor. This adds a Godot 4.3 pipeline built as one contract and four independent parts. `@nodetool-ai/protocol` gains the slot manifest a game template declares, the per-slot fill a generation node stamps on its asset, and the acceptance check both sides share. `nodetool.game.SpriteSheet` / `Tileset` / `SeamlessImage` (image-nodes) and `SoundEffect` / `MusicLoop` (audio-nodes) measure a generated asset against its slot, stamp the fill, and persist a stamped asset when the context can create one. `@nodetool-ai/godot` is a pure writer that turns a filled manifest into `SpriteFrames` and `TileSet` resources (atlas regions and collision from the fill, deterministic ids), audio import sidecars, plus a reader and a reference checker. `@nodetool-ai/godot-templates` ships three runnable projects (platformer, top-down, shoot-em-up) with placeholders at the exact paths the writer emits, and a headless runner (import, `--check-only` per script, a smoke scene). The `godot` capability module joins them: `list_game_templates`, `export_godot_project` (template + one asset per slot → a project in the workspace, verified under Godot when a binary is present), and `verify_godot_project`. A `godot-game` skill drives the whole loop in phases. Godot was chosen over Unity/Unreal because glTF is its native 3D format and every project file is plain text, so no SDK or editor plugin is needed.

## Verification

- [x] `npm run test:affected` — the lockfile change selects everything; the touched suites were run directly: protocol (1087 passed), godot (19), godot-templates (33 with real Godot 4.3, 22 + 13 named skips without), image-nodes `game` (14), audio-nodes `game` (7), agents `capabilities-godot` (11, real Godot import + script check + smoke green on an exported platformer), agents registry/coverage/dispatcher (28), cli harness-registry + capability-coverage (43), base-nodes `exported-node-coverage` (1). The 14 base-nodes shader failures in this container are the documented missing-Vulkan-driver case (`No WebGPU adapter available`), unrelated to the diff.
- [x] `npm run typecheck` — web and electron pass; mobile could not run here because `mobile/node_modules` is absent in this container (`Cannot find module 'react-native'`), nothing in the diff touches mobile
- [x] `npm run lint` — exit 0
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — 5/5 selfchecks passed

Checks proven to fail before restore: every `checkSlotFill` rule has a red case in `packages/protocol/tests/game-assets.test.ts`; the Godot golden diff fails on a one-byte edit; `checkGodotProject` fails on a removed copy, a renamed sub_resource id, a removed ext_resource, and a wrong `load_steps`; the templates' `res://` scanner asserts it found references; a syntax-broken `player.gd` fails `checkScripts` and a broken hook fails `verify_godot_project`; the exporter's first real-Godot run failed twice for real reasons (dropped `project.godot`, then a TileSet with no physics layer) before going green.

## Agent capabilities

- [x] `list_game_templates`, `export_godot_project`, `verify_godot_project` name `packages/agents/tests/capabilities-godot.test.ts` in `packages/cli/src/harness/capability-table.ts`; the `godot` surface is in the harness registry

## New checks

- [x] The check was inverted once and observed failing; the failing command is in **Verification** above
- [x] The templates' reference scanner and the base-nodes export audit both assert they found targets

https://claude.ai/code/session_01LhNnoPGvp1gkamhRu2qyCN

---
_Generated by [Claude Code](https://claude.ai/code/session_01LhNnoPGvp1gkamhRu2qyCN)_


georgi (commented):
Found four actionable issues in the generation/export path. F1 blocks stored-image inputs, F2 makes exported top-down levels solid, F3 breaks non-PNG image imports, and F4 discards gameplay edits during the documented art iteration loop.

Validation: ran Godot 4.3 against the shipped top-down template and a copy with the actual writer-generated TileSet. The original smoke test passed; the generated version failed with `SMOKE FAIL: chaser did not move toward the player`. Also reproduced Godot's import failure for WebP bytes copied to `title.png`, and exercised the decoder and layout functions in isolated source-based checks. I did not run the full monorepo test suite.

packages/image-nodes/src/nodes/game.ts
```diff
@@ -0,0 +1,537 @@
...
+async function measureImage(
+  nodeName: string,
+  image: unknown,
+  context?: ProcessingContext
+): Promise<Measured> {
+  const sharp = await loadSharp();
+  if (!sharp) {
+    throw new Error(`${nodeName}: ${SHARP_UNAVAILABLE_MESSAGE}`);
+  }
+  const buf = await decodeImage(image, context);
```
georgi: F1 [P1] Resolve stored image refs through the asset resolver

The documented input is `{type: "image", uri: "asset://<id>.png", asset_id: "<id>"}`, but this `decodeImage` helper only checks inline data and `context.storage.retrieve()`. It never calls `resolveAssetBytes` or consults `assetStorage`. Server contexts use a separate asset adapter, and persisted assets live under `<user_id>/<id>.<ext>`; probing the temporary store with `asset://...` or `/api/storage/<id>.png` cannot read them. Consequently SpriteSheet, Tileset, and SeamlessImage throw `image input is required` for the stored generation outputs the skill tells the agent to pass. Use `loadMediaRefBytes` or the context-aware loader in `image-io.ts`. Add a case with an asset-only ref and separate temp/asset stores; the existing tests supply inline bytes.

packages/godot/src/writer.ts
```diff
@@ -0,0 +1,298 @@
...
+  // Every tile is solid: a ground tileset is what a slot of this kind fills,
+  // and a template's TileMapLayer relies on the physics layer for collision.
+  const hw = w / 2;
+  const hh = h / 2;
+  const polygon = `PackedVector2Array(${-hw}, ${-hh}, ${hw}, ${-hh}, ${hw}, ${hh}, ${-hw}, ${hh})`;
+  const tiles: string[] = [];
+  for (let i = 0; i < fill.count; i++) {
+    const cell = `${i % fill.columns}:${Math.floor(i / fill.columns)}/0`;
+    tiles.push(`${cell} = 0`);
+    tiles.push(`${cell}/physics_layer_0/polygon_0/points = ${polygon}`);
```
georgi: F2 [P1] Preserve non-solid floor tiles in the top-down export

Not every tileset is ground: the shipped top-down template uses tiles 0–3 as walkable floor and tile 4 onward as walls. Its Floor and Walls layers share this TileSet, and Floor covers every room cell. Adding a collision polygon to every tile therefore fills the whole room with solid bodies when export replaces `tiles_floor.tres`. I reproduced this with Godot 4.3: the original template's smoke test passes, but replacing only its TileSet with this writer's output causes `SMOKE FAIL: chaser did not move toward the player`. Carry per-tile collision requirements in the contract or otherwise preserve the template's floor/wall distinction, and cover an exported top-down project in the integration tests.

packages/godot/src/writer.ts
```diff
@@ -0,0 +1,298 @@
...
+      case "tileset": {
+        const out = tileSet(slot, fill);
+        files.push(out.file);
+        copies.push(out.copy);
+        break;
+      }
+      case "image": {
+        copies.push({
+          path: `assets/images/${slotFileStem(slot.slot_id)}.png`,
+          asset_id: slot.asset.asset_id
```
georgi, line 297: F3 [P2] Convert image bytes before assigning a PNG filename

The image nodes preserve the input encoding and persist JPEG/WebP assets with that encoding, while `copyAssets` copies their bytes unchanged. This branch unconditionally names those bytes `.png`; `spriteFrames` and `tileSet` do the same. A valid WebP/JPEG fill therefore passes manifest validation but produces a project Godot cannot import. With real WebP bytes at the generated `assets/images/title.png` path, Godot 4.3 reports `ERR_FILE_CORRUPT`, then fails loading the title texture and main scene. Normalize image outputs to PNG before persistence/export, or preserve their extensions and rewrite the corresponding template references and import sidecars. Cover a real non-PNG input, rather than PNG placeholders only.

packages/agents/src/capabilities/godot.ts
```diff
@@ -0,0 +1,502 @@
...
+    if (rel === "project.godot") {
+      const text = readFileSync(full, "utf8").replace(
+        /^config\/name=".*"$/m,
+        `config/name=${JSON.stringify(name)}`
+      );
+      await workspace.write(under(dir, rel), rewriteRefs(text).text, "text/plain");
+    } else if (REFERENCING_EXTENSIONS.has(ext)) {
+      const { text, changed } = rewriteRefs(readFileSync(full, "utf8"));
+      if (changed) rewritten.push(rel);
+      await workspace.write(under(dir, rel), text, "text/plain");
```
georgi, line 297: F4 [P2] Preserve edited hooks when re-exporting an art change

The new skill's P3 has the agent implement gameplay in the exported hooks, and P4 explicitly instructs it to regenerate an art slot and re-export. This loop always reloads every script and scene from `templateDir` and overwrites the workspace copy, so following that workflow silently resets the user's movement tuning, enemy behavior, and level edits. I verified that an edit to exported `scripts/player.gd` disappears on the next `layOutProject` call. Separate initial template creation from asset refresh, preserving existing hooks/scenes during refresh (including rewriting audio references in those existing files), or add an explicit asset-update operation and use it in P4.

packages/image-nodes/src/nodes/game.ts
georgi: Fixed in 465e46b6. The three image nodes read their input through `loadMediaRefBytes` from `@nodetool-ai/runtime/media-ref-bytes` now, which takes inline bytes, then `asset://` through `resolveAssetBytes`, then a storage key. New case in `packages/image-nodes/tests/game.test.ts`: an asset-only ref against a context whose `resolveAssetBytes` answers and whose `storage.retrieve` returns null asserts the resolver was called once and the temporary store never.

---

packages/godot/src/writer.ts
georgi: Fixed in 465e46b6. The tileset slot spec carries `solid`: `"all"` (the default) or a list of row-major tile indices, refined to stay below `count`. The writer emits a collision polygon only for those tiles, and the top-down manifest names tiles 4 to 11. Coverage: a writer test that asserts tiles 0 to 3 carry no polygon and 4 to 11 do, an exporter test reading the written `tiles_floor.tres`, and the real-Godot integration test now runs `it.each` over platformer, topdown and shmup, each exported from its own placeholders and smoked. The top-down smoke passes with the generated TileSet.

---

packages/godot/src/writer.ts
georgi: Fixed in 465e46b6, on the normalize side. `measureImage` re-encodes anything that is not already PNG with sharp before the fill is stamped, so the inline output and the stored asset are PNG and the `.png` name the writer assigns is true. As a second guard, `loadFilledManifest` refuses an image asset whose content type is not `image/png` and names the slot, so a stale asset stored before this change cannot reach the writer. Tests: a real WebP input to `SeamlessImage` asserts PNG magic on the output bytes and on the stored content, `image/png` as the content type, and sharp decoding it back as a 64x32 PNG; the exporter test seeds an `image/webp` asset and asserts the refusal.

---

packages/agents/src/capabilities/godot.ts
georgi: Fixed in 465e46b6. `export_godot_project` into a directory that already holds a `project.godot` runs in `refresh` mode: writer resources and asset copies are replaced, every template file already present is kept, and audio references inside kept files are still rewritten when a slot's extension moved. The result reports `mode` and `files_preserved`. A new `overwrite` argument restores the template copy. The skill's P4 now says to re-export with the same name and never pass `overwrite` there. Tests: an edited `player.gd` and `player.tscn` survive a second export and are reset by `overwrite: true`; a kept `player.tscn` has `sfx_jump.wav` rewritten to `.ogg` when the refreshed slot is `.ogg`, with no dangling reference.

---

PR #5613, opened 2026-09-05
feat(assets): keep the prompt and generation settings on generated assets

georgi:
## What changed

A generated asset was only reproducible by joining to the ledger row that recorded the call — a separate table with its own retention — or by reading a workflow that may have been edited since. The workflow auto-save kept the prompt and nothing else; the generation seam kept only a `generation_id`. Both write paths now stamp the same fragment, built by `buildAssetGenerationMetadata` in `packages/protocol/src/asset-generation.ts`: the prompt plus provider, model, capability or node type, and the scalar settings (seed, resolution, guidance, voice). Values that cannot be re-run — conditioning bytes, streams, abort signals, nested request objects — are dropped rather than truncated, and the prompt, each string setting and the number of settings are capped. Which model ran is the one setting that is never a scalar, so the actor's `generation_complete` properties filter now carries a model reference through, reduced to `type`/`id`/`name`/`provider`. The asset info panel reads the fragment back with `readAssetGenerationMetadata`: the prompt with a copy button, then the model and each setting, so the recipe is at hand when someone wants another variant.

## Verification

- [x] `npm run test:affected` — the stale `origin/main` ref in this container selects every workspace, so the touched suites were run directly instead: `@nodetool-ai/protocol` (61 files), `@nodetool-ai/kernel` (70 files), `@nodetool-ai/websocket` (256 files, 2914 tests) all pass; `@nodetool-ai/runtime` passes 176/177, the one failure being `tests/providers/video-frame-fallback.test.ts` → `Error: spawn ffmpeg ENOENT`, which reproduces identically with the changes stashed. `web` was covered by `npx jest --findRelatedTests src/components/context_menus/AssetInfoPanel.tsx` — 212 suites, 1814 tests, all passing.
- [x] `npm run typecheck` — `typecheck:web` and `typecheck:electron` pass. `typecheck:mobile` fails on missing modules (`react-native`, `expo-*`): `mobile/node_modules` is absent in this container and `mobile/` is not a root workspace. No mobile file is touched by this diff.
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate` — 4/4 selfchecks passed (workflow-execution, api-server, web-editor surfaces).

Both new checks were inverted once and observed failing:

```
# protocol: dropped the ignored-param filter in buildAssetGenerationMetadata

# web: stopped rendering the settings rows in AssetInfoPanel
✕ shows the prompt, the model and each setting (113 ms)
Tests  1 failed, 2 passed, 3 total
```

## Agent capabilities

No capability is added and no capability's declared contract changes.

## New checks

The diff adds tests, not a rule or an audit.

- [x] Both new suites were inverted once and observed failing; the failing output is in **Verification** above
- [x] Not applicable — no audit that scans files was added

https://claude.ai/code/session_01TDBkBTEfqEXbrfLWomkmVu

---
_Generated by [Claude Code](https://claude.ai/code/session_01TDBkBTEfqEXbrfLWomkmVu)_

### DIFF
--- a/packages/kernel/src/actor.ts
+++ b/packages/kernel/src/actor.ts
@@ -186,12 +186,32 @@ export interface ActorResult {
 }
 
 /**
- * Pick the scalar input properties (string/number/boolean) from a resolved
- * input dict. Drops nested objects, arrays, and binary refs so the
- * `generation_complete` event stays small and carries only persistable
- * generation params (prompt, seed, model, …). Reserved keys (`_control_context`
- * and other `_`-prefixed internals) are stripped. Returns `null` when nothing
- * scalar remains.
+ * A resolved model property (`{type: "image_model", id, name, provider}`),
+ * reduced to the fields that identify it. Which model ran is the one
+ * generation setting that is never a scalar, and without it a saved asset
+ * cannot say what produced it.
+ */
+function modelReference(value: unknown): Record<string, unknown> | null {
+  if (value === null || typeof value !== "object" || Array.isArray(value)) {
+    return null;
+  }
+  const ref = value as Record<string, unknown>;
+  if (!isString(ref.type) || !ref.type.endsWith("_model")) return null;
+  if (!isString(ref.id) || ref.id.length === 0) return null;
+  const out: Record<string, unknown> = { type: ref.type, id: ref.id };
+  if (isString(ref.name)) out.name = ref.name;
+  if (isString(ref.provider)) out.provider = ref.provider;
+  return out;
+}
+
+/**
+ * Pick the persistable input properties from a resolved input dict: scalars
+ * (string/number/boolean) plus model references, reduced by
+ * {@link modelReference}. Drops every other nested object, array and binary
+ * ref so the `generation_complete` event stays small and carries only
+ * generation params (prompt, seed, model, …). Reserved keys
+ * (`_control_context` and other `_`-prefixed internals) are stripped. Returns
+ * `null` when nothing remains.
  */
 function scalarInputProperties(
   inputs: Record<string, unknown>
@@ -205,7 +225,10 @@ function scalarInputProperties(
       isBoolean(value)
     ) {
       out[key] = value;
+      continue;
     }
+    const model = modelReference(value);
+    if (model) out[key] = model;
   }
   return Object.keys(out).length > 0 ? out : null;
 }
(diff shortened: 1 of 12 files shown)


georgi:
Two checks failed on `c7d55dd0`. Both are pre-existing timing races in code this diff does not touch, so I am not fixing either here. Re-running the failed jobs once to confirm; proposed patches below for whoever owns them.

**1. `Quality Gate (npm run check) / test-packages-core` → `@nodetool-ai/runtime` → `tests/run-budget.test.ts`**

```
FAIL  RunBudget > keeps the first reason when a second ceiling is hit later
AssertionError: expected 'deadline' to be 'cost'
 ❯ tests/run-budget.test.ts:321:36
```

The test builds `createRunBudget({ capUsd: 0, deadlineMs: 10 })` and then asserts, on the very next statement, that the first refusal is recorded as `cost`. The deadline clock starts at construction, so any pause longer than 10 ms between those two adjacent statements — a GC pause or scheduler preemption on a shared runner — lets the deadline win the race and the reason is recorded as `deadline`. Nothing in this PR runs between them: the diff's only `runtime` change is inside `ProcessingContext.persistGenerationBytes`, which `run-budget.ts` never calls. The other 3352 tests in that package passed.

Proposed patch: widen the window so the assertion is about ordering rather than about scheduler luck — `deadlineMs: 500` with the later `setTimeout` raised to 600 ms keeps the test's intent (cost is recorded first, deadline expires afterwards, cost still wins) and removes the race.

**2. `Quality Gate (npm run check) / test-packages-nodes` → `@nodetool-ai/audio-nodes` → `tests/audio-transform-decode.test.ts`**

```
FAIL  undecodable mp3 input throws instead of passing through > Normalize throws
Error: Test timed out in 5000ms.
 ❯ tests/audio-transform-decode.test.ts:120:3
```

`Normalize throws` is the first test in that describe block, so it pays the cold start of the whole decode fallback chain in `decodeAudioToWav`: WebAudio via the `node-web-audio-api` native addon, then an ffmpeg spawn. Its five siblings, running warm right after it, took 395–618 ms each. In this same job, `lib-audio-dsp-fallback.test.ts > WebAudio branch (Node)` took **5153 ms** — the cold WebAudio load alone exceeds the 5000 ms default timeout. The diff touches no audio code and no decode path.

Proposed patch: give the package an explicit `testTimeout` in `packages/audio-nodes/vitest.config.ts` (20 000 ms) so the first decode in a file is not racing the default, or warm `loadOfflineAudioContext()` in a `beforeAll`. No assertion changes and nothing skipped.

Verified locally on this head: `@nodetool-ai/audio-nodes` 105/105 pass, and `@nodetool-ai/runtime` 3352/3353 (the single failure being an ffmpeg-absent suite in this container, which reproduces identically with the changes stashed). `@nodetool-ai/protocol`, `@nodetool-ai/kernel` and `@nodetool-ai/websocket` pass in full, as do `lint`, `typecheck`, `bundle`, `docker`, `harness-gate` and both browser E2E legs in CI.

---

PR #5616, opened 2026-09-05
feat(chrome-extension): chat in Chrome's side panel

georgi:
## What changed

Puts the NodeTool chat in Chrome's side panel: a model picker fed from every provider the server has configured, streamed replies, the agent's tool calls as they happen, approval cards for gated calls, and the same conversation threads the web and desktop apps read and write. Chat only — no media generation modes, no workflow editor. The panel is a port of `examples/chat_app`, which is built on `@nodetool-ai/sdk`; the extension is outside the npm workspace and can depend on neither, so the SDK's chat socket and the tRPC procedures the UI needs are mirrored under `chrome-extension/src/lib/` — the same arrangement `src/lib/protocol.ts` already has with `packages/browser`, and both files' headers say so. The toolbar popup keeps the CDP relay controls and gains an **Open chat** button, which attaches CDP to the active tab before opening the panel (`chrome.sidePanel.open` needs a user gesture, which the service worker has no way to supply).

Four commits:

| Commit | What it adds |
|---|---|
| `daa670e6` | The panel: model picker, streaming transcript, tool-call rows, thread list, server settings, the mirrored SDK layer |
| `dc7e9f00` | Tool approvals and permission modes — `tool_approval_request`, `plan_approval_request`, `secret_request` cards, a plan/default/auto picker persisted to storage, agent mode with a Chrome-page system prompt, and Vitest with the extension's first tests |
| `203536b7` | The relay attaches to a *debuggable* tab — Chrome refuses its own pages, other extensions' pages and the Web Store |
| `1d839d5b` | No default model: sending stays blocked until the user picks one the server offers; `execute_code` rendered as a titled code block; copy buttons, starter prompts, Escape-to-close |

Two things the port could not inherit, both found by driving the built extension against a live server rather than by reading:

1. **A turn ends on `chunk.done`, not on the first assistant `message` frame.** The agent loop persists an assistant message carrying only `tool_calls` before any text streams, so ending the turn there drops the placeholder mid-turn and every chunk that follows lands nowhere. `web/src/core/chat/chatProtocol.ts` uses `done` for the same reason. A five-minute safety timeout releases the composer if `done` never arrives, matching the web client.
2. **The server host must match `host_permissions`.** A cross-origin `fetch` from an extension page is CORS-checked, and the server sends no `Access-Control-Allow-Origin` for a `chrome-extension://` origin — so a host outside the manifest's grant fails every HTTP request while the chat WebSocket, which is not CORS-gated, still connects and shows a green dot. `127.0.0.1` joins `localhost` and HTTPS in `host_permissions`; anything else (a LAN box over plain HTTP) goes through `optional_host_permissions` and is requested when the user saves it in the panel's settings.

## Verification

Driven against a real server (`NODETOOL_ENABLE_FAKE_PROVIDER=1 node packages/websocket/dist/server.js`) with the built extension loaded into a real Chromium: the panel filled the model picker from six providers, connected, sent a turn, streamed the answer, rendered the tool chips, listed the thread, and deleted it, with no error banner and no console errors. The CORS finding above came out of that run — `127.0.0.1` was blocked with `No 'Access-Control-Allow-Origin' header is present` while `localhost` returned 200 on every call.

- [x] `npm run typecheck && npm run test && npm run build` inside `chrome-extension/` — clean, 36 tests across 8 files.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 1/1 selfchecks passed`, 65 files / 1176 tests.
- [x] `npm run typecheck` — web and electron clean.
- [x] The seven `npm run check:*` boundary checks, including `check:lockfile` against the new nested `chrome-extension/package-lock.json`.

**The `chunk.done` fix was inverted and observed failing.** Reverting to "end the turn on the first assistant `message` frame", against a replay server reproducing the captured frame ordering (tool-call message immediately, answer chunks at 2s, `done` at 3s):

```
at unlock: {"unlocked":true,"chips":0,"answer":"","tick":0}
saw streamed answer before unlock: false
FAIL: the streamed answer was never on screen
FAIL: composer unlocked at tick 0 with no answer shown
FAIL: expected 1 tool chip at unlock, saw 0
FAIL: the persisted answer did not survive the end-of-turn refetch
```

With the fix in place, the same run:

```
at unlock: {"unlocked":true,"chips":1,"answer":"Hello from the mock turn!","tick":29}
saw streamed answer before unlock: true
PASS: the panel keeps the turn open until `chunk.done`
```

An earlier version of that assertion passed against *both* builds — the live fake provider's turns finish in under a second, so the end state converges either way. That is what the replay server exists for.

### Known red: `test-app`, not this PR's

`TimelineInspector.motionControls.test.tsx` fails 4 of its 14 tests. The diff touches zero `web/` files, and the same 4 fail on the base branch alone at both `b2be1bf1` and `dbc39dad`. Detail and a pointer for whoever owns it are in [this comment](https://github.com/nodetool-ai/nodetool/pull/5616#issuecomment-5558282312). Everything else on the head is green, including `harness-gate`.

## Agent capabilities

## New checks

No check, rule, or audit added. The verification harnesses above are scratch tooling, not committed — `chrome-extension/` is outside CI by design, and the README says what a dedicated workflow would run.

https://claude.ai/code/session_01Lt2rk1C1voKkZDzdPK7duQ


georgi:
`Quality Gate (npm run check) / test-app` is failing on `c10642f0`, and the failure is not this PR's.

**What fails.** `web/src/components/timeline/Inspector/__tests__/TimelineInspector.motionControls.test.tsx` — 4 of its 14 tests, out of 1358 suites / 14990 tests in that leg:

- writes the stagger unit and the unit it starts from
- captions a stagger that runs past the clip
- says when in and out together outrun the clip
- captions an easing this build cannot read

**Why it is not this PR's.** The diff against `main` touches zero `web/` files — `git diff --name-only origin/main...HEAD` is `chrome-extension/**` plus `docs/chrome-extension.md`. The same 4 tests fail identically on the base branch with this PR nowhere in the tree, reproduced at `b2be1bf1` and again at the current tip `dbc39dad`:

```
Test Suites: 1 failed, 1 total
Tests:       4 failed, 10 passed, 14 total
```

It reproduced deterministically both times, so I have not spent a re-run on it: a re-run distinguishes a flake, and this already reproduces on `main` alone.

**No fix exists to port.** `main` has advanced twice since (`574f52bc`, `dbc39dad`); neither touches this suite, and no open PR does.

**Where a fix would go.** The expected copy is implemented — `InspectorMotionFields.tsx:82` holds "Not an easing this build reads; plays linear" — so this looks like a wiring gap rather than missing strings. The sibling test "captions an unparseable easing on the animation's own field" passes with the same setup shape and puts `easing` on the animation itself, while the failing easing case puts it on a keyframe inside `custom.curves`. The other three cover the stagger unit, stagger overrun, and combined in+out overrun. So the caption path appears to be reached for an animation's own fields but not for per-keyframe or stagger-derived ones.

I have left that to the timeline owners rather than widen a chrome-extension PR into motion-graphics code.

**Everything else on this head is green**, including `harness-gate`. That leg is what the merge commit `c10642f0` fixed: it had failed with two capability mapping violations (`preview_timeline_frame`, `render_timeline`), which came from this branch being behind `main` rather than from anything in the diff — the gate reads a branch that lacks main's new contract hashes as having changed them. Merging `main` in cleared it.

---

georgi (commented):
Reviewed the full diff (11,367 lines across 30 files). Well-structured Chrome extension side panel chat.

**Security**: Clean. `react-markdown` without `rehype-raw` prevents XSS. Tool args rendered in `<pre>` via `JSON.stringify`. Help URLs validated to `https:` via `safeHelpUrl()`. Auth token stored in `chrome.storage.local`. No `eval`, no `dangerouslySetInnerHTML`. `ensureHostAccess` properly calls `chrome.permissions.request` for non-localhost hosts.

**Chrome best practices**: Manifest V3, module service worker, minimal permissions (`sidePanel` only new), `optional_host_permissions` for opt-in non-standard hosts. Strong default CSP with bundled React (no inline scripts, no CDN).

**Test coverage**: 8 test files covering CDP tab selection, chat socket framing, popup, model selection, thread auto-selection, tool call rendering, message list, and approval cards.

**Two items**:

1. **Docs inaccuracy**: The docs and PR description claim the panel has no approval surface and sends in permissive mode, but the code defaults to `"default"` (line 7183) and renders `ToolApprovalCard`/`PlanApprovalCard`/`SecretRequestCard`. Update `docs/chrome-extension.md` lines about "no approval cards" and "permissive mode."

2. **Minor bug in `handleStop`**: Plan approvals are filtered by `request.thread_id !== null` but should use `request.thread_id !== threadId` to match the pattern in `finishTurn`. Current logic keeps all approvals with non-null thread_id regardless of which thread is stopped.

`harness-gate` passed (confirmed). `test-app` was still in progress at last check.

---

georgi:
Both review items are fixed in `f29872f9`.

**1. `handleStop` bug.** Confirmed and fixed. The filter kept every request whose `thread_id` was non-null, so stopping a thread left that thread's own approvals standing and dropped only the thread-less ones — the opposite of the intent. Rather than inline `request.thread_id !== threadId`, it now calls `withoutRequestsForThread`, the helper `finishTurn` already applies to all three approval kinds, so the two paths cannot drift again.

**2. Docs inaccuracy.** Three stale claims, all written before the approval surface existed:

- `docs/chrome-extension.md` "Tool calls run without asking … permissive mode" → now describes the picker: Plan (no actions), Default (write/execute/external parks on a card), Auto, starting on Default and remembering the last choice.
- `docs/chrome-extension.md` Limitations, "tool-approval prompts are the full app's job" → now says media generation and the workflow editor are, and that tool, plan and secret approvals do render in the panel.
- `chrome-extension/README.md` `permission_mode: "auto"` / "renders no approval cards" → now names the persisted default and points at `ApprovalCards.tsx`.

The PR description carried the same claim and has been rewritten; it now also lists all four commits and notes the known `test-app` red.

Verified before pushing: `typecheck`, `test` (36 across 8 files) and `build` in `chrome-extension/`, all clean. `grep` confirms no "permissive" or `permission_mode: "auto"` claim survives in the docs, the README, or `App.tsx`'s header comment.

---

PR #5620, opened 2026-09-05
feat(examples): add ten single-job example apps

georgi:
## What changed

The shipped example apps were all studios: several jobs behind one surface. This adds ten apps with the shape of a Runway-style app, one upload and a few choices and one result, each binding existing `nodetool-base` templates: Vary Image, Product Reshoot, Product Shot Video, Multi-Shot Video, Scene Builder, Video Restyle, AI Spokesperson, Upscale Image, Vertical Cut and Ad Maker. To express them, `scripts/build-example-apps.mjs` gains a `video` control bound to a variable, and `text` and `select` accept `{ node, prop }` the way `slider` already did, so a Select can drive a prompt property inside the graph instead of needing an Input node. The eleven existing bundles build byte-for-byte the same. Bundles, preview JSON, the marketing mini-app entries and `docs/plans/example-apps.md` are regenerated or updated to match.

A review pass then fixed a runtime bug the new apps exposed: a widget bound to `op:<id>/prop:<node>#<prop>` was merged into the run graph by node id alone, in three copies of the same loop (`useAppRuntime`, `buildTriggerSubgraph`, and the app-debug simulator). Two operations whose workflows share a node id drove each other's run, which Upscale Image does with `up.scale`. Overlays are now scoped to the operation being run and the three sites share one implementation. The builder also rejects a property binding naming a property the node does not set, Vary Image gains a free-text directions field (the `Edit a Still with Words` template appends it to the instruction), slider ranges stay inside the values the nodes accept, and the ten apps have marketing screenshots so their `/apps/*` entries index.

## Verification

- [x] `node scripts/build-example-apps.mjs` — 21 bundles written and validated through `app debug --no-run`; `--check` reports 0 files out of date; the eleven original bundles are unchanged
- [x] `propTarget` property check proven to fail: a spec pointed at `prop: "strenth"` fails the build with `Vary Image: slider binds property "strenth" on node "ed" (nodetool.image.ImageToImage), which Edit a Still with Words does not set`
- [x] Overlay scoping tests proven to fail before the fix: with the operation filter removed, `packages/execution/tests/app-debug-simulate.test.ts` fails with `expected { scale: 4 } to match object { scale: 2 }` and `web/src/components/appbuilder/__tests__/nodeBinding.test.ts` fails with `Map { "up" => { - "scale": 2, + "scale": 4 } }`; both pass with it restored
- [x] `npm run test --workspace=packages/execution` — 48 files, 444 tests pass
- [x] `npm run test --workspace=packages/base-nodes` — 31 files, 1911 pass; `packages/websocket` — 256 files, 2808 pass; `packages/agents` example-apps-regen and `packages/cli` suites pass
- [x] `cd web && npx jest src/components/appbuilder src/app_preview` — nodeBinding, runtime and preview suites pass
- [x] `npm run typecheck` — web and electron pass; mobile fails only because its separate dependency tree is not installed in this container
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — 7/7 selfchecks passed
- [x] `npm run dev:nodetool -- validate "…/Edit a Still with Words.json"` — valid, 6 nodes, 5 edges
- [x] `marketing`: `node scripts/generate-miniapp-entries.mjs` (21 entries, 21 with screenshots, all indexable) and `node --test scripts/__tests__/*.test.mjs` (13 pass); screenshots captured with `web/scripts/screenshot-app-previews.mjs`

## Agent capabilities

Skipped: no capability added or changed.

## New checks

- [x] The check was inverted once and observed failing; the failing command is in **Verification** above (the builder's property-name check and both overlay-scoping tests)
- [x] Not an audit; the builder check fails on the first bad binding rather than by counting matches

https://claude.ai/code/session_01NjeordLdNTaof3WYUyTSZG


scripts/example-apps/apps.mjs
```diff
@@ -771,5 +773,690 @@ export const EXAMPLE_APPS = [
...
+  // ── 17 ─────────────────────────────────────────────────────────────────────
+  {
+    slug: "video-restyle",
+    name: "Video Restyle",
+    emoji: "🎨",
+    featured: false,
+    tagline: "Repaint a clip in a new style while its motion stays put.",
+    description:
+      "Upload footage, name the look and what must survive, and a video-to-video model applies the style. Strength is the dial: low keeps the original read, high commits to the new look.",
+    note: "🔑 Needs a FAL key (Lucy Edit). Billed per clip.",
```
georgi: [P2] F1: Match the key instructions to the bundled providers

Video Restyle says a FAL key is sufficient, but its bundled `restyle` node uses `provider: "replicate"`, model `decart/lucy-edit-2`. A user following these instructions cannot run it with only FAL configured. The same mismatch occurs in AI Spokesperson (the voice runs `inworld/realtime-tts-1.5-max` through Replicate, not an Inworld key) and Multi-Shot Video (the keyframe runs `gpt-image-2-text-to-image` through KIE, not OpenAI). Correct these three notes and regenerate the app bundles, previews and marketing entries so setup instructions name the providers execution actually uses.

scripts/example-apps/apps.mjs
```diff
@@ -771,5 +773,690 @@ export const EXAMPLE_APPS = [
...
+              "claymation, soft studio light, fingerprints in the clay, stop-motion cadence",
+              "black and white 16mm documentary, heavy grain, high contrast",
+              "oil painting, thick impasto brushwork, warm gallery light",
+              "neon cyberpunk, wet reflections, magenta and cyan rim light",
+              "pencil sketch on paper, cross-hatched shading, visible paper grain"
+            ]
+          },
+          { text: "preserve", op: "restyle", label: "What must survive" },
+          {
+            slider: { node: "restyle", prop: "strength" },
```
georgi: [P2] F2: Remove strength controls that the selected models ignore

This slider changes `restyle.strength`, but Lucy Edit 2 declares no strength input. `ReplicateProvider.videoToVideo()` explicitly removes it via `pruneToDeclaredInputs()`, so 0.1 and 1 produce identical request parameters despite the app promising different preservation levels. Vary Image's new `How far to go` slider has the same problem: `fal-ai/nano-banana/edit` declares no strength, and `FalArgsBuilder.set()` drops it. Confirmed by executing both provider helpers against the shipped manifests. Remove these controls and the strength-dependent claims, or bind them to an effective prompt parameter/use a model that supports strength.

scripts/example-apps/apps.mjs
```diff
@@ -771,5 +773,690 @@ export const EXAMPLE_APPS = [
...
+              "Slow tilt up from the ground to the sky"
+            ]
+          },
+          {
+            slider: { node: "vid", prop: "duration" },
+            op: "motion",
+            label: "Seconds",
+            min: 2,
+            max: 6,
+            step: 1,
```
georgi: [P2] F3: Offer durations supported by Scene Builder's LTX model

The bundled motion workflow selects `fal-ai/ltx-2.3/image-to-video/fast`, whose manifest supports 6, 8, 10, 12, 14, 16, 18 and 20 seconds. This slider offers 2, 3, 4 and 5 seconds, none of which the model can generate. `FalArgsBuilder.set("duration", value)` drops those values, leaving the endpoint's six-second default, so choosing a shorter clip still produces and bills a six-second generation. Use supported duration choices (for example 6/8/10), or explicitly trim the generated video if shorter outputs are intended.

georgi (commented):
Reviewed 71ec1170fd44a8175165151ac29a10965345cfd4. Three actionable findings in the new app definitions, detailed inline. The operation-scoping change looks correct on inspection.

Verification: `node scripts/build-example-apps.mjs --check --skip-validate` passes with zero files out of date. Executed the actual FalArgsBuilder and Replicate pruneToDeclaredInputs implementations in isolation against the shipped manifests: both edit-strength parameters disappear from provider requests, and Scene Builder's requested durations are dropped. No paid provider calls or full test-suite run.

scripts/example-apps/apps.mjs
georgi: Fixed in 183f6739. Confirmed against the bundled graphs: `restyle` is `provider: "replicate"` (Lucy Edit 2), `speech` is Replicate (Inworld TTS), `keyframe` is KIE (GPT Image). The three notes now name those providers, and bundles, previews, marketing entries and screenshots are regenerated.

---

scripts/example-apps/apps.mjs
georgi: Fixed in 183f6739. Both strength sliders are removed along with the "strength is the dial" copy; the manifests confirm neither `fal-ai/nano-banana/edit` nor `decart/lucy-edit-2` declares a strength input. The same class of bug hit AI Spokesperson's pace slider: the Replicate TTS path sends `speed` while `inworld/realtime-tts-1.5-max` declares `speaking_rate`, so that slider is removed too.

---

scripts/example-apps/apps.mjs
georgi: Fixed in 183f6739. The slider now offers 6, 8 and 10 seconds (step 2, default 6), the values the fast endpoint's `duration` enum accepts without the 25 fps requirement that 12 and above carry.

---

PR #5621, opened 2026-09-05
feat(timeline): DAW-style MIDI tracks (model, synth, playback, render, agent ops, tempo grid, instruments, piano roll)

georgi:
## What changed

Adds MIDI tracks to the timeline as a DAW-style music surface, in three slices on one branch.

**A1 (model, synth, three audio hosts, agent ops).** A `midi` track owns an instrument; a `midi` clip carries its notes inline, in ticks (PPQ 960) relative to the clip's content start, so split and trim behave like an audio clip's region over its source (the notes are copied, the in-point moves, hidden notes survive). A constant document `tempo` converts ticks to milliseconds; milliseconds stay the master clock, and `set_tempo` rescales midi clips around the tempo offset while leaving every other clip where it is. Sound comes from one dependency-free voice renderer in `packages/timeline/src/midi` (polyBLEP saw and square, ADSR, biquad lowpass, deterministic, gated at the clip window) whose buffer feeds the existing audio pipeline unchanged: `AudioGraph` in the browser for playback and export, and a WAV into the ffmpeg mix for the server `RenderTimeline` node. The agent gets `add_midi_clip`, `set_notes`, `set_tempo` and `set_track_instrument` on both the server capability and the `ui_timeline_*` twin, the validator learns the midi rules, and `timeline-validate` gains a selfcheck so the harness gate reaches the new suites.

**A2 (editing without a piano roll).** Pure functions for transpose, quantize (six divisions, strength, start or start-and-length) and velocity scaling, six instrument presets, and bars:beats arithmetic with a tempo grid generator. Project settings gain a tempo group; the ruler can show bars and beats with a selectable grid division, and snapping reads the tempo grid whenever a midi track exists or the ruler is in bars. A midi track header carries a preset select and an inline instrument editor that auditions each change through the same voice renderer the clip plays with. The inspector's midi section transposes, quantizes and scales velocity as single undo entries. The agent gets `transpose_clip`, `quantize_notes` (reporting moved notes) and `scale_velocity` on both surfaces, `set_track_instrument` accepts a preset id, and `get_state` reports the matching preset and each midi clip's start in bars and beats.

**A3 (piano roll).** A midi clip opens a clip editor below the tracks region (double-click, the clip menu, or the inspector): keyboard column, bars:beats ruler over the content, the note grid on one canvas with the clip window shaded, a velocity lane, and a playhead from the playback time channel. Click adds and auditions a note, drag moves or resizes the selection snapped to the grid (Alt bypasses), marquee selects, and the keyboard nudges, duplicates and deletes; every gesture is one undo entry. The note arithmetic (add, move as a clamped group, resize, duplicate, rectangle hit) lives in `packages/timeline/src/midi/notesEdit.ts` so any surface edits notes by the same rules. On a phone the editor replaces the tracks region.

**Review fixes (fa9c51fa).** The render worker is spawned from a static `import.meta.url` in its own module, mapped to a stand-in under Jest so the worker path is tested. A rendered buffer is already the clip's window, so it plays from source offset zero rather than seeking by the in-point again. The first midi track, and a loaded document with midi tracks and no tempo, record the default tempo explicitly, so an undo after the first tempo change saves a value the PATCH merge can read.

Still to come: `.mid` import and export, a sampler instrument, Web MIDI recording (A4).

## Verification

- [x] `npm run test:affected` selects every workspace because `protocol` changed. Package suites through turbo: every touched package green (`protocol`, `timeline`, `agents`, `execution`, `cli`, `websocket`, `models`, `video-nodes`). The remaining failures are environment-only and in code this diff does not touch: `base-nodes` and `image-nodes` fail on `No WebGPU adapter available` (no Vulkan driver in the container, CI installs lavapipe), `runtime` `video-frame-fallback` and `video-nodes` `timeline-time-remap-decode` fail on `spawn ffmpeg ENOENT`. App suites: web green (the inspector suite that was red on `main` passes since #5637 was merged in), electron 63/63, mobile 79/79.
- [x] `npm run typecheck`: web and electron clean. Mobile reports 36 errors in files this diff does not touch (`app_runtime`, `graph_editor`, storyboard); the same 36 reproduce on `origin/main` in a clean worktree, so they predate this branch.
- [x] `npm run lint`: exit 0, including the enforced anti-slop config and `lint:design`.
- [x] `npm run dev:nodetool -- harness gate --base origin/main`: every selfcheck passes, including the new `timeline-validate` selfcheck and `capability-suites`.

## Agent capabilities

- [x] `npm run capabilities:check` passes (table re-synced after the `edit_timeline` contract grew).
- [x] The seven ops are covered by `packages/agents/tests/capabilities-timeline-midi.test.ts` and the mapped `timelines-op-input.test.ts`: an agent-authored four-note phrase, tempo rescale, refused bad notes and wrong-track instrument, transpose with the pitch cap, quantize with the moved count, velocity clamp, preset resolution and refusal. The headless `nodetool timeline debug --interact` replay (add midi track, off-grid phrase, quantize, transpose, preset) validates clean and is pinned as a test against the real bridge and validator.

## New checks

- [x] Each validator rule (`midi_clip_off_midi_track`, `non_midi_clip_on_midi_track`, `midi_notes_invalid`, `midi_clip_retimed`, `midi_clip_silent`, `midi_tempo_missing`) has a firing and a silent case in `packages/execution/tests/timeline-midi.test.ts`, plus the case that quantize output validates clean. The scene-model midi skip and the server mix cases were each inverted once and observed failing. The bars:beats grid test fails when `barStartMs` drops the offset; the piano roll add and drag tests were inverted once each and observed failing.
- [x] The tempo round-trip is pinned at both persistence seams: `packages/models/tests/timeline-sequence.test.ts` (row to sequence) and `packages/websocket/tests/trpc-timeline.test.ts` (update merge to get), plus the store-side save → undo → save payload case.
- [x] The worker path of `midiRender` runs under Jest through the mapped stand-in; the windowed-buffer scheduling has the split-clip case at the head and mid-seek.

https://claude.ai/code/session_01Udggi5do5iAXWPMQQyXeNn


georgi:
Reviewed head `412a0457e0fbb4443b309cc94377e5ac0c835ead`. I recommend fixing these three issues before merging.

1. **F1 [P1] MIDI playback resolves the worker relative to the page, so synthesis fails in browsers.** [midiRender.ts:65–72](https://github.com/nodetool-ai/nodetool/blob/412a0457e0fbb4443b309cc94377e5ac0c835ead/web/src/components/timeline/preview/midiRender.ts#L65-L72)
   `new Function("return import.meta.url")` throws a syntax error even when called from an ES module: the constructed function is parsed as script code. Consequently this always falls back to `location.href`. The actual `getMidiClipBuffer` path, with Worker present and a page URL of `https://app.example.com/timelines/123`, constructed `https://app.example.com/timelines/midiRender.worker.ts`, rather than the module's worker. The dynamic base also prevents the normal Vite worker URL transformation. The inline fallback only covers an absent Worker, and PreviewArea swallows the failed render, leaving MIDI clips silent. Use a statically analyzable module-relative worker URL or a Vite worker import, and test the Worker-present path. The added tests explicitly exercise only the Worker-absent branch.

2. **F2 [P1] Trimmed/split MIDI buffers have their source offset applied twice in browser playback and export.** [AudioGraph.ts:476–478](https://github.com/nodetool-ai/nodetool/blob/412a0457e0fbb4443b309cc94377e5ac0c835ead/web/src/components/timeline/preview/AudioGraph.ts#L476-L478)
   `renderMidiClip` already renders just the clip window, starting the resulting buffer at zero. However, the buffer-backed path retains the original clip and subsequently calls `timeRemapAudioSegments(clip)`, which uses `clip.inPointMs` as `sourceStartMs`. I ran the actual renderer and AudioGraph with `startMs=2000, inPointMs=2000, durationMs=2000` and a note inside that window: it produced an audible two-second buffer, then scheduled `source.start(2, 2, 2)`, starting at the buffer's end. This silences the right half of a midpoint split; smaller in-points skip notes and truncate playback. Schedule rendered MIDI windows from source offset zero, while retaining the timeline position and mid-clip seek offset. Add a nonzero-in-point scheduling/export case. The server's `audioClipFilter` already handles this distinction.

3. **F3 [P2] Undoing the first tempo change cannot persist the restored tempo.** [timeline router:330](https://github.com/nodetool-ai/nodetool/blob/412a0457e0fbb4443b309cc94377e5ac0c835ead/packages/websocket/src/trpc/routers/timeline.ts#L330)
   A new editor-authored MIDI document has `tempo: undefined` and plays at the implicit 120 BPM. Set it to 60, let autosave finish, then undo. The undo snapshot restores undefined and the original clip positions/durations (as the new store test asserts). `buildTimelineDocumentPayload` emits undefined, JSON omits the field, and `input.document.tempo ?? current.tempo` retains the server's 60 BPM while saving the restored clips. Reopening therefore plays the restored notes at 60 rather than 120 BPM. A focused payload/merge probe confirmed this mismatch. Persist an explicit default when restoring the implicit tempo, or define a clear/reset representation at the API boundary, and test save → undo → save → reload.

Validation: focused execution of the actual MIDI renderer and AudioGraph with a mock audio context, the actual Worker-construction path with a recording constructor, and the payload/router tempo merge. I did not run the full application suites or a browser end-to-end test. GitHub also currently reports merge conflicts with the base branch.

georgi (commented):
Reviewed the diff (100 files, +10,422 lines). Well-structured feature with good architecture:

- Clean layering: protocol → timeline (pure functions) → execution (validator) → agents (bridge ops) → websocket (persistence) → web (stores, components). No layer reaches down.
- MIDI modeled as a new track type extending the existing discriminated union, reusing clip windowing model.
- Voice renderer is pure arithmetic, deterministic, dependency-free. Same code runs in browser (Web Worker) and server (WAV for ffmpeg).
- Milliseconds stay the master clock. Ticks are content-relative, converted through tempo. Existing audio/video clips unaffected.
- ~3,400 lines of tests across 20 new test files.
- All new components use `ui_primitives`. No raw MUI imports.

**Blockers before merge**:

1. **Merge conflict**: `mergeable_state` is `dirty`. Needs rebase onto current main.
2. **Cross-PR conflict**: PR #5628 (timeline zip export) touches overlapping timeline files. Whichever merges second will need a rebase.

**Minor note**: `new Function("return import.meta.url")` in `midiRender.ts` matches the existing pattern in `histogramAsync.ts`. Flagging for visibility but not blocking.

This is too large for confident line-by-line review in a single pass, but the architecture, test coverage, and convention adherence are all solid.

---

georgi:
All three fixed in fa9c51fa, on top of a merge of the current `main`.

**F1 worker URL.** The worker is now spawned from a static `new URL("./midiRender.worker.ts", import.meta.url)` in its own module, `preview/midiRenderWorkerClient.ts`, the same shape as `lib/workflow/browserWorkerClient.ts`, so Vite rewrites it into the built worker asset. Jest maps that module to a stand-in that answers a request the way the worker does (`__mocks__/midiRenderWorkerClient.ts`), and `midiRender.test.ts` now runs the Worker-present path: it asserts one worker was spawned and terminated and that the samples equal the inline render. The `new Function("return import.meta.url")` pattern this copied from `utils/histogram/histogramAsync.ts` has the same defect there; I left that file alone as out of scope.

**F2 double offset.** `AudioGraph.addClips` marks a caller-supplied buffer as windowed and schedules it over one stretch from source offset zero at unit rate, keeping the timeline position and measuring a mid-clip seek into the window alone. `AudioGraph.midi.test.ts` adds the split case you ran (`startMs=2000, inPointMs=2000, durationMs=2000`): it now schedules `start(2, 0, 2)` from the head, and `start(0, 1, 1)` when the playhead is one second into it. The export path goes through the same call, so it is covered by the same change.

**F3 tempo after undo.** The store now records the default tempo explicitly when the first midi track is added, and `loadSequence` does the same for a document that has midi tracks and no tempo, so no midi document plays at an implicit tempo any more. The undo test now asserts the restored tempo is the explicit 120 rather than `undefined`, and a new case runs set 60 → payload → undo → payload and asserts the second payload carries 120, which is the value the router merge you traced then stores. The router's own test already pins that a defined tempo wins the merge.

One thing to flag: `web`'s `TimelineInspector.motionControls.test.tsx` fails four cases on this branch, and the same four fail on a pristine `origin/main` checkout; `main`'s Test run on `dbc39dad` is red on the same suite (test-app leg, line 323). It comes from #5584 and touches nothing this PR changes, and I saw no fix on `main` yet, so I have not ported one.

---

georgi:
Root cause of the `TimelineInspector.motionControls` failures that block CI here and on `main`, since auto-merge is now waiting on them:

`aeda2232` (in #5584) added the stagger unit and stagger-from selects, the "Staggered over N words this runs Nms" caption, the "In and out together run Nms" caption, and the per-keyframe "Not an easing this build reads; plays linear" caption to `web/src/components/timeline/Inspector/ClipAnimations.tsx`, with the test in the same commit. That branch then merged `main`, which had `1ffe6dfb` (the inspector memoization rewrite of the same file), in `f7e5fe6d`; the conflict in `ClipAnimations.tsx` was resolved on `main`'s side, so the memoized component kept only the word-offset field (`… animation word stagger offset`) and lost the four controls the test still expects. `main` at `dbc39dad` fails the same four cases for that reason.

The fix belongs on `main`: restore the four pieces from `aeda2232`'s `ClipAnimations.tsx` (lines 429, 437, 462 and 602 in that revision) into the memoized structure. It is unrelated to this PR's diff, so I have not ported it here.

On the review's other blocker: the conflict with `main` is resolved (merged twice, at `412a0457` and `2abd50ed`), and `mergeable_state` is back to "blocked" on checks. I will merge `main` again if #5628 lands first.

---

PR #5626, opened 2026-09-06
feat: add opt-in MuAPI image and video nodes

Anil-matcha:
## Summary

- Add an opt-in `@nodetool-ai/muapi-nodes` pack for MuAPI image and video workflows.
- Register text-to-image, text-to-video, and image-to-video nodes with the built-in pack catalog.
- Use MuAPI's asynchronous submit/poll API with safe output downloads and secret-backed authentication.

## Verification

- Strict TypeScript source check passed.
- 8 provider-pack tests passed with mocked HTTP responses.
- `git diff --check` passed.
- No live MuAPI generation request was made.

## Links

- MuAPI API reference: https://muapi.ai/docs/api-reference
- MuAPI access keys: https://muapi.ai/access-keys

### DIFF
--- a/packages/muapi-nodes/src/muapi-base.ts
+++ b/packages/muapi-nodes/src/muapi-base.ts
@@ -1,0 +1,383 @@
+import type { ImageRef, VideoRef } from "@nodetool-ai/node-sdk";
+import {
+  assertSafePublicHttpsUrl,
+  detectImageMime,
+  loadMediaRefBytes,
+  safeFetch,
+  type ProcessingContext
+} from "@nodetool-ai/runtime";
+import {
+  fetchWithRetry,
+  imageRefFromBytes,
+  pollUntilTerminal,
+  TERMINAL_FAILURE_STATES,
+  TERMINAL_SUCCESS_STATES
+} from "@nodetool-ai/runtime/provider-transport";
+
+export const MUAPI_BASE_URL = "https://api.muapi.ai/api/v1";
+
+export const MUAPI_IMAGE_MODELS = [
+  "flux-3-text-to-image",
+  "flux-3-dev"
+] as const;
+
+export const MUAPI_IMAGE_ASPECT_RATIOS = [
+  "16:9",
+  "9:16",
+  "1:1",
+  "4:3",
+  "3:4",
+  "2:3",
+  "3:2",
+  "21:9"
+] as const;
+
+export const MUAPI_VIDEO_ASPECT_RATIOS = [
+  "16:9",
+  "9:16",
+  "1:1",
+  "4:3",
+  "3:4",
+  "21:9"
+] as const;
+
+export const MUAPI_IMAGE_RESOLUTIONS = ["1k", "2k", "4k"] as const;
+export const MUAPI_VIDEO_RESOLUTIONS = ["480p", "720p", "1080p"] as const;
+
+export function normalizeMuapiVideoDuration(value: unknown): number {
+  const duration = typeof value === "number" ? value : Number(value);
+  if (!Number.isFinite(duration)) return 5;
+  return Math.min(10, Math.max(4, Math.round(duration)));
+}
+
+const MUAPI_FAILURE_STATES = new Set([
+  ...TERMINAL_FAILURE_STATES,
+  "failure"
+]);
+
+const OUTPUT_KEYS = [
+  "url",
+  "image",
+  "image_url",
+  "video",
+  "video_url",
+  "file_url",
+  "output",
+  "outputs",
+  "images",
+  "data",
+  "result",
+  "prediction",
+  "media",
+  "content",
+  "file"
+] as const;
+
+export interface MuapiPollOptions {
+  pollIntervalMs?: number;
+  maxAttempts?: number;
+  signal?: AbortSignal;
+}
+
+export interface GenerateMuapiMediaOptions {
+  apiKey: string;
+  endpoint: string;
+  payload: Record<string, unknown>;
+  kind: "image" | "video";
+  pollIntervalMs?: number;
+  maxAttempts?: number;
+  signal?: AbortSignal;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+function readString(value: unknown): string | undefined {
+  if (typeof value !== "string") return undefined;
+  const trimmed = value.trim();
+  return trimmed || undefined;
+}
+
+function readIdentifier(value: unknown): string | undefined {
+  const stringValue = readString(value);
+  if (stringValue) return stringValue;
+  if (typeof value === "number" && Number.isFinite(value)) {
+    return String(value);
+  }
+  return undefined;
+}
+
+async function readJsonResponse(
+  response: Response
+): Promise<{ body: unknown; text: string }> {
+  const text = await response.text();
+  if (!text.trim()) return { body: {}, text };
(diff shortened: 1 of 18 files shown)


packages/muapi-nodes/src/muapi-base.ts
```diff
@@ -0,0 +1,383 @@
...
+  pollUntilTerminal,
+  TERMINAL_FAILURE_STATES,
+  TERMINAL_SUCCESS_STATES
+} from "@nodetool-ai/runtime/provider-transport";
+
+export const MUAPI_BASE_URL = "https://api.muapi.ai/api/v1";
+
+export const MUAPI_IMAGE_MODELS = [
+  "flux-3-text-to-image",
+  "flux-3-dev"
```
georgi, line 21: [P2] Use available image endpoints before exposing Text to Image

Both selectable image routes, flux-3-text-to-image and flux-3-dev, are explicitly marked coming soon in [MuAPI's FLUX 3 catalog](https://muapi.ai/flux-3). This leaves the new Text to Image node with no documented live model, including its default. The mocked submission test succeeds regardless of real endpoint availability. Please use a documented live image route with its supported payload, or defer registering the image node until these routes are available and provide provider-backed contract evidence.

packages/muapi-nodes/src/muapi-base.ts
```diff
@@ -0,0 +1,383 @@
...
+  signal?: AbortSignal
+): Promise<string> {
+  const init: RequestInit = {
+    method: "POST",
+    headers: jsonHeaders(apiKey),
+    body: JSON.stringify(payload)
+  };
+  if (signal) init.signal = signal;
+
+  const response = await fetch(apiUrl(endpoint), init);
```
georgi, line 160: [P2] Register the new transport in the URL-egress inventory

The existing packages/runtime/tests/url-egress-audit.test.ts scans every packages/*/src file. Its PLAIN_FETCH regex matches both fetch(apiUrl(...)) calls here, and its screening scan matches safeFetch/assertSafePublicHttpsUrl, but packages/runtime/tests/url-egress-inventory.ts contains no entry for this new file. Consequently the plain-fetch classification, non-guarded-entry, and screened-file classification assertions fail when the audit runs. Add the appropriate fixed-host and guarded entries and document the guarded surface in docs/url-egress-inventory.md, then run the runtime URL-egress audit as well as the pack tests.

georgi (changes requested):
Reviewed all 18 changed files at b2e0b0a6, including dependency/lockfile changes, registration, credential handling, and submit/poll/download behavior. Two issues need addressing before merge, detailed inline.

Security scope: no evidence of malicious code in this diff, no new third-party dependency or install hook, and no workflow/CI changes. The pack is disabled by default, only reads MUAPI_API_KEY, and downloads outputs through the existing safeFetch wrapper. This is a source review, not an endorsement of the external provider.

Validation: git diff --check passed. Reproduced the new file's matches against the existing URL-egress audit regex and confirmed the inventory has no MuAPI entry. Full Vitest/build and live generation were not run. The four PR workflow runs report action_required, so CI has not validated this revision. Please also add tests exercising each node's actual process() payload against the documented provider contract; the current mocks do not establish endpoint availability.

georgi:
The current PR adds provider-specific workflow nodes, but that does not make MuAPI available across NodeTool. Please implement MuAPI as a first-class runtime provider and keep any provider-specific nodes as optional thin wrappers.

Suggested scope:

1. Add `PROVIDER_IDS.MUAPI = "muapi"` in `packages/protocol/src/api-types.ts` and include it in the applicable full/self-serve provider policy sets.
2. Add `MuapiProvider extends BaseProvider` under `packages/runtime/src/providers/`, export it, and register it with `registerBuiltinProvider(PROVIDER_IDS.MUAPI, MuapiProvider, { MUAPI_API_KEY: "" })`.
3. Implement model discovery through `getAvailableImageModels()` and `getAvailableVideoModels()`. Return only documented live endpoints and set accurate `supportedTasks` and per-model constraints. At minimum, implement the currently live `text_to_video` and `image_to_video` routes. Do not expose the FLUX 3 image routes while MuAPI marks them coming soon.
4. Implement the standard generation methods, `textToImage`, `imageToImage`, `textToVideo`, and `imageToVideo`, only where a live model supports them. Translate NodeTool's canonical params to MuAPI payloads inside the provider. Reuse the submit/poll/upload/download transport, preserving single-attempt billable POSTs, cancellation, GET retries, redirect-aware `safeFetch`, response-size limits, and media-type validation.
5. Preserve provider metadata. Record MuAPI's request ID and actual response cost where the runtime generation/tracing seam expects them, and sanitize provider error bodies so credentials or signed URLs cannot reach logs.
6. Wire credential verification if MuAPI exposes a harmless authenticated account/balance endpoint, and add the provider to settings/display-name mappings.
7. Add the required URL-egress inventory and reader-facing documentation entries for every fixed-host and provider-returned URL surface.
8. Add contract tests for model discovery and every implemented capability, including payload mapping, upload shape and 10 MB image limit, terminal states, abort during polling, timeout, redirect/SSRF rejection, oversized or wrong-type output, transient download retry, and no retry of job-creating POSTs.

Why this matters: NodeTool's generic image/video nodes call `ProcessingContext.runProviderPrediction`, and the model APIs enumerate `BaseProvider.getAvailable*Models()`. The chat media composer, inference endpoints, CLI generation, agent media capabilities, model selectors, storyboard/sketch/timeline generation, and generic workflow nodes all converge on that provider seam. Registering MuAPI there makes it available across those surfaces without duplicating transport and parameter logic in separate nodes.

Acceptance criterion: after configuring `MUAPI_API_KEY`, a live MuAPI video model appears in the shared model catalog and can run through the generic Text to Video and Image to Video nodes, chat generation, the media inference endpoint, CLI generation, and agent media tools, with one shared provider implementation and passing generation-seam and URL-egress audits.

georgi (changes requested):
Good-quality contribution that follows the repo's node patterns correctly. Two blocking issues:

**1. Missing URL egress inventory entries (will fail the egress audit test)**

`muapi-base.ts` uses bare `fetch()` in `submitMuapi()` and `uploadMuapiImage()`. These are fixed-host calls to `api.muapi.ai` (safe from SSRF), but `packages/runtime/tests/url-egress-audit.test.ts` scans all `packages/*/src/*.ts` for bare `fetch(<var>)` calls and requires classification.

Add entries to the inventory:
- `fixed-host` entries for the API calls (submit + upload) addressing `api.muapi.ai`
- A `guarded` entry for the download path (which correctly uses `safeFetch`)
- Update `docs/url-egress-inventory.md` with the guarded surface

**2. Missing tsconfig reference**

`tsconfig.json` only references `../node-sdk` but the package depends on `@nodetool-ai/runtime`. Add `{ "path": "../runtime" }` to avoid potential `tsc --build` ordering issues.

**What's good:**
- Node structure follows BaseNode + `@prop` decorators correctly
- API key handling matches `atlascloud-base.ts` pattern
- `downloadMuapiOutput` properly uses `safeFetch` for redirect checking
- `uploadMuapiImage` validates upload URL with `assertSafePublicHttpsUrl`
- Submit is correctly not retried (non-idempotent POST)
- Tests cover API key resolution, happy path, failure states, upload multipart, node registration
- Package structure, workspace registration, and builtin-packs registration are all correct
- No junk files

---

georgi:
@claude implement the provider as described in the comment above

georgi:
continued in https://github.com/nodetool-ai/nodetool/pull/5666

PR #5641, opened 2026-09-06
fix(web): put the output node's input handle on its edge and header row

georgi:
## What changed

The Output node's input handle sat 8px clear of the node's left edge and ~15px below the header, floating in blank body. Two causes, both fixed. `--node-body-padding: 8px` was a claim the body could not keep — `.react-flow__node .node-body` (nodes.base.css) sets `padding: 0` and outranks the emotion class, so the body renders flush while left handles still offset themselves by `-6px - var(--node-body-padding)`; the var now says `0px`, matching what the body actually renders. And the floating `HandleColumn` measures `spacing(4)` from a positioned ancestor that, in `OutputNode`, contains the header — an offset meant to clear a header applied on top of one; a new `layout="header"` variant centers the first handle row on the 24px header instead.

## Verification

Measured in Chromium against the seeded `wf-editor-journey` graph (journeys backend, `PLAYWRIGHT_CHROMIUM_PATH=/opt/pw-browsers/chromium-1194/chrome-linux/chrome`) with a throwaway spec that read bounding boxes, then deleted:

| | node body | header | handle |
|---|---|---|---|
| before | x 1070, w 280 | y 241–265 | x 1057–1063, y 262–274 |
| after | x 1070, w 280 | y 241–265 | x 1065–1071, y 249–261 |
| base node (`prompt_input`, unchanged) | x 610, w 320 | — | x 605–611 |

After the change the handle ends at the body's inner border edge (1071 for a body at x=1070), the same relationship a base node has (611 for a body at 610), and its center sits on the header row instead of 15px below it. Computed padding confirmed the mismatch directly: `padding: 0px | var=8px` before, `0px | 0px` after.

- [x] `npm run test:affected` — the two failures are environmental and reproduce without this diff: `@nodetool-ai/image-nodes` fails 52 shader tests with `No WebGPU adapter available (Node/Dawn)` (no Vulkan ICD in this container, documented in AGENTS.md), and `TimelineInspector.motionControls.test.tsx` fails 4 tests on a clean tree too.
- [x] `npm run typecheck` — web and electron clean; `mobile` fails on missing `react-native`/`expo` modules, which is the un-installed mobile tree, not this diff.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — Gate: 22/22 selfchecks passed.

## Agent capabilities

## New checks

https://claude.ai/code/session_01XbFK46mCr9hkQqWjkffioT

---
_Generated by [Claude Code](https://claude.ai/code/session_01XbFK46mCr9hkQqWjkffioT)_


georgi:
`Quality Gate (npm run check) / test-app` failed on 394837cd with 4 failures in `web/src/components/timeline/Inspector/__tests__/TimelineInspector.motionControls.test.tsx`. Not this PR's: the diff touches three node-editor files and nothing under `timeline/`, and the same 4 tests fail locally on the base commit with this branch's changes stashed.

The fix was already on main — #5637, "restore the motion controls a merge dropped", which rewrites `ClipAnimations.tsx` and `ClipCustomCurves.tsx`, the components those tests assert on. Merged main into the branch (810ff6a6) rather than waiting for it. That suite now passes locally (14/14), as do the 241 suites related to the changed files, `tsc --noEmit` for web and electron, `npm run lint`, and `harness gate`.

---

PR #5645, opened 2026-09-06
feat(recipes): ship each recipe as an app, not a pile of workflows

georgi:
## What changed

A recipe was a list of separate example workflows: to run the chain you copied four to six graphs into the library and carried each result to the next by hand. Each recipe now names the shipped example apps that surface it, and the first one is built for that chain — its operations are the recipe's steps, and the handoffs the manifest describes in prose are variables carried between them. Four new example apps, one per shipped recipe: **Trailer Room** (premise → beat sheet → shot list → scored teaser, where the cut the video step produces feeds the score step), **SKU Factory** (one packshot drives the cutout, studio scene, seasonal relight and listing copy; the turntable and print master stay on their own buttons because they cost more), **Dubbing Desk** (transcribing writes the script into an editable field the revoice, subtitle and lip-sync steps all read), and **Viral Ad Engine** (one offer drives both the copy registers and the hook-and-thumbnail set). Each recipe also names a companion app that covers part of the chain on purpose — Film Studio, Product Reshoot, AI Spokesperson, Brand & Social. The new `apps` field on the recipe bundle resolves against the shipped app bundles exactly the way steps resolve against the examples, so a recipe naming an app this install lacks is dropped rather than half-offered; on the Examples page the card's primary action installs the app (which creates the workflows it binds), and the open chain still opens any step as a graph and still offers "Add all N as workflows".

## Merge with main

Main's example-app listing (#5646) moved `ExampleAppSummary` into the protocol and gave it a `thumbnailUrl`, colliding with this branch's `summarize()` helper in `example-apps.ts`. The helper stayed and now fills the thumbnail, so the recipe's app lookup and the app listing build one summary. Two follow-ons from the same collision: `exampleRecipeApp` extends the shared `exampleAppSummary` instead of restating its fields, and the install client this branch had added to `applicationBundle.ts` is dropped for the zod-validated one main landed in `utils/exampleApps.ts` — so a recipe install now invalidates the applications and workflows queries and reports a failure the way the example-app strip does.

## Verification

Run on the merged head (`eca212d7`) unless noted:

- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 12/12 selfchecks passed`
- [x] `npm run validate:examples` — `288/288 graphs validate cleanly`
- [x] `npm run typecheck` — web, electron and mobile clean
- [x] `npm run lint`
- [x] The 10 web suites around recipes, apps and the bundle client — 75 tests
- [x] `npx tsx marketing/scripts/generate-recipes.mjs --check` — `4 recipes up to date`
- [x] `node scripts/build-example-apps.mjs` (pre-merge head) — `validated 25 bundle(s)`, each through `nodetool app debug --no-run`. The four new bundles carry the same class of warnings the existing 21 do (no `disabledWhen` on run buttons, no error Alert), verified against `photo-studio` as the baseline.
- [x] `npm run test:affected` (pre-merge head) — 94/107 tasks pass. The one failing package is `@nodetool-ai/image-nodes` (52 GPU shader tests), which this diff does not touch: `vkCreateInstance failed with VK_ERROR_INCOMPATIBLE_DRIVER` / "No WebGPU adapter available (Node/Dawn)", the missing Vulkan ICD documented in AGENTS.md as an environment gap, not a test break. The selection widened to every workspace because two changed files (`scripts/example-apps/apps.mjs`, `scripts/validate-examples.mjs`) belong to no workspace.

## New checks

Two checks pin "the recipe's own app binds the whole chain". Both were inverted once — swapping `apps[0]` and `apps[1]` in `viral-video-ad-engine.recipe.json` so the companion app leads — and observed failing:

```
 × hands every recipe an app that binds the whole chain
-   "missing": [],
+   "missing": [
      Tests  1 failed | 6 passed (7)
```

```
  FAIL  packages/base-nodes/nodetool/examples/recipes/viral-video-ad-engine.recipe.json
        app "brand-and-social" is this recipe's app but does not bind Ad Copy in Three Registers,
        Ad Loop from a Product Photo, Cut a Landscape Clip for Vertical
```

- [x] The check was inverted once and observed failing; the failing output is above
- [x] The resolver's own drop is pinned the same way: `listExampleRecipes` with an apps directory that holds nothing returns `[]`, and the same call with the real directory returns every shipped slug — so a resolution that never looked at the apps could not pass.

## Agent capabilities

## Not in scope

The marketing `/recipes` pages and their downloadable `.nodetool` bundles still describe the chain as workflows only. `packWorkflowsBundle` packs workflows, not app bundles, so surfacing the app there is a bundle-format change worth its own PR. `generate-recipes --check` passes unchanged.

https://claude.ai/code/session_01UbfZaxSuSwyTsrqvJudBQ5


georgi:
`Quality Gate (npm run check) / test-packages-core` failed on `eca212d7`, in `packages/runtime/tests/host-binaries.test.ts`:

```
FAIL  tests/host-binaries.test.ts > runHostBinary abort signal
      > does not settle until a SIGTERM-ignoring child is actually gone
AssertionError: expected 346 to be greater than or equal to 4500
```

This PR touches no file under `packages/runtime`, but the test has a real race worth fixing rather than re-running, so I pushed the fix instead of standing down.

**The race.** The test spawns a child that ignores SIGTERM, aborts after a fixed 300 ms, and asserts the promise settles no sooner than 4500 ms — the SIGKILL that lands five seconds after the signal. The child only installs `process.on("SIGTERM")` once node has booted, which measures ~35 ms idle on this machine. When that boot outruns the 300 ms timer the signal reaches a child with no handler, the default action kills it immediately, and the promise settles at once — 346 ms in CI, i.e. 46 ms after the abort.

**Reproduced.** Running the unmodified test under 24 busy loops on 4 CPUs gives the same failure shape:

```
--- OLD (fixed 300ms) under load ---
AssertionError: expected 321 to be greater than or equal to 4500
      Tests  1 failed | 46 skipped (47)
--- NEW (wait for pid.txt) under load ---
      Tests  1 passed | 46 skipped (47)
```

**The fix** (`bdef3494`) aborts when the child's pid file appears rather than after a fixed wait. The child writes it immediately after registering the handler, so the file is proof the handler is installed; the wait is bounded, so a child that never starts fails the assertion instead of hanging. The full file passes (47 tests), and the assertion it was protecting — that the promise does not settle while the child is still alive — is unchanged.

---

PR #5647, opened 2026-09-06
docs(timeline): design and task plan for 3D clips on the timeline

georgi:
## What changed

Two documents under `docs/plans/`, no code. `3d-timeline-clips.md` is the technical design for putting a glTF scene on the timeline: a `model3d` clip on video and overlay tracks rather than a new track type, the model as the clip's asset with a `model3dStyle` block for camera, animation, lighting and background, the glTF animation following the clip's source time, four additive camera channels plus an `orbit` preset in the existing animation system, one three.js render session in `packages/video-nodes` shared by the browser preview, the browser export and the headless agent preview, and a Blender bake stored on the clip for final quality. `3d-timeline-clips-tasks.md` cuts it into twelve agent-executable tasks across three milestones with file pointers, dependency graph and acceptance tests, following the format of `motion-design-tasks.md`.

## Verification

- [x] `npm run test:affected -- --dry-run` reports `nothing — no workspace owns the changed files` (docs only)
- [ ] `npm run typecheck` (no code changed)
- [ ] `npm run lint` (no code changed)
- [ ] `npm run dev:nodetool -- harness gate --base main` (docs-only diff maps to no harness)

## Agent capabilities

Skipped: no capability is added or changed by this diff. The design names the capabilities a later task adds.

## New checks

Skipped: no check, rule or audit is added.

https://claude.ai/code/session_016T3mw6LABZatUNrN6E4Wxi

---
_Generated by [Claude Code](https://claude.ai/code/session_016T3mw6LABZatUNrN6E4Wxi)_

### DIFF
--- a/docs/plans/3d-timeline-clips-tasks.md
+++ b/docs/plans/3d-timeline-clips-tasks.md
@@ -27,7 +27,7 @@ M1: T1 (schema) ──► T2 (scene model + channels)   T3 (render session) is i
     T2 + T3 ──► T5 (agent frames on the server)
     T1 + T2 ──► T6 (ops, tools, capabilities)
 M2: T4 ──► T7 (add clip, drag-and-drop)   T4 + T6 ──► T8 (inspector)   T4 ──► T9 (lane thumbnails, clip frames)   T8 ──► T10 (preview orbit)
-M3: T6 + T8 ──► T11 (bake)   every task ──► T12 (docs, harness registry)
+M3: T6 + T8 ──► T11 (opaque bake)   T11 ──► T13 (alpha bake)   every task ──► T12 (docs, harness registry)
 ```
 
 ---
@@ -92,7 +92,8 @@ camera channels, and offers an `orbit` preset.
    `sourceTimeSec` (from `clipSourceTimeSec` times `animation.speed`).
 2. In `computeActiveLayersWithHorizon`, a `model3d` clip with a fresh bake
    (hash matches, T1's function) emits a `video` layer with
-   `assetId = bake.assetId`; otherwise a `model3d` layer with
+   `assetId = bake.assetId` and `bakeSourceTimeSec = (timeMs - clip.startMs) / 1000`
+   (design §D6, time origin); otherwise a `model3d` layer with
    `assetId = effectiveAssetId(clip)`. Count live `model3d` layers against
    `MAX_MODEL3D_LAYERS = 2`; over it, push a dropped layer with the new
    reason `model3d_layer_cap`.
@@ -110,8 +111,10 @@ camera channels, and offers an `orbit` preset.
 
 - `packages/timeline/tests/sceneModel*.test.ts`: a `model3d` clip yields a
   `model3d` layer with the right `sourceTimeSec` after a trim and a 2x speed;
-  a fresh bake yields a `video` layer; a stale bake yields `model3d`; a third
-  active 3D clip is dropped with `model3d_layer_cap`.
+  a fresh bake yields a `video` layer whose `bakeSourceTimeSec` is 0 at the
+  clip's first frame for a clip with a 5 s in point and a 2x speed; a stale
+  bake yields `model3d`; a third active 3D clip is dropped with
+  `model3d_layer_cap`.
 - Animation tests: `orbit` over a 4 s clip samples `cameraAzimuth` 180 at 2 s;
   a keyframed `cameraZoom` folds multiplicatively with the style's zoom in
   `resolveModel3DCamera`.
@@ -131,11 +134,14 @@ and `orbitOffset`).
 **Steps:**
 
 1. Split `renderGlbToPng` into `createModel3DRenderSession(glb, options)`
-   (load, meshopt, bounding sphere, lights, background, renderer, mixer) and
-   `session.render(frame)` (camera from `ClipModel3DCamera` through the
+   (load, meshopt, bounding sphere, lights, background, renderer, mixer with
+   the actions `options.animation` selects: the named clip, or every clip)
+   and `session.render(frame)` (camera from `ClipModel3DCamera` through the
    existing `computeFraming` and `orbitOffset`, `scene` mode picks the named
-   glTF camera, `mixer.setTime(frame.timeSec)`, render, return the canvas).
-   `session.animations` and `session.cameras` list the glTF's names.
+   glTF camera, `mixer.setTime(t)` with `t` wrapped at the selected actions'
+   longest duration when `loop` is on and clamped to it when off, render,
+   return the canvas). `session.animations` and `session.cameras` list the
+   glTF's names; an unknown `clipName` throws naming the available ones.
    `renderGlbToPng` becomes create, render once, `convertToBlob`, dispose.
 2. `render3d-page.ts`: `__nodetoolRenderGlbFrames(glbBase64, options, frames[])`
    returns one PNG per frame from one session.
@@ -148,9 +154,14 @@ and `orbitOffset`).
 
 - Existing `RenderToImage` tests pass unchanged.
 - New unit test on the pure part: `scene` mode with an unknown camera name
-  throws a message naming the available cameras; `frames` with two times on
-  an animated fixture yield two different PNGs (the existing headless test
-  harness, skipped without Chrome the way the current one is).
+  throws a message naming the available cameras; the loop and clamp time
+  mapping is a pure function (`mixerTimeFor(t, duration, loop)`) tested at
+  `t` past the end for both modes.
+- Headless tests on a fixture with two named animations (the existing
+  harness, skipped without Chrome the way the current one is): selecting
+  each name yields a different PNG at the same time; loop off at 1.5x the
+  duration equals the last frame; rendering `t = 0.2` after `t = 5` equals
+  rendering `t = 0.2` first.
 
 ### T4 — Browser layer source, preview, export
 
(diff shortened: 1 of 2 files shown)


georgi:
Reviewed `de878ac` as a design/task-plan PR, against the existing renderer, timeline export and Blender implementation. No runtime code changes or tests were run. I would resolve these contract gaps before using the tasks as an implementation specification.

1. **F1 [P1] Transparent bakes cannot use the existing RenderAnimation output.** [D6](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/docs/plans/3d-timeline-clips.md#L174-L183) / T11 pass `transparent` and automatically substitute the result for the live layer. However, [render_animation.py](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/packages/blender-nodes/blender_ops/ops/render_animation.py#L129-L141) requires `.mp4` and hardcodes MPEG4/H264, documented as `yuv420p`. This cannot retain the live layer's alpha. Baking a model over footage would replace transparency with an opaque background. Add an alpha-preserving output path compatible with the browser decoder, or explicitly refuse transparent bakes until supported. Acceptance should composite a baked transparent model over a colored background and verify that the background remains visible.

2. **F2 [P1] Define the baked video's time origin and include timing in its dependency hash.** [D6](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/docs/plans/3d-timeline-clips.md#L176-L183) / T2 / T11 render a clip range, hash style + asset + camera curves, then substitute the video asset. The existing [export resolver](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/web/src/components/timeline/render/TimelineRenderer.ts#L479-L489) seeks video using `clipSourceTimeSec(layer.clip, timeMs)`, including the original in-point, speed and remap. If a trimmed clip starting at source 5 s is baked into a clip-local video starting at 0 s, substitution seeks that video to 5 s at the clip's first frame. Define whether bakes represent source time or evaluated clip-local time, and specify the corresponding seek mapping. Include all bake-affecting timing and sequence settings in the hash, excluding the bake record itself. Test a nonzero in-point, speed/remap, and a duration change after baking.

3. **F3 [P2] Pass animation selection and loop policy into the session contract.** [D5's API](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/docs/plans/3d-timeline-clips.md#L141-L160) accepts only lighting/background at creation and camera/time/dimensions per frame. Neither entry point accepts `animation.clipName` or `animation.loop`, although D2/D3 and T8 expose both. T3's `mixer.setTime` cannot infer them from a timestamp. Add the animation configuration and define how edits reset/reconfigure actions. Test two named animations, loop-off beyond the end, and backward scrubbing after reaching the end.

4. **F4 [P2] Group headless sessions by render configuration, not just asset ID.** [T5 step 1](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/docs/plans/3d-timeline-clips-tasks.md#L207-L211) renders once per asset, but D5 fixes lighting, intensity and background when creating a session. Two clips sharing one GLB with different backgrounds or lighting will consequently use one clip's settings in the agent preview. The decoded-image cache's style hash cannot correct pixels already rendered with the wrong options. Group by asset plus session configuration, or allow those settings per frame. Also specify browser session invalidation when these style fields change. Test the same asset in two clips with different backgrounds/intensities.

5. **F5 [P2] The existing Blender job does not reproduce orbit on animated models.** [D6](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/docs/plans/3d-timeline-clips.md#L185-L189) says the orbit preset is reproduced exactly and T11 rejects only custom camera keyframes. In [render_animation.py](https://github.com/nodetool-ai/nodetool/blob/de878acd93bac132839d04c07a26d726ffeb2029/packages/blender-nodes/blender_ops/ops/render_animation.py#L112-L120), `_animate_orbit` runs only when `not animated and mode == "orbit"`. An animated GLB with the orbit preset therefore loses its camera motion when baked. Add a Blender prerequisite allowing model animation and orbit together, or reject that combination explicitly until supported. Acceptance needs an animated fixture with a simultaneous orbit and camera checks at multiple times.

georgi:
All five confirmed against the code and addressed in a22b1020.

- **F1 (transparent bakes)**: D6 now splits the bake into an opaque path (the existing MP4/H.264 op, unchanged) and a transparent path that renders an RGBA PNG sequence with `film_transparent` and encodes it to WebM VP9 `yuva420p` through the ffmpeg helper, the same alpha format `outputFormats.ts` already declares. The op refuses a transparent style until that path ships (T13), and a browser that cannot decode the alpha bake falls back to the live layer with a `bake_undecodable` degradation. T13's acceptance composites a baked transparent model over a solid color and asserts the color is visible.
- **F2 (time origin, hash)**: bakes are clip-local evaluated time. The scene model emits the baked video layer with `bakeSourceTimeSec = (timeMs - clip.startMs) / 1000` and both resolvers seek by it instead of `clipSourceTimeSec`. The hash covers style minus `bake`, asset id, animations, in/out point, duration, speed, time remap, and the sequence's fps, width and height. T2 and T11 test a 5 s in point with 2x speed, and a duration and fps change after baking.
- **F3 (animation in the session contract)**: `Model3DSessionOptions` now carries `animation` (`clipName`, `loop`, `speed`) at creation, the mixer wraps or clamps at the selected actions' duration, a changed option disposes and recreates the session. T3 tests two named animations, loop off past the end, and a backward scrub after the end.
- **F4 (grouping)**: browser pool entries and headless groups are keyed by asset id plus `Model3DSessionOptions`. T4 tests that a background change recreates the session and a camera change does not; T5 tests two clips on one asset with different backgrounds making two calls.
- **F5 (orbit on animated models)**: T11 starts with a Blender prerequisite that keyframes the orbit camera whenever `camera_mode == "orbit"`, with the model's animation playing underneath, and tests an animated fixture with a 180° sweep at first, middle and last frame.

---

georgi:
Re-reviewed `a22b102`. F1, F3, F4 and F5 are addressed at the design/task-plan level. F2 is partially addressed: the clip-local playback field and expanded hash resolve the consumer-side contract, but two gaps remain before this is ready to implement.

1. **F2 [P1, still open] The bake producer cannot yet generate evaluated clip-local time.** [T11 step 4](https://github.com/nodetool-ai/nodetool/blob/a22b10201f4d9cbdabb3a919b9970dd9c4369107/docs/plans/3d-timeline-clips-tasks.md#L360-L366) still builds a `RenderAnimation` request from camera terms, fps and a frame range derived from duration. The existing [RenderAnimationParams](https://github.com/nodetool-ai/nodetool/blob/a22b10201f4d9cbdabb3a919b9970dd9c4369107/packages/blender-nodes/src/job.ts#L65-L74) only supports consecutive source frames at the output fps. It has no per-output-frame source-time mapping, animation selection or loop policy. For a clip with a 5 s in-point and 2x speed, output frames need model times `5 + 2*i/fps`; changing the output seek to zero does not make Blender generate those poses. A reversed/nonlinear remap cannot be expressed by a start/end range at all. Add an explicit Blender prerequisite for sampling the evaluated model time independently of output frame number (including animation selection/looping), or refuse unsupported timing/configurations until it exists. Test actual model poses before/after baking at several times, rather than only mocked video seeks and hash invalidation. Also fix D6's unit expression to `clip.startMs + 1000*i/fps`.

2. **F6 [P2, new] Apply bake-local seeking to the server frame decoder too.** [T11 step 3](https://github.com/nodetool-ai/nodetool/blob/a22b10201f4d9cbdabb3a919b9970dd9c4369107/docs/plans/3d-timeline-clips-tasks.md#L343-L359) names the two browser resolvers, but [the agent video branch](https://github.com/nodetool-ai/nodetool/blob/a22b10201f4d9cbdabb3a919b9970dd9c4369107/packages/agents/src/timeline-preview/frames.ts#L504-L510) calls `decodeVideoFrameAt(bytes, layer.clip, timeMs)`, whose [implementation](https://github.com/nodetool-ai/nodetool/blob/a22b10201f4d9cbdabb3a919b9970dd9c4369107/packages/agents/src/timeline-preview/frames.ts#L195-L202) recomputes `clipSourceTimeSec`. A fresh bake is emitted as `kind: "video"`, so T5's new `model3d` branch will not handle it. The browser would seek a trimmed bake to 0 s while `preview_timeline_frame` seeks the same bake to 5 s, returning the wrong frame or no frame. Add this decoder to T11 and pass the resolved bake-local time into it. Acceptance should run a fresh trimmed bake through the agent frame path and assert its first requested source timestamp is zero.

Validation: inspected both updated documents and the existing Blender job and agent decoder at this head. Docs-only review, no runtime tests run.

georgi:
Both addressed in b588e8a5.

- **F2 (producer)**: the bake no longer asks Blender for a consecutive frame range. `render_animation` gains `frame_times` (one model time per output frame), `animation_name` and a camera per frame; the op selects that action, calls `frame_set` with the subframe per entry, places the orbit camera per still, and renders a PNG each, which the TS side muxes. The timeline computes the list with `clipSourceTimeSec` at each output frame, `animation.speed`, and loop wrap or clamp against the animation duration read from the glTF sampler accessors, so a reversed remap is a decreasing list and no camera configuration is refused. Unit fixed to `clip.startMs + 1000 * i / fps`. T11 acceptance now compares real Blender stills: a fixture baked with `frame_times: [0, 1.0, 0.5]` against `render_image` at 1.0 and 0.5, plus the samples-list tests for the 5 s in point at 2x, reversal and loop. The still-per-frame loop is also what T13's transparent bake uses, so both formats share one op path.
- **F6 (agent decoder)**: `decodeVideoFrameAt` takes the resolved seek time instead of recomputing `clipSourceTimeSec`, and the agent `video` branch passes `bakeSourceTimeSec` when present. Acceptance runs a fresh trimmed bake through the agent frame path with `forEachVideoFrame` mocked and asserts the first requested source timestamp is zero.

---

georgi:
Re-reviewed `b588e8a`. F2 and F6 are now addressed at the design/task-plan level. The sampled Blender producer separates model time from output frame number, and T11 explicitly routes bake-local seek times through the browser preview, browser export and agent decoder. The acceptance criteria now cover retimed/reversed samples, rendered poses and the agent's first seek timestamp. No new blocking findings in this update.

Two implementation clarifications to carry into T11:

- Preserve D2/D5's default of playing all animations when `animation_name` is absent. The new producer text describes selecting one action and muting the rest, so its tests should also cover the unnamed/default selection and models with no animations.
- D6 sends a resolved `ClipModel3DCamera`, while T11 names the existing Blender `CameraParams[]`. Define the mapping or extend that DTO to preserve `sceneCameraName` and `targetOffset`, which the existing camera params do not carry. Include a named scene camera in the bake tests.

The design is ready to proceed with those details tracked in implementation. This is a docs-only review, not verification of the future renderer. No runtime tests run.

georgi:
Both carried into T11 in 08ecf313: `animation_name` absent leaves every action playing (the D2 default), a model with no animations renders its rest pose, and `BakeCameraParams` extends `CameraParams` with `scene_camera_name` and `target_offset` with a tested mapping from `ClipModel3DCamera`. The Blender op test now covers the unnamed selection, a no-animation fixture, and a named scene camera.

---

PR #5653, opened 2026-09-06
docs(graph-resources): design and tasks for typed storyboard, entity, script and timeline nodes

georgi:
## What changed

Adds `docs/graph-resources/design.md` and `tasks.md`. Storyboards and entities cannot enter a workflow graph today (no `StoryboardRef`, no `entity` value type, no `nodetool.storyboard.*` or `nodetool.entity.*` nodes), and the creative nodes pass screenplays as untyped dicts that never touch a persisted board or the entity library. So every variant of a piece a director approved is re-directed by chat. The design makes the four documents typed graph values, adds derive nodes (`RecastStoryboard`, `FillScript`, `WriteScript`, `RetargetTimeline`, `FillTimelineText`, `LoadGameTemplate`/`SlotPrompt`/`ExportGodotProject`) that create new rows and never write the template, puts the mappings in pure functions shared by nodes and the existing agent capabilities, and walks four pipelines end to end: per-SKU product ads, localized explainers, a Godot asset re-skin, and one cut in three aspect ratios. Docs only, no code.

## Verification

Docs-only diff. Forbidden-word grep against `docs/WRITING_STYLE.md` returned no matches on both files.

- [x] `npm run test:affected` (no code changed, nothing selected)
- [x] `npm run typecheck` (no code changed)
- [x] `npm run lint` (no code changed)
- [x] `npm run dev:nodetool -- harness gate --base main` (docs map to no harness)

## Agent capabilities

Skipped: no capability added or changed.

## New checks

Skipped: no check added.

https://claude.ai/code/session_018fCmdFQqnSFSzNvCWVfQZi

---
_Generated by [Claude Code](https://claude.ai/code/session_018fCmdFQqnSFSzNvCWVfQZi)_

### DIFF
--- a/docs/graph-resources/design.md
+++ b/docs/graph-resources/design.md
@@ -56,6 +56,12 @@ export interface StoryboardRef {
   id?: string | null;
   /** Optional inline StoryboardDocument, for tests and the debug harness. */
   data?: unknown;
+  /**
+   * Set only by a node that created or derived the row in this run. A ref
+   * from the picker or from LoadStoryboard never carries it (§4.2, write
+   * contract).
+   */
+  writable?: boolean;
 }
 ```
 
@@ -76,9 +82,11 @@ Storyboard document lineage (`packages/models/src/storyboard.ts`,
 ```ts
 StoryboardDocument
 ├─ templateId?: string | null        // board this one was recast from
-└─ recastKey?: string | null         // the substitution identity (sorted
-                                     // entity ids joined "+"), so a re-run
-                                     // finds its previous copy
+├─ recastKey?: string | null         // the canonical substitution mapping
+│                                    // (§3.1), so a re-run finds its copy
+└─ templateFingerprint?: string | null
+                                     // hash of the template content the copy
+                                     // was last derived from (§3.1, reuse)
 ```
 
 Script document lineage (`packages/models/src/script.ts`):
@@ -163,22 +171,30 @@ render plan are storyboard-only and need both, and both `video-nodes` and
 ```ts
 // recast.ts
 export interface RecastInput {
-  document: StoryboardDocument;
+  document: StoryboardDocument;      // the template, as it is now
   /** Entities on the source board, resolved. */
   boardEntities: Entity[];
   /** Incoming cast. Each replaces the board entity it targets. */
   cast: Array<{ entity: Entity; replaces?: string /* board entity id or name */ }>;
+  /** The copy a previous run made for this mapping, when reusing (below). */
+  existing?: StoryboardDocument;
 }
 export interface RecastResult {
-  document: StoryboardDocument;      // new board document, lineage stamped
+  document: StoryboardDocument;      // new or re-derived copy, lineage stamped
   substitutions: Array<{ from: Entity; to: Entity }>;
   /** Shots whose rendered prompt changed and lost their takes. */
   invalidatedShotIds: string[];
   /** Shots whose prompt is unchanged and kept keyframe/clip versions. */
   keptShotIds: string[];
+  /** Shots the template no longer has; their takes on `existing` are gone. */
+  droppedShotIds: string[];
   recastKey: string;
+  templateFingerprint: string;
 }
 export function recastStoryboard(input: RecastInput): RecastResult;
+export function templateFingerprint(
+  doc: StoryboardDocument, entities: Entity[]
+): string;
 ```
 
 Rules, each pinned by a test:
@@ -194,8 +210,30 @@ Rules, each pinned by a test:
   `injectEntities`, hashes them the way `RenderInputs.prompt_hash` is written
   today, and clears `keyframe*`/`clip*`/`status` only when the hash moved. A
   product-only frame on a board whose model changed keeps its takes.
-- `templateId` = source board id, `recastKey` = sorted resulting entity ids
-  joined with `+`, `timeline_id` not copied.
+- `templateId` = source board id. `recastKey` is the **mapping**, not the
+  result: one token per resolved substitution, `<sourceEntityId>><destEntityId>`,
+  one per append, `+<destEntityId>`, sorted and joined with `,`. Input order
+  does not matter; role assignment does. On a board with characters A and B,
+  A→X, B→Y and A→Y, B→X are different keys and different copies.
+- Copies keep the template's shot ids, so a later re-derive merges by id.
+  `timeline_id` is not copied; the approved cut is carried by
+  `AssembleTimeline` (§4.2).
+- `templateFingerprint` hashes the fields a derivation reads and nothing
+  else: each shot's text fields, `camera`, `entity_ids`, `render_mode`,
+  `duration_seconds`, its scene's `lighting`; the board's `style`,
+  `aspectRatio`, `imageModel`, `videoModel`; each board entity's `descriptor`.
+  Takes and versions are not in it.
+
+**Reuse re-derives, it does not return.** With `existing` set, the function
+still derives from the *current* template, then merges: for every shot id in
+both documents the existing copy's `keyframe*`, `clip*`, `status` and
+`render_inputs` are carried over, and the hash rule above decides whether they
+survive. Shots new on the template appear; shots the template dropped
+disappear and are named in `droppedShotIds`. A template edit to one shot's
+action therefore invalidates that shot on every reused copy and no other, and
+the next `only_stale` render sees the new prompt against the old
+`render_inputs`. When the stored `templateFingerprint` equals the current one
+the merge is skipped and `existing` comes back unchanged.
 
 ```ts
 // render-plan.ts
@@ -233,6 +271,21 @@ export function fillTimelineText(
 ): { sequence: TimelineSequence; filled: string[]; unresolved: string[] };
 // `{{key}}` in ClipTextStyle.text and caption text. Unresolved keys are
 // reported, never blanked.
+
+// clone.ts
+export function cloneTimelineForBoard(
+  seq: TimelineSequence,
+  from: { boardId: string },
+  to: { boardId: string }
+): TimelineSequence;
+// A new sequence (fresh id, `templateId` = the source sequence id). Every
+// clip owned by `from.boardId` (`storyboardBoardId` matches) is re-stamped
+// to `to.boardId` with its media cleared; every other clip and every track
+// is copied verbatim. The next re-assemble fills the owned clips from the
+// copy's renders and, through `foreignTimelineParts` (`reassemble.ts`),
+// leaves the text overlay, the music bed and any other edit alone. Hand
+// edits on the shot clips themselves (a trim, a moved cut) are rebuilt from
+// the shots, exactly as today's re-assemble does.
 
 // retarget.ts
 export function retargetSequence(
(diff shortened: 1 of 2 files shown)


georgi (commented):
Reviewed both documents at 87c8eda against the current storyboard rendering and timeline assembly code. Four design issues should be resolved before implementation: substitution-key collisions, reuse after template edits, loss of the approved cut in E1, and a template-write guard that permits the first accidental write. Inline comments include concrete cases and suggested checks. This is a docs-only review; no application tests were run.

docs/graph-resources/design.md
```diff
@@ -0,0 +1,537 @@
...
+  several or none, the entry is appended and nothing is renamed.
+- Substitution rewrites the old name to the new name, whole word, case
+  insensitive, in `action`, `motion`, `dialogue`, `narration`, `slug`, and in
+  each shot's explicit `entity_ids`, and on `entityIds`/`screenplay.entity_ids`.
+- Invalidation is by prompt, not by touch: for each shot the function
+  recomputes `keyframePrompt` and `clipPrompt` with the new cast through
+  `injectEntities`, hashes them the way `RenderInputs.prompt_hash` is written
+  today, and clears `keyframe*`/`clip*`/`status` only when the hash moved. A
+  product-only frame on a board whose model changed keeps its takes.
+- `templateId` = source board id, `recastKey` = sorted resulting entity ids
```
georgi: F1 [P2] Include the replacement mapping in recastKey. Sorting only the resulting entity IDs makes two different recasts identical: on a board with characters A and B, A→X/B→Y and A→Y/B→X both yield `X+Y`, despite producing different shot text. With `reuse_existing: true`, the second graph receives the first graph's board and renders the wrong casting. Canonicalize the resolved source-ID→destination-ID pairs (and append operations), so input ordering is irrelevant but role assignment is preserved. Add a fixture for the swapped assignment.

docs/graph-resources/design.md
```diff
@@ -0,0 +1,537 @@
...
+| `RecastStoryboard` | `storyboard`, `cast: list[entity]`, `replaces?: list[str]`, `name?: str`, `reuse_existing: bool = true` | `storyboard` (the copy), `invalidated: list[str]`, `kept: list[str]` |
+| `RenderStills` | `storyboard`, `targets?: list[str]`, `max_shots: int = 24`, `concurrency: int = 2`, `only_stale: bool = true` | `storyboard`, `keyframes: list[image]`, `rendered: list[str]`, `skipped: list[str]`, `failed: list[str]` |
+| `RenderClips` | `storyboard`, `targets?`, `max_shots: int = 8`, `require_keyframe: bool = true`, `concurrency: int = 1`, `only_stale: bool = true` | `storyboard`, `clips: list[video]`, `rendered`, `skipped`, `failed` |
+| `AssembleTimeline` | `storyboard`, `name?: str` | `timeline: timeline`, `skipped_shots: list[str]`, `retimed: list[dict]` |
+
+`RecastStoryboard` with `reuse_existing` looks up a board with the same
+`templateId` and `recastKey` in the same project and returns it instead of a
+new one; a re-run then re-renders only what `only_stale` finds. Render nodes
```
georgi: F2 [P2] Define how reused copies receive template changes. After the first run, editing a shot's action, style, or model on the source board leaves `templateId` and `recastKey` unchanged. This rule returns the old copy without applying those edits. `only_stale` then compares the old copy's inputs with its old renders, so it cannot detect the changed source. Either version the derivation using a source-content fingerprint or refresh the reused document through an explicit merge/invalidation rule. Add a rerun fixture that edits source direction while keeping the cast IDs unchanged and verifies the affected shot changes.

docs/graph-resources/design.md
```diff
@@ -0,0 +1,537 @@
...
+ForEachRow(dataframe)                       → row {sku, name, image, descriptor, price}
+LoadImageFile(row.image)                    → image
+CreateEntity(image, kind: prop, name: row.name,
+             descriptor: row.descriptor, key: row.sku)   → entity
+Constant.Storyboard("Hero 9:16")            → template
+RecastStoryboard(template, cast: [entity], replaces: ["Product"])
+                                            → board'  (invalidated: shots naming Product)
+RenderStills(board')                        → board'  (only stale shots render)
+RenderClips(board', require_keyframe: true) → board'
+AssembleTimeline(board')                    → timeline'
```
georgi: F3 [P2] Carry the approved timeline into the per-SKU derivation. E1 puts `{{name}}` and `{{price}}` in the template cut, but the graph only loads the storyboard, and §3.1 explicitly drops `timeline_id`. The existing assembly path creates a fresh sequence from `buildStoryboardTimeline` when no timeline is linked; that builder receives shots/narration/music, not the template cut or its text clips. Therefore the timeline passed to `FillTimelineText` contains neither placeholder, and the resulting ad loses the approved overlay (as well as other cut-only edits). Specify a timeline-copy/media-replacement step that preserves the template's text and edit structure, and make the E1 fixture assert the filled overlay survives export.

docs/graph-resources/design.md
```diff
@@ -0,0 +1,537 @@
...
+| `RecastStoryboard` | `storyboard`, `cast: list[entity]`, `replaces?: list[str]`, `name?: str`, `reuse_existing: bool = true` | `storyboard` (the copy), `invalidated: list[str]`, `kept: list[str]` |
+| `RenderStills` | `storyboard`, `targets?: list[str]`, `max_shots: int = 24`, `concurrency: int = 2`, `only_stale: bool = true` | `storyboard`, `keyframes: list[image]`, `rendered: list[str]`, `skipped: list[str]`, `failed: list[str]` |
+| `RenderClips` | `storyboard`, `targets?`, `max_shots: int = 8`, `require_keyframe: bool = true`, `concurrency: int = 1`, `only_stale: bool = true` | `storyboard`, `clips: list[video]`, `rendered`, `skipped`, `failed` |
+| `AssembleTimeline` | `storyboard`, `name?: str` | `timeline: timeline`, `skipped_shots: list[str]`, `retimed: list[dict]` |
+
+`RecastStoryboard` with `reuse_existing` looks up a board with the same
+`templateId` and `recastKey` in the same project and returns it instead of a
+new one; a re-run then re-renders only what `only_stale` finds. Render nodes
+refuse a board whose `templateId` is null **and** which is referenced by
+`recastKey` from another board (it is a template) unless
```
georgi: F4 [P2] Protect source boards before any derived row exists. Even if this reverse lookup uses `templateId` (the actual parent pointer, whereas `recastKey` contains entity IDs), a newly approved board has no children yet. Wiring it directly to `RenderStills` therefore passes this guard and writes over the source on the first run. A recast board later used as a template also bypasses the check because its own `templateId` is non-null. Define write permission independently of whether another board already references it, for example an explicit read-only/template state or a writable derived-ref contract with an explicit override. Test the first-run miswire and a derived board reused as a template.

docs/graph-resources/design.md
georgi: Fixed. `recastKey` is now the canonical mapping: one token per resolved substitution `<sourceEntityId>><destEntityId>` and one per append `+<destEntityId>`, sorted and joined. A→X,B→Y and A→Y,B→X produce different keys and different copies. The swapped-assignment fixture is in §7 and in the recast task.

---

docs/graph-resources/design.md
georgi: Fixed. Reuse re-derives instead of returning: `recastStoryboard` takes `existing`, derives from the current template, and merges takes by shot id (copies keep the template's shot ids) under the same prompt-hash rule, so an edited action invalidates exactly that shot on every reused copy. Added `templateFingerprint` over the derivation-relevant fields as the short-circuit when nothing changed, and `droppedShotIds` for shots the template removed. The rerun fixture (edit source direction, cast ids unchanged) is in §7 and the task list.

---

docs/graph-resources/design.md
georgi: Fixed. Added `cloneTimelineForBoard` (§3.2): a new sequence with the template's clips copied verbatim, the board-owned shot clips re-stamped to the copy and cleared. `AssembleTimeline` on a copy with no cut clones the template's `timeline_id` sequence first, then re-assembles in place through `foreignTimelineParts`, so only shot clips are regenerated and the `{{name}} — {{price}}` overlay and music survive. Stated honestly that hand edits on the shot clips themselves are rebuilt, as today's re-assemble does. E1 and the fake-mode fixture now assert the filled overlay is on the exported sequence.

---

docs/graph-resources/design.md
georgi: Fixed with the writable-derived-ref contract. `StoryboardRef` gains `writable?: boolean`, set only by a node that created or derived the row in this run (`RecastStoryboard`, and the render nodes on the ref they pass on). The picker, `LoadStoryboard` and `StoryboardShots` never set it. Render and assemble nodes refuse a ref without it, naming the `allow_writes` override. Permission no longer depends on children existing, so the first-run miswire fails before spend and a derived board picked from the library is protected the same way. Both cases are in the §7 node tests.

---

docs/graph-resources/design.md
```diff
@@ -0,0 +1,617 @@
...
+**Reuse re-derives, it does not return.** With `existing` set, the function
+still derives from the *current* template, then merges: for every shot id in
+both documents the existing copy's `keyframe*`, `clip*`, `status` and
+`render_inputs` are carried over, and the hash rule above decides whether they
+survive. Shots new on the template appear; shots the template dropped
+disappear and are named in `droppedShotIds`. A template edit to one shot's
+action therefore invalidates that shot on every reused copy and no other, and
+the next `only_stale` render sees the new prompt against the old
+`render_inputs`. When the stored `templateFingerprint` equals the current one
+the merge is skipped and `existing` comes back unchanged.
```
georgi: F2 [P2] Include incoming-cast content in the reuse shortcut. `templateFingerprint(doc, entities)` is defined over the template and its board entities, while `recastKey` records only source/destination IDs. In E1, rename an existing SKU through `CreateEntity` while retaining its `source.key`: the destination ID, mapping key, and template fingerprint all stay unchanged. This shortcut returns the old copy before substitution can rewrite the old product name, so the new name never reaches its shots. The same concern applies to any cast content consumed during derivation. Hash all resolved derivation inputs, including the incoming cast's names and relevant fields, or always derive before deciding which takes to retain. Add a rerun fixture that changes a destination entity's name under the same ID with the template unchanged.

docs/graph-resources/design.md
```diff
@@ -0,0 +1,617 @@
...
+```
+
+A re-skin is swapping the style entity on the `SlotPrompt` node and running
+again. `Cave Pixel` seasons every slot the same way the storyboard render
+path seasons a shot, so the pack is consistent by construction.
+
+### E4. One cut, three ratios
+
+```
+Constant.Storyboard("Launch film 16:9") → AssembleTimeline → timeline
```
georgi: F5 [P2] Make the three-ratio example comply with the new write contract. `Constant.Storyboard` produces a read-only ref, and §4.2 now requires `AssembleTimeline` to reject that ref without `allow_writes`. E4 therefore stops before either retarget node runs. For the stated goal of adapting an approved cut, start from `Constant.Timeline` for that existing cut and derive both ratios directly. This also preserves the shot trims and placement that §3.2 says re-assembly rebuilds. If fresh assembly is intentional, show the explicit writable derivation or override and adjust the example's promise. Include E4 in the fake-mode execution checks.

georgi (commented):
Re-reviewed both documents at 3f292c8 and checked the revised contracts against current render-record and timeline assembly behavior. F1's mapping collision and F4's first-write case are addressed. The clone step addresses F3's missing-overlay case, with the loss of shot-level edits now explicitly documented. F2 is only partially addressed: the new fingerprint shortcut excludes incoming-cast changes. One additional inconsistency is that E4 now violates the write contract. Two P2 inline comments below. Docs only; no application tests run.

docs/graph-resources/design.md
georgi: Fixed by dropping the shortcut. `templateFingerprint` is gone from the lineage fields, `RecastResult` and the tasks. Reuse now always derives from the current template and the current cast, then keeps takes by prompt hash. A renamed destination entity under the same id has the new name rewritten into its shots and invalidates exactly those, and an edited descriptor invalidates the shots it seasons since it is in the injected prompt. The rename-under-same-id fixture is in §7 and the recast task.

---

docs/graph-resources/design.md
georgi: Fixed. E4 now starts from `Constant.Timeline` on the approved cut; `RetargetTimeline` derives new rows from the read-only ref, so the trims and placements carry over and nothing is re-assembled. E4 is included in the fake-mode debug runs, with the run asserting the source cut's clip placements are unchanged on both retargets.

---

georgi (commented):
Re-reviewed dd255d4, including the changes since 3f292c8 and their matching tasks. Both remaining findings are addressed:

- F2: the fingerprint shortcut is removed. Reuse always derives from the current template and incoming cast before retaining takes, with the same-ID rename regression case specified in both documents.
- F5: E4 now starts from the existing Constant.Timeline node and derives the two ratios without storyboard assembly. All four examples are included in the planned fake-mode runs, and the task checks E4's clip placements.

No new actionable findings in this revision. All findings from my previous reviews are addressed at the design level. This remains a docs-only proposal; the listed fixtures are implementation requirements, not tests executed by this review. No application tests run.

PR #5662, opened 2026-09-06
feat(setup): guided creation flows for all five document types

georgi:
Turns the New Project surface into five guided flows — storyboard, video, script, image and workflow — that take a creator from one sentence to a finished document in three steps. Implements [docs/creation-flows/prd.md](docs/creation-flows/prd.md).

## Shape

One shell (`SetupFlow`) plus four shared pieces (`OptionCardGrid`, `PresetTileGrid`, `PlanReview`, `AlternativesColumn`). A flow adds a config, a plan generator and a generate action — not its own wizard.

Three rules hold across all five:

- **Cheap text before spend.** Step 2 ends in a plan the creator edits as text. Nothing renders, voices or places a node before step 3's button.
- **Resume by stage.** The stage is one optional field on the produced document. Every field is optional and additive, so every existing document opens exactly as it does today.
- **Headless parity.** Each document operation a flow performs has a `ui_*` tool that does the same thing.

## Review pass

The assembled flows were reviewed for design and UX, and the second half of this branch is the remediation. The changes worth a reviewer's attention:

| Area | Before | Now |
|---|---|---|
| Async | A late model answer could advance the flow or overwrite a newer draft | A request owns the stage it started from; obsolete answers are dropped |
| Choice controls | Mutually exclusive cards used `aria-pressed`, every tile a tab stop | Radio groups with roving focus; one tab stop, arrow keys |
| Preset cards | A late-loading sample restructured the grid and moved focus | One shape per grid, reserved preview, focus preserved |
| Plan fields | Committed per keystroke — typing `900` committed 9s on the way through | Commit on blur or Enter; unmount drops an abandoned draft |
| Imports | Attribution and preserve-words lived in module state, lost on reload | On the document, visible to the server and the `ui_*` tools |
| Failures | Silently dropped | Per-variation reasons, failed voice lines with retry, build/validation/test-run outcomes on the landing |
| Composer context | Attachments and entities dropped at the door | Carried as durable `asset://` locators, never inline bytes |

Two review findings were **rejected as spec conflicts** rather than implemented: the workflow flow's build-then-run is specified (§ 11.3, label fixed in Appendix A), and the three-step stepper with review as a substep of step 2 is specified (§ 6.2). Both got the disclosure fix the creator actually needed instead — the test run is now stated before the click, and a merged stepper entry returns to the review the creator left.

## Verification

- `npm run build:packages` — 62/62
- `npm run typecheck` — clean (packages, web, electron, mobile)
- `npm run lint` — 0 errors
- Web suites for setup, timeline, script, storyboard, sketch, workflow and tools — 2957 passing across 256 suites
- `packages/protocol` — 1268 passing
- `npm run dev:nodetool -- harness gate --base origin/main` — 11/11 selfchecks

## Follow-ups, deliberately not in this PR

- The image flow should host its stepper in the project tab like the other four (§ 6.1). It is the only flow with no route back to the entry cards, and its overlay shape forces it to mark the sibling editor `inert`. The move is blocked on the flow writing into the editor's own store instance.
- A tRPC route returning a shipped example's materialized graph. The listing returns an empty graph, so a copy always lands in a new workflow id and the placeholder must be discarded.
- PRD table drift to apply in §§ 7.10, 8.6 and 11.6.
- Open spec gap: § 8.7 says generation creates one clip per beat, but with imported media a beat *is* an existing clip.

https://claude.ai/code/session_01HiTw8wmw2RmQgpxmEBMVPu


web/src/stores/timeline/TimelineStore.ts
```diff
@@ -1352,6 +1360,22 @@ export const createTimelineStore = (
...
+            const beats = state.setup?.beats;
+            if (!beats?.some((beat) => beat.id === beatId)) {
+              return {};
+            }
+            return {
+              setup: {
+                ...state.setup,
+                stage: state.setup?.stage ?? "review",
+                brief: state.setup?.brief ?? "",
+                beats: beats.filter((beat) => beat.id !== beatId)
```
georgi, line 1374: F1 [P2] Include setup-only mutations in the timeline autosave trigger

The new `removeBeat` action changes only `setup`, as does the new Voiceover switch. `useTimelineAutosave` includes setup in its payload and `sameDocument`, but its subscription's `docUnchanged` comparison still checks only sequenceId/tracks/clips/markers/transcript/scriptEnabled/tempo. It returns before scheduling either edit. Reproduced: switch voiceover off, remove a beat, let timers run, then unmount; zero update calls occur. Reopening before generation brings back the removed beat and the old voiceover choice. Track `setup` in the subscription baseline/comparison so these new controls persist without requiring an unrelated clip edit.

georgi (commented):
Reviewed c6141e6. Three actionable findings below, all involving loss of user edits. F1 and F2 were reproduced by executing the actual source with isolated store/React/network dependencies. F3 follows from the discard handler and the 750 ms debounced persistence paths. I did not run the full application test suite.

web/src/components/setup/storyboard/useStoryboardSetupFlow.ts
```diff
@@ -124,13 +151,65 @@ export const useStoryboardSetupFlow = ({
...
+
+  /**
+   * Run the Director, keeping the screenplay it replaces so the review step
+   * can put it back (F15). What the run was answering is recorded by the run
+   * itself, so the UI and the headless path cannot disagree about it.
+   */
+  const runDirector = useCallback(
+    async (requestedShots: number): Promise<boolean> => {
+      const board = useStoryboardStore.getState().getBoard(boardId);
+      keepPreviousScreenplay(boardId, board?.screenplay ?? null);
```
georgi: F2 [P2] Snapshot the edited board shots before offering screenplay restore

`board.screenplay.shots` retains the Director's original output, while review edits go through `updateShot`/`patchShot` into `board.shots` only. Saving `board.screenplay` here therefore omits the edits the creator is trying to protect. Reproduced: generate a screenplay, edit a shot's action, rewrite, then restore the previous screenplay; the action returns to the original model wording rather than the edited wording. Added/deleted/reordered shots have the same problem. Capture the current board's shots (and current editable title/other restored fields), or use an appropriate board checkpoint, before starting the rewrite.

web/src/components/projects/NewProjectSurface.tsx
```diff
@@ -621,46 +873,216 @@ const NewProjectSurface = () => {
...
-      type: SETUP_TAB_TYPE[setupTarget.kind],
-      ref: setupTarget.id,
-      mode: "edit",
-      title: setupTarget.name,
-      projectId: setupTarget.projectId
-    });
-    closeTab(tabId("project-new", PROJECT_NEW_REF));
-  }, [closeTab, openTab, setupTarget]);
+    const brief = await readSetupBrief(target);
+    await deleteSetupDocument(target);
```
georgi: F3 [P2] Capture the live brief before deleting a changed flow

`readSetupBrief` reads the server, but the storyboard/script fields write into local stores and persist after a 750 ms debounce. If the user edits the brief and confirms Change flow before that save completes, this reads the previous text and then deletes the only draft containing the new text. The confirmation explicitly promises that what they typed comes with them. The same loss occurs when autosave is pending or has failed. Read the mounted document's live brief (or await a successful flush before reading), and preserve an intentionally empty brief too instead of retaining the old composer prompt.

georgi:
@claude address comments and fix CI

georgi (commented):
Re-reviewed 69104f1f, including all changes since c6141e6. No new actionable findings in this update.

- F1 fixed: the autosave subscriber now tracks setup. Re-ran the source-level reproduction; switching voiceover off and deleting a beat produces one PATCH containing both changes.
- F2 fixed: the restore snapshot uses the live board. Re-ran the source-level reproduction; edited shot text and title survive rewrite and restore.
- F3 fixed: all four hosts pass their mounted document's brief to the discard handler, which also preserves an intentionally empty string. Verified the callback wiring and removal of the server read.

Also inspected the CLI pricing stub and harness process-result changes. Validation was limited to isolated source execution for F1/F2 and code review for the remaining changes, not the full application suite. CI for this commit is still queued/running, so its reported CI fixes are not yet confirmed by completed checks.

PR #5663, opened 2026-09-06
feat: graph resources — approve a document once, re-run it in a graph

georgi:
## What changed

Implements `docs/graph-resources/design.md`, all five phases. Storyboards, scripts, timelines and entities become typed graph values, so a document a director approved in an editor can be re-run from a workflow instead of re-directed by chat. `StoryboardRef` joins `ScriptRef`/`TimelineRef` and carries a `writable` flag; `entity`, `game_slot` and `slot_fill` become node value types; the four documents gain optional lineage fields that old rows load unchanged. The derivations are pure functions in a new `@nodetool-ai/storyboard` package and in `timeline`/`protocol` — `recastStoryboard`, `planShotRenders`, `fillScript`, `fillTimelineText`, `cloneTimelineForBoard`, `retargetSequence`, `slotPrompt` — each called by both a node and the agent capability rather than forked between them. On top sit `nodetool.entity.*`, `nodetool.storyboard.*`, two script nodes, two timeline nodes and a new `@nodetool-ai/game-nodes`, plus web pickers, four shipped examples and a `graph-resources` harness entry.

Two design principles are load-bearing in the diff. **A derive node writes a new row, never its source**: `RenderStills`, `RenderClips` and `AssembleTimeline` refuse a `StoryboardRef` without `writable: true`, checked before the board is read, so a first-run miswire (picker → `RenderStills`) fails with zero reads and zero spend. **A re-run pays only for what changed**: `CreateEntity` upserts on a caller-supplied key, `RecastStoryboard` re-derives its previous copy for the same substitution, and the render nodes skip a shot whose prompt hash has not moved.

The last commit is separate and breaking, done at the repo owner's request: **`lib.apple` is removed** — 17 AppleScript nodes, both DSL trees, both sandbox guest packs, docs and catalog. A saved workflow using a `lib.apple.*` node no longer resolves. These were the only platform-gated nodes in the repo, so every generator that reads the live registry produced different output off a Mac than on one; removing them makes `codegen:dsl:check`, `build:sandbox-dsl:check` and `build:sandbox-flow:check` pass on Linux for the first time, and lets the two darwin exemptions in the DSL and sandbox-pack tests be deleted rather than merely satisfied.

## Verification

- [x] `npm run test:affected` — 100/111 tasks. The one failure is `@nodetool-ai/image-nodes`: 52 failed, and `grep -c "No WebGPU adapter available"` over the same run is also 52, the documented missing-Vulkan-ICD condition in this container.
- [x] `npm run typecheck` — web and electron exit 0 (`npx tsc --noEmit` in each, captured directly). Mobile fails with `Cannot find module 'react-native'`; `mobile/node_modules` does not exist here and mobile is deliberately not a root workspace.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 16/16 selfchecks passed`, `graph-resources (41 file(s))` selected.

Also run: `npm run build:packages` 64/64, `npm run check:agents-docs` (40 files, each paired and reachable), `npm run backend:smoke` (bundle boots and answers `/health`), `npm run validate:examples` 292/292, `npm run fixtures:graph-resources` all clean, `npm run capabilities:check` current.

Checks inverted to prove they can fail, each restored afterwards:

- Godot template staging — removed `shmup/manifest.json`, `npm run backend:smoke` exit 1: `Registered asset directory @nodetool-ai/godot-templates/templates is missing file(s) after staging: shmup/manifest.json.`
- `listStoryboards` project filter — disabled it, `session-model-interfaces` went `1 failed | 30 passed`.
- Porcelain path unquoting — reverted the call, both new `parsePorcelainLine` cases went red.
- The fixture assertions — breaking E1's `values` output gives `FAIL E1: the overlay is still a template: "{{name}} — {{price}}"`; a bogus clip in E4's expectation gives `FAIL E4: vertical moved the cut`.

One process note: an early "typecheck passed" reading came from `npm run typecheck | tail`, whose exit code is `tail`'s — it was failing. Every exit code quoted above is captured directly.

## Agent capabilities

No capability is added and no declared contract changes — design §P5 is explicitly "no new agent tools". `render_storyboard_stills`, `render_storyboard_clips` and `filterStale` now call the shared package internally; their names, arguments and outputs are unchanged and their suites (111 tests) stay green. `npm run capabilities:check` reports the table current at 275 capabilities.

## New checks

- [x] Each new check was inverted once and observed failing — the four commands and their failing output are in **Verification** above.
- [x] The `graph-resources` harness selfcheck runs the pure suites plus three fake-mode fixture runs that assert on content (E1's filled overlay reached the exported sequence; E4's retargets left the source cut's clip placements alone), so it cannot pass by matching nothing. The sandbox pack's registry-coverage test additionally pins its own wrapper count against the regex that finds them, and now has no platform exemption to hide behind.

## Known gaps, recorded not worked around

`docs/graph-resources/tasks.md` carries a "Phase 5 deviations" section. The examples were written against what shipped rather than widening the node surface to fit the design's diagrams: `Collect` cannot take five lanes (E3 routes through `ForEach`), `ForEachRow` into `Code` loses its scope so E1 takes a SKU's fields as inputs, and `validateGraph` requires a `*_model` prop that the render nodes deliberately leave unset to use the board's own model.

Separately, and **not fixed here**: `nodetool debug` reports `✅ ran clean` on a graph containing a node type that does not exist. `Graph.loadFromDict` defaults to `skipErrors: true` (`packages/kernel/src/graph.ts:384`), silently dropping unknown nodes and their edges, and `ExecutionSession.create` never overrides it — affecting `nodetool debug`, `workflows run`, the eval graph runner and `packages/websocket/src/headless-job-runner.ts`. `nodetool validate` catches it. The one-line flip is not safe on its own: the resolver also returns null for Python-only nodes whose worker metadata has not loaded, so a correct fix separates "no metadata but the executor may still resolve it" from "nothing can execute this" and refuses at `create()` time. Pre-existing, found while proving one of the checks above red-capable.

https://claude.ai/code/session_01Rvyu1vGSG4PN99rYJHu7uW

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rvyu1vGSG4PN99rYJHu7uW)_


georgi:
Reviewed head `8d6aacc5845c5423bbc647542f9bb2b7bed19bd8`. Three actionable findings:

1. **F1 [P1] Preserve the approved shot edit when assembling a recast board.** [storyboard.ts:1155–1163](https://github.com/nodetool-ai/nodetool/blob/8d6aacc5845c5423bbc647542f9bb2b7bed19bd8/packages/video-nodes/src/nodes/storyboard.ts#L1155-L1163)
   `templateCut()` clones the approved sequence and re-stamps its shot clips to the destination board, but `foreignTimelineParts()` immediately removes all those clips. Replacing them with `assembled.clips` rebuilds the shots in board order, discarding the approved placements, trims, transforms, and other clip edits. Reproduction through `AssembleTimelineNode.process()`: a template shot at `startMs=1500, durationMs=1000, inPointMs=2000` becomes `startMs=0, durationMs=4000, inPointMs=0`. Titles remain at their old times while the footage underneath changes. The same path also discards edits on the copy's own timeline on subsequent runs. Refill the existing shot clips by shot identity while preserving their edit fields, and explicitly handle added/deleted shots. Add an assembly test with a reordered or trimmed template clip; the current test only checks media replacement and foreign-clip retention.

2. **F2 [P1] Do not fall back to a different keyed entity by display name.** [document-model-interfaces.ts:49–61](https://github.com/nodetool-ai/nodetool/blob/8d6aacc5845c5423bbc647542f9bb2b7bed19bd8/packages/websocket/src/lib/document-model-interfaces.ts#L49-L61)
   When an explicit `source.key` has no match, `findUpsertTarget()` still searches by kind/name, including rows with a different key. Creating `SKU-RED` and then `SKU-BLUE`, both named `T-Shirt`, returns the first asset id twice with `created: false` on the second call. Its descriptor, reference image, and source key are overwritten with the blue SKU's values. Besides corrupting the first entity, this gives both products the same recast mapping and can make the second product reuse the first product's board. An explicit unmatched key should create a distinct entity, or at least never adopt a row carrying another key. Reserve name-based matching for the no-key case and cover two distinct keys with the same name.

3. **F3 [P2] Look up reusable copies beyond the latest 50 boards.** [storyboard.ts:719–725](https://github.com/nodetool-ai/nodetool/blob/8d6aacc5845c5423bbc647542f9bb2b7bed19bd8/packages/video-nodes/src/nodes/storyboard.ts#L719-L725)
   `findExisting()` searches `listBoards()`, which calls `context.listStoryboards({ projectId })` without a limit or pagination. The real adapter delegates to `Storyboard.listByProject()`, whose default is the 50 most recently updated rows. Once a matching copy falls outside that window, `reuse_existing=true` treats it as missing, creates another board, and loses access to the previous renders. A catalog batch larger than the window can keep duplicating and re-rendering its variants on each pass. The focused lookup reproduction returned `null` for a matching copy at position 51. Add a model lookup scoped by owner/project/templateId/recastKey, or exhaust pagination before deciding no copy exists.

Validation: exercised the PR's actual assembly node and timeline helpers, entity adapter, and reuse lookup in a focused Node harness with mocked persistence dependencies. The pagination mock uses the model's verified default limit of 50. These are focused reproductions, not a full repository test run.

georgi:
Re-reviewed `207f9d85f2949404ea204abff3fe1d8e782a9eb0`, including all changes since the previous review. Two findings remain open:

- **F1 [P1], still open:** [AssembleTimeline still removes the approved shot clips and replaces them with a fresh assembly](https://github.com/nodetool-ai/nodetool/blob/207f9d85f2949404ea204abff3fe1d8e782a9eb0/packages/video-nodes/src/nodes/storyboard.ts#L1155-L1163). The focused reproduction still changes `startMs=1500, durationMs=1000, inPointMs=2000` into `startMs=0, durationMs=4000, inPointMs=0`. Preserve the existing shot edits when substituting the copy's rendered media.
- **F3 [P2], still open:** [The reuse lookup still searches only the returned board list](https://github.com/nodetool-ai/nodetool/blob/207f9d85f2949404ea204abff3fe1d8e782a9eb0/packages/video-nodes/src/nodes/storyboard.ts#L719-L725), which defaults to the latest 50 rows. A matching copy at position 51 still returns `null`, causing another copy and potentially another render bill. Use a targeted lookup or exhaust pagination.

**F2 is fixed for the reported case.** The actual updated entity adapter now returns distinct ids for `SKU-RED` and `SKU-BLUE` sharing `T-Shirt`, and preserves the first entity's key. The added tests cover both distinct keyed entities and deliberate adoption of a keyless entity.

Validation: reran the focused source harness with mocked persistence dependencies. Reviewed the incremental marketing catalog and E2E exclusion changes; no additional finding in those changes. Full repository suites were not run. I would hold merge for F1 and F3.

georgi:
F1 and F3 are fixed in `d7ac0bb5`. Both were reproduced before being written and mutation-checked after.

**F1.** `refillShotClips` in `packages/timeline/src/reassemble.ts`, beside `foreignTimelineParts` — it answers the other half of the same question: what the assembling document owns and still must not rebuild. A shot clip is identified by `(storyboardShotId, mediaType)`, so a re-assemble swaps its media and leaves placement, trim and transform alone. Your case now holds: `startMs=1500, durationMs=1000, inPointMs=2000` survives with only the asset swapped. Three cases are explicit — an added shot takes a sibling's placement when the shot already has a clip (a video clip and its audio twin) and is otherwise appended after everything already placed; a deleted shot's clip goes; a shot with nothing rendered yet holds its place rather than vanishing. With no previous cut the output is the assembly itself, so a first assemble is unchanged.

One deliberate exception: a **script-linked** board still rebuilds, because there the words decide each shot's length and a re-assemble after a re-voice has to be free to re-time. Called out in the code and in the doc.

This contradicted design §3.2, which said shot-clip hand edits "are rebuilt from the shots, exactly as today's re-assemble does" — while §4.2 said a copy "inherits the approved cut, not a bare assembly". I implemented §4.2's promise and rewrote the §3.2 sentence to match rather than leaving the doc disagreeing with the code.

**F3.** `Storyboard.findRecast({userId, projectId?, templateId, recastKey})` pushes both identity fields into SQL, wired through `findRecastStoryboard` on the context and `storyboardModelInterfaces()`. Since they live inside the document JSON, the predicate is dialect-branched off `getDbType()` — `json_extract` on SQLite, `document::json ->>` on Postgres. **The Postgres branch has not executed**: there is no Postgres in this container, so only the SQLite branch is covered by tests. That is the one part of this fix I cannot claim is verified.

**Evidence.** 4 node tests red before F1, 25/25 after; mutating `refilledClip` to take the assembled placement turns 2 reassemble tests red. The F3 model test was red with `findRecast is not a function` (its window assertion passing first, so it wasn't vacuous), and swapping the lookup back to a listing scan turns the websocket test red. Locally: `build:packages` 64/64, timeline 882 passed, models 919, websocket 2875, web+electron `tsc --noEmit` exit 0, `lint` exit 0, `harness gate --base origin/main` 16/16.

**One thing I did not fix, deliberately.** `assemble_storyboard_timeline` in `packages/agents` still rebuilds shot clips through `foreignTimelineParts`, so the agent and editor path loses the same edits the node used to. It is pre-existing rather than introduced here, and fixing it widens this PR into the capability surface that design §8 says stays unchanged — but it does mean the two paths now differ. `refillShotClips` is exported and ready for it. Say the word if you would rather it land here than as a follow-up.

---

georgi:
Re-reviewed `d7ac0bb59125000ed28f73e5317432da2572fe77`. The original F1 and F3 reproductions now pass, and F2 remains fixed. One new issue in the cut merge:

**F4 [P2] Include preserved foreign clips when computing where new shots are appended.** [reassemble.ts:191–199](https://github.com/nodetool-ai/nodetool/blob/d7ac0bb59125000ed28f73e5317432da2572fe77/packages/timeline/src/reassemble.ts#L191-L199)

`cursorMs` only considers `kept` and `added`, although `foreignClips` are retained in the final cut. An approved cut can contain a manually inserted end card on the same track after the board's shot clips. Reproduction: an owned shot occupies 0–1000 ms, and a foreign end card occupies 1000–6000 ms on that video track. Adding another board shot places it at 1000 ms on the same track, directly overlapping the end card, rather than appending after the existing cut. This defeats the stated rule that new shots join without disturbing already placed content. Include the preserved foreign clips in the append boundary, and add a test with a trailing foreign clip sharing the destination track.

Validation: exercised the actual assembly node and timeline helpers in the focused source harness. The original trim now stays at `1500/1000/2000`; the scoped lookup finds its copy beyond 60 newer rows and rejects another owner. The lookup ran against native in-memory SQLite through a minimal query-builder shim, using the actual model method and adapter. The SKU regression still passes. Full repository suites and the PostgreSQL branch were not run. The documented script-linked rebuild exception and pre-existing agent assembly behavior are not additional findings here.

georgi:
Merged `main` again (`f01e7811` 3D clips, `5b9541a1` guided setup flows). One conflict, in `packages/protocol/src/api-schemas/storyboards.ts`: both sides added optional fields to `storyboardDocument`, so both are kept — `templateId`/`recastKey` from this PR alongside `importSource`/`setupShotCount`/`setupDirectedFrom` from #5662.

**One failure in that merge is not this PR's, and I've ported a fix rather than leave the leg red.** `npx tsc --noEmit` in `web` fails on `src/components/setup/storyboard/__tests__/setupChoices.test.ts` — #5662's `Shot` factory omits `status`, which `Shot` has always required. Evidence it is main's and not mine:

- `git diff origin/main -- web/src/components/setup/storyboard/__tests__/setupChoices.test.ts` → empty; the failing file is byte-identical to main's.
- `git diff origin/main -- packages/protocol/src/creative.ts | grep status` → empty; the `status: ShotStatus` member it violates is byte-identical to main's. My only change to that file is the entity `source` field.

So it reproduces on main. `fe787e69` adds `status: "planned"` to that factory — one line, in main's file, no-ops once main carries its own fix. Web typecheck is exit 0 after it, and the 8 setup-storyboard suites stay green (100 tests).

Post-merge state locally: `build:packages` 64/64, protocol 1310, models 919, timeline 930, storyboard 41, video-nodes 310, websocket 2875, `lint` exit 0, web and electron `tsc --noEmit` exit 0, `codegen:dsl:check` up to date (69 namespaces, 430 nodes).

---

georgi:
F4 fixed in `d19e962e`. You were right, and it defeated the rule the docstring claims.

`refillShotClips` measured the append boundary over `kept` and `added` only, while `foreignClips` survive into the same cut — so a hand-placed end card at 1000–6000 on the shots track got a newly added shot laid over it at 1000. The boundary is now the end of everything that survives:

```ts
let cursorMs = [...kept, ...added, ...foreignClips].reduce(...)
```

Test added with a trailing foreign clip on the destination track, as you asked: the added shot now starts at 6000, and the end card comes through untouched. Reverting that one term fails exactly this test (`expected { …(13) } to match object { startMs: 6000, …(1) }`) and nothing else — so it isn't passing by accident.

`build:packages` 64/64, timeline 931 passed / 43 skipped, storyboard-nodes 25/25, `lint` exit 0.

Still open from my previous comment, your call: `assemble_storyboard_timeline` in `packages/agents` still rebuilds shot clips through `foreignTimelineParts`, so the agent and editor path loses the edits the node now preserves — including this end-card case. Pre-existing rather than introduced here, and fixing it reaches into the capability surface design §8 says stays unchanged, so I have left it. `refillShotClips` is exported and ready if you would rather it land in this PR.

---

PR #5665, opened 2026-09-06
feat(setup): Game guided flow to a Godot project (E6)

georgi:
## What changed

A sixth entry card, **Game**, on the New Project surface. It runs on the existing `SetupFlow` shell: an Idea step with three pinned inspiration chips and a blank-template escape hatch, a Design step that picks the Godot template and shows the written design (premise, loop, cast, level, one prompt per slot) as editable text before anything is spent, and a Look step with six game style presets and one model per slot kind (image required, sound and music optional with placeholder fallbacks). `Build your game` places a slot-filling workflow graph without a model call (generate, resize to the slot's exact pixels, check, stamp), runs it once, and lands in the node editor with a checklist in the canvas chat dock: assets checked, exported directory, verified only when Godot actually ran, and next steps (open folder, download zip, play-test with the project agent, regenerate).

The graph ends in `nodetool.game.ExportGodotProject` from main's `game-nodes` package (merged in), which gains the flow's behaviours: empty fills export the template with placeholder art, unfilled slots keep theirs, a zip lands beside the directory, and `verified` stays false unless a Godot run objected to nothing. The `export_godot_project` capability shares the same join. Four `ui_game_*` tools and their server mirrors give the flow headless parity; a `game-flow` harness entry runs every suite it depends on.

Design: `docs/creation-flows/game-prd.md`. Tasks: `docs/creation-flows/tasks.md § P10`.

## Verification

- [x] `npm run test:affected` equivalent per package: protocol `-- game` 189 tests, game-nodes 45, agents `godot game-setup` 31 (4 skipped, Godot absent), base-nodes `game-graph-chips` 12 (every chip's placement validates against the real registry with zero issues), websocket `trpc-games style-presets` 15, image-nodes and audio-nodes `-- game` 22 and 12, sandbox-compiler `dsl-pack-guest` 25; full web suite 1426 suites / 15689 tests; electron 692 tests
- [x] `npm run typecheck` (web and electron clean; mobile needs `mobile/node_modules`, absent in the build container)
- [x] `npm run lint` (no errors)
- [x] `npm run dev:nodetool -- harness gate --base origin/main`: 95 files mapped to 14 surfaces, `Gate: 17/17 selfchecks passed`
- [x] `npm run check:lockfile`, `npm run check:circular`

Checks inverted and seen red, then restored: `writeGameSetup` settings passthrough; the graph builder's "no edge references an unplaced node" assertion; the chips test's `issues` assertion; the checklist's verified-only-when-`true` rule; the tile script's `--check` (one byte changed in a PNG). A CI flake in `packages/audio-nodes/tests/cv.test.ts` (pacer clock read after construction) was made robust on this branch.

## Agent capabilities

- [x] `npm run capabilities:check` passes (table synced)
- [x] `set_game_setup`, `design_game`, `update_game_design`, `build_game` name `packages/agents/tests/capabilities-game-setup.test.ts` in the capability table

## New checks

- [x] Every check above was inverted once and observed failing
- [x] `scripts/make-game-style-tiles.mjs --check` fails when it renders fewer than nine tiles, so it cannot pass by comparing nothing

https://claude.ai/code/session_01TDqhsPX3jAxwKFUvvK9qkf


web/src/hooks/game/useGameTemplates.ts
```diff
@@ -0,0 +1,67 @@
...
+  hooks: template.hooks
+});
+
+export function useGameTemplates(): UseQueryResult<GameTemplate[], Error> {
+  return useQuery({
+    queryKey: GAME_TEMPLATES_QUERY_KEY,
+    queryFn: async (): Promise<GameTemplate[]> => {
+      const client = trpcClient as unknown as GameTemplatesRouter;
+      const answer = await client.games.templates.query();
+      return answer.templates;
```
georgi: F1 [P1] Read the array returned by games.templates

`gamesRouter.templates` declares `z.array(gameTemplateSummary)` and returns `listTemplates().map(...)`, so the tRPC result is already `GameTemplate[]`. Reading `answer.templates` returns undefined. Consequently the Game flow has no template cards, both template alternatives stay disabled, and the normal Design step cannot select a template. Return the array directly and remove the incompatible client cast. The hook test currently mocks `{ templates: [...] }`, which disagrees with the real router and hides the blocker.

web/src/components/setup/game/gameRunSummary.ts
```diff
@@ -0,0 +1,194 @@
...
+        nodes,
+        resultFor: () => undefined,
+        errorFor: () => undefined
+      });
+    }
+    const key = (nodeId: string) => `${workflowId}:${jobId}:${nodeId}`;
+    return summarizeGameRun({
+      nodes,
+      resultFor: (nodeId) =>
+        outputResults[key(nodeId) as keyof typeof outputResults],
```
georgi: F2 [P1] Read named node outputs instead of the terminal stream buffer

`ResultsStore.outputResults` is populated by `output_update`, which stores individual values and appends them by node ID without retaining their handle names. For the generated export node, `output` is connected to `output_project` and therefore suppressed by the runner's terminal-output filter. Its unconnected directory/files/verified/verification/errors/archive handles accumulate into an array under the export node ID. `readExportOutput` expects one object with named fields, so a successful export still yields directory=null, archive=null and verified=false. This keeps Open project folder, Download project and Play-test disabled. Read the focused job's completed `liveGenerations[].outputs` (and persisted history for reloads), or consume the actual Output node deliberately. Add a test using the runner/reducer event shape rather than injecting an already assembled export object.

web/src/components/projects/NewProjectSurface.tsx
```diff
@@ -1130,6 +1199,15 @@ const NewProjectSurface = () => {
         />
       );
     }
+    if (setupTarget.kind === "game") {
+      return (
+        <GameSetupHost
+          workflowId={setupTarget.id}
```
georgi, line 1205: F3 [P2] Restore unfinished game setup when reopening its workflow

This is the only mount of `GameSetupHost`, and reaching it depends on `setupTarget`, which is component-local state initialized to null. After refreshing or closing/reopening during Design or Look, that state is gone. Reopening the saved workflow routes through `TabContent` to `WorkflowEditorSurface`, which never checks `settings.game.stage` and opens the empty canvas instead. The saved answers therefore cannot resume the flow despite being persisted. Route workflows with a nonterminal game stage back through the setup host, and test a real unmount/reopen rather than remounting the hook with an already supplied workflow ID.

packages/agents/src/capabilities/workflows.ts
```diff
@@ -1314,6 +1348,444 @@ async function validateBuiltGraph(
...
+      return {
+        error:
+          "No image model is chosen. Set image_model to `provider:model_id` with set_game_setup — find_model has the ids."
+      };
+    }
+    const musicTile = owned.game?.music_model ?? "";
+    const musicModel = musicTile === "" ? null : modelValue(musicTile, "music_model");
+    if (musicTile !== "" && !musicModel) {
+      return {
+        error: `music_model "${musicTile}" is not \`provider:model_id\`. Omit it to keep the template's placeholder music.`
```
georgi: F4 [P2] Normalize the browser's placeholder sentinel in build_game

Clicking `Keep the placeholder music` in the Look step persists `music_model: "keep-placeholder"`. The browser normalizes that value to null, but this server mirror passes it to `modelValue` and returns an invalid-model error. The analogous `sfx_node_type` sentinel is passed to the registry as a node type, producing placement issues. Thus a valid setup created in the UI cannot be handed over to the headless builder after explicitly choosing placeholder audio. Normalize both sentinel values the same way as `useGameSetupFlow`, preferably through a shared protocol helper, and cover a saved browser setup in the capability tests.

georgi (commented):
Reviewed 4e1a67d. Four actionable findings below, including two P1 blockers in the browser flow.

Validation: traced the new flow through the actual tRPC router, workflow event reducer, editor routing, and server capability. Isolated execution of the source hook with the server's array response returns undefined. Feeding the export node's terminal output buffer into the source summary leaves directory/archive null and verified false. The current tests mock different response shapes, so they miss these integration failures. Full project tests and a live Godot/browser run were not performed.

web/src/hooks/game/useGameTemplates.ts
georgi: Fixed in ba4cc0a. The hook reads the array `games.templates` returns, and the router's slot type is now the protocol's `gameSlotSpec`, so the client cast is gone and `GameTemplate` is the router's own output type. The hook test mocks the real array shape; `useGameStylePresets` lost its cast the same way (its shape already matched).

---

web/src/components/setup/game/gameRunSummary.ts
georgi: Fixed in ba4cc0a. Confirmed, and one row worse than described: the same edge suppression hit every checker's asset handle (each feeds `fills`), so `Assets checked` was stuck at 0 as well. The summary now reads the focused job's completed `liveGenerations[].outputs`, which keep handle names, are written by `generation_complete` and a completed `node_update`, and carry `verification.reason`. New test `gameRunSummary.events.test.tsx` drives the real `handleUpdate` with the runner's frames (one `output_update` per handle, node completions, `job_update`) and reads the summary off the stores. Inverting it back to `outputResults` reproduces exactly the reported nulls.

---

web/src/components/projects/NewProjectSurface.tsx
georgi: Fixed in ba4cc0a. `WorkflowEditorSurface` now renders `GameSetupHost` while `settings.game.stage` is not `done`, with `onFinish` flipping the same tab to the canvas, so a reload or reopen during Design or Look resumes the flow. New test `WorkflowEditorSurfaceGame.test.tsx` mounts the surface with stage `look`, asserts the Look step and no canvas, unmounts and remounts to assert it resumes there, and asserts `done` and a workflow with no `game` key open the canvas. Not widened to `settings.setup`: the Workflow flow's host needs the New Project composer's example-copy machinery, so that stays a separate gap.

---

packages/agents/src/capabilities/workflows.ts
georgi: Fixed in ba4cc0a. One protocol helper, `gameAudioChoice` with `GAME_PLACEHOLDER_SENTINEL` in `game-graph.ts`, now normalizes both `music_model` and `sfx_node_type` for the browser flow (`useGameSetupFlow`) and the `build_game` capability. The capability suite gains "builds a browser setup that kept the placeholder audio": no issues, no music or sfx chain placed.

---

web/src/components/setup/game/gameRunSummary.ts
```diff
@@ -0,0 +1,227 @@
...
+  const liveGenerations = useResultsStore((state) => state.liveGenerations);
+  const errors = useErrorStore((state) => state.errors);
+
+  return useMemo(() => {
+    const nodes = toRunNodes(nodeStore?.getState().nodes ?? []);
+    if (jobId === undefined) {
+      return summarizeGameRun({
+        nodes,
+        outputsFor: () => undefined,
+        errorFor: () => undefined
```
georgi, line 305: F2 follow-up [P2] Restore completed export results after a page reload

The live-run fix works, but both `focusedJob` and `liveGenerations` are non-persisted stores. Finish an export, then refresh: this branch supplies no outputs, so the persisted build checklist comes back with directory/archive null and all three export actions disabled. `startRunReconciliation` only reattaches scheduled/queued/running jobs, so it does not hydrate this completed run. The saved `settings.game.build` contains placement/run-start status but no export outcome either. Restore the last completed export from durable job results, or persist its outcome and hydrate the checklist from it, without marking a new in-progress run complete. Add a finished-run reload test with fresh stores.

web/src/components/setup/game/gameRunSummary.ts
```diff
@@ -0,0 +1,227 @@
...
+  // checker is the one whose result means the asset passed its check.
+  const bySlot = new Map<string, GameRunNode[]>();
+  for (const node of slotNodes) {
+    const slotId = node.slotId as string;
+    bySlot.set(slotId, [...(bySlot.get(slotId) ?? []), node]);
+  }
+  let checked = 0;
+  for (const [, chain] of bySlot) {
+    if (chain.some((node) => outputsFor(node.id) !== undefined)) {
+      checked += 1;
```
georgi, line 165: F5 [P2] Count the checker rather than any completed node in its chain

`gameGraphPlacement` assigns the same `setupStepId` to the generator, resize, checker and preview. With the new `liveGenerations` source, `chain.some(...)` becomes true immediately when TextToImage finishes, even though SpriteSheet/Tileset/SeamlessImage has not checked the asset. A generator-only completion reproduces `checked: 1, total: 1`, and a later checker failure leaves that slot counted as checked. Restrict this count to successful completion of the slot's checker node (or its accepted fill), and cover generator-complete/checker-pending and checker-failed cases.

georgi (commented):
Re-reviewed ba4cc0a against 4e1a67d. F1, F3 and F4 are addressed in the code. F2's live-run output mismatch is fixed, but its reload case remains below. Also found F5: the checked count now advances on generation rather than validation.

Validation: isolated execution of the source hook confirms the real template array is returned, the shared audio helper normalizes placeholder values, and named completed export outputs populate the checklist. The same source reports 1/1 assets checked with only the generator completed, and loses directory/archive when the in-memory stores are reset. Traced reopening through WorkflowEditorSurface and confirmed run reconciliation only restores in-flight jobs. Full project tests and live browser/Godot execution were not run.

web/src/components/setup/game/gameRunSummary.ts
georgi: Fixed in 674f2493. The export outcome is now written once onto `settings.game.build.export` (keyed by job id) when the focused job completes, so the landing panel reads it back after a reload. The live job still wins while one is focused; the persisted record is only used when no job is focused.

---

web/src/components/setup/game/gameRunSummary.ts
georgi: Fixed in 674f2493. The checked count now counts a slot only when its checker node (`SpriteSheet`/`Tileset`/`SeamlessImage`/`SoundEffect`/`MusicLoop`) produced outputs without an error. The checker node types are exported from the graph builder as `GAME_CHECKER_NODE_TYPES`, with a test asserting each slot chain contains exactly one, so the summary and the builder cannot drift.

---

georgi (commented):
Re-reviewed 674f2493 against ba4cc0a. F5 is fixed: generator completion and checker failure no longer count as accepted assets. F2 now restores a finished export when the checklist observed completion, and a newly focused run correctly suppresses the old result. One remaining F2 lifecycle case is below.

Validation: isolated execution of the source summary/hook with controlled stores and effects covered generator-only completion, failed/successful checkers, mounted completion followed by fresh stores, a new focused run, and completion while the panel is unmounted. Traced the actual mount gates in FloatingToolBar and WorkflowEditorSurface. Full project tests and a live browser/Godot run were not performed.

web/src/components/setup/game/gameRunSummary.ts
```diff
@@ -0,0 +1,321 @@
...
+  useEffect(() => {
+    if (jobId === undefined || live.directory === null || build === null) {
+      return;
+    }
+    if (build.export?.job_id === jobId || written.current === jobId) {
+      return;
+    }
+    written.current = jobId;
+    void setGame({
+      [GAME_BUILD_KEY]: { ...build, export: exportRecord(jobId, live) }
```
georgi, line 378: F2 follow-up [P2] Persist export completion independently of the visible checklist

This is the only writer of `settings.game.build.export`, but the hook only runs inside `GameLandingPanel`. `FloatingToolBar` mounts that panel only while `conversationOpen`, and the toolbar itself is mounted only for the active workflow tab. Repro: start a build, collapse the chat dock before export completes, wait for completion, then refresh before reopening the dock. No effect observed completion, so no export record was saved. The refreshed stores are empty and reconciliation excludes completed jobs, leaving directory/archive null again. Switching tabs or closing the page during generation has the same gap. Move persistence to the run-completion path, or recover the completed export from durable job results on reopen. Extend the reload test to unmount the panel before completion rather than pre-seeding its saved export.

georgi: Fixed in dbe92c02d. WebSocket runs now persist their normalized final outputs on the durable Job row. The jobs API exposes those outputs only when explicitly requested, and `useGameRunSummary` recovers the latest completed project export when no live job or saved export is available, then writes it to `settings.game.build.export`.

The regression test starts with the run focused, unmounts the summary before completion, clears the in-memory run stores, and reopens against the completed job output. It asserts the export actions and checklist recover without a pre-seeded export record.

PR #5666, opened 2026-09-06
Add muapi nodes

georgi:
## Verification

- [ ] `npm run test:affected` (the full `npm run test` +
      `npm run test:packages` pass is CI's job, not yours)
- [ ] `npm run typecheck`
- [ ] `npm run lint`
- [ ] `npm run dev:nodetool -- harness gate --base main`

## Agent capabilities

Skip this section if the diff adds no capability and changes no capability's
declared contract (name, description, input schema, permission category,
`needsToolCallId`).

- [ ] `npm run capabilities:check` passes
- [ ] Every added or re-declared capability names an eval case or a suite in
      `packages/cli/src/harness/capability-table.ts`, or carries a gap note
      saying what a check for it would look like

## New checks

Skip this section if the diff adds no check, rule, or audit.

- [ ] The check was inverted once and observed failing; the failing command is
      in **Verification** above
- [ ] An audit asserts it found its targets, so it cannot pass by matching
      nothing


georgi:
@claude 

1. **F1, P1: Image-to-video sends the wrong input field.** [[Provider line 144](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-provider.ts#L144)](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-provider.ts#L144) and the dedicated node send `images_list: [imageUrl]`. MuAPI requires `image_url`, so requests omit the required source image. Fix both paths and their mocked expectations. [[Endpoint documentation](https://muapi.ai/playground/flux-3-image-to-video)](https://muapi.ai/playground/flux-3-image-to-video).

2. **F2, P2: Video options contradict the API contract.** [[Transport lines 44–45](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L44)](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L44) advertise 4–10 seconds and 480p. MuAPI documents 5–20 seconds and 720p/1080p. This exposes unsupported choices and silently shortens valid requests. I reproduced a 20-second request becoming 10 seconds. [[Endpoint documentation](https://muapi.ai/playground/flux-3-text-to-video)](https://muapi.ai/playground/flux-3-text-to-video).

3. **F3, P2: Actual generation costs are discarded.** [[Cost parser](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L248)](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L248) expects a scalar `cost` or `credits_used`. MuAPI returns `cost.amount_usd`. Passing the documented response into this function returns `null`, leaving the receipt without the reported charge. Read the USD field and test the real response shape. [[API reference](https://muapi.ai/docs/api-reference)](https://muapi.ai/docs/api-reference).

4. **F4, P2: Download size protection happens after allocation.** [[Download line 355](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L355)](https://github.com/nodetool-ai/nodetool/blob/fb1ed8abd00235b712d1bef327c5be530792b265/packages/runtime/src/providers/muapi-transport.ts#L355) buffers the entire response before enforcing the 256 MiB limit. Without an accurate `Content-Length`, an oversized response can exhaust process memory. Read the stream incrementally and cancel when it crosses the limit.

Validation: source inspection, vendor documentation, and isolated execution of the cost/duration functions. Full tests were not run because dependencies are absent; CI was still pending.

georgi:
@claude address comment

georgi (commented):
Reviewed head `fa64da6e850a241ef192bafb324e69b3f4a3c23d`. I would not merge this revision yet. I confirmed two issues, with inline comments and suggested regression coverage:

1. **[P1] The new cost receipt breaks the backend build.** The transport passes a number where the protocol requires a structured cost object. This is the actual TS2322 failure in the current head's CI run, not just a hypothetical type mismatch.
2. **[P2] Retried CDN error bodies still bypass the download size limit.** The shared retry helper buffers discarded 429/5xx responses before the new bounded reader gets control. The successful-response path is bounded, but the retry path remains unbounded.

The earlier `image_url`, supported duration/resolution, and nested `cost.amount_usd` parsing fixes are present. The streaming-reader fix also works on a direct successful response; the second finding identifies the remaining retry-path gap rather than repeating the original finding unchanged.

Verification: reviewed the full 29-file diff and relevant shared protocol, receipt, and HTTP transport code; inspected the [failing CI build log](https://github.com/nodetool-ai/nodetool/actions/runs/34095508947/job/101664018136); ran an isolated TypeScript reproduction of the receipt contract (the current call fails, the structured-object call passes); and ran an isolated native-ReadableStream reproduction of the retry/limit ordering. With a scaled-down 20-byte limit, a direct 200 response was cancelled after crossing the limit at 24 bytes, while a retried 503 response was fully buffered at 800 bytes before a 16-byte final response passed the limit check. These were extracted-function/minimal-contract checks, not the repository test suite. I did not run the full suite locally or make live MuAPI generation calls.

packages/runtime/src/providers/muapi-transport.ts
```diff
@@ -0,0 +1,517 @@
...
+      },
+      onTimeout: () =>
+        new Error(
+          `MuAPI request ${requestId} did not finish within ${maxAttempts} poll attempts`
+        )
+    }
+  );
+
+  const cost = readMuapiCost(result);
+  if (cost !== null) recordGenerationReceipt({ cost });
```
georgi, line 350: [P1] Record the structured cost required by GenerationReceipt

`readMuapiCost()` returns `number | null`, but `GenerationReceipt.cost` in `packages/protocol/src/messages.ts` is an object with a required `amount` field. Consequently this line fails `npm run build:packages` with TS2322 and prevents the backend from building, including installations that do not enable the MuAPI node pack. The [current head's CI build](https://github.com/nodetool-ai/nodetool/actions/runs/34095508947/job/101664018136) reports exactly `muapi-transport.ts(350,48): Type 'number' is not assignable to type '{ amount: number; ... }'`.

Record the reported USD value as `recordGenerationReceipt({ cost: { amount: cost, currency: "USD" } })`. Update the added provider test, which currently expects `cost: 0.42`, to assert the structured receipt and preferably validate it with `generationReceiptSchema`. A minimal type-check reproduction fails with the current call and passes with the structured object.

packages/runtime/src/providers/muapi-transport.ts
```diff
@@ -0,0 +1,517 @@
...
+async function muapiDownload(
+  url: string,
+  kind: Extract<MediaKind, "image" | "video">,
+  signal?: AbortSignal
+): Promise<Uint8Array> {
+  assertSafePublicHttpsUrl(url);
+  const init: RequestInit = signal ? { signal } : {};
+  const response = await fetchWithRetry(url, init, {
+    fetchImpl: (input, requestInit) =>
+      safeFetch(String(input), requestInit as RequestInit)
```
georgi, line 416: [P2] Bound or cancel discarded CDN responses inside the retry path

The output size checks below run only after `fetchWithRetry()` returns. However, `http-transport.ts` calls `resp.arrayBuffer()` on every discarded retryable 429/5xx response. A CDN response that streams an oversized 503 body therefore gets buffered in full before `readBodyWithLimit()` is reached, bypassing the advertised 256 MiB ceiling and potentially exhausting the worker's memory.

I reproduced this ordering with the retry and reader functions in an isolated native-stream harness: with the reader limit scaled to 20 bytes, a direct 200 body was cancelled at 24 bytes, but a retried 503 body was buffered through all 800 bytes before the final 16-byte response succeeded. Cancel discarded response bodies or drain them incrementally with an enforced bound in the retry helper. Add regression coverage using a real streamed 503 response. The current download mock throws from `arrayBuffer()`, but the retry helper catches that exception, so that mock does not protect this path.

PR #5667, opened 2026-09-06
feat(agents): external MCP servers on the agent toolbelt

georgi:
## Summary

Users can add external MCP servers (Blender, Hugging Face, a local script) in Settings → MCP Servers → External servers, over stdio or HTTP. Their tools join the agent toolbelt on the next chat turn.

- Each remote tool is a plain `Tool` named `mcp_<server>_<tool>`. That one shape is why it works on every provider without provider changes: the base loop and the OpenAI Responses loop (Codex included) dispatch through `executeTool`, the Claude Agent SDK wraps the belt in its in-process MCP server, and a CodeAct guest imports from `@nodetool-ai/sandbox-nodetool/mcp`.
- `McpClientPool` keeps one client per server per user: reconnects on config change, closes removed servers, caches discovery with `nextCursor` paging, evicts on failure, serializes reconciliation.
- Config shape (`McpServerConfig`) is shared through `@nodetool-ai/protocol` by the tRPC router, the persistence layer, the pool, and the settings UI.

## Security

- A literal header or env value moves into an encrypted secret named `MCP_<SERVER>_<KEY as hex>` on save; only the `${reference}` is stored. References resolve from the user's own secrets, never `process.env`.
- The cloud profile refuses stdio servers and fetches HTTP servers through `safeFetch`, which re-checks every redirect hop. Private and http URLs are refused on save and probe.
- Resolved secret values are redacted from stderr, error logs, and probe results, per connection.
- Remote tools have no capability entry, so `gateTools` classifies them `external` and the permission ladder applies.

## Verification

- `npm run lint` clean, `harness gate` 4/4, web typecheck clean (with a larger heap; the default heap runs out on this machine regardless of the change).
- New suites: `packages/agents/tests/external-mcp-tools.test.ts` (a real MCP server over an in-memory transport), `packages/websocket/tests/external-mcp.test.ts` (persistence, credential externalization, cloud gates).
- Agents, websocket, and chat-turn suites pass. Three Codex review passes; every finding closed.

## Out of scope

Agent nodes in workflows and the CLI belt do not load external servers yet; chat is the surface here.

https://claude.ai/code/session_01XiCF7BrK6J3ybmB2ULvFND

### DIFF
--- a/packages/agents/src/tools/external-mcp-tools.ts
+++ b/packages/agents/src/tools/external-mcp-tools.ts
@@ -379,8 +379,7 @@ export class McpClientPool {
           .slice(0, MAX_STDERR_CHARS);
         if (text) log.debug("MCP server stderr", { server: config.id, text });
       });
-      closeOnAbort(signal, stdio);
-      await client.connect(stdio);
+      await startTransport(client, stdio, signal, config.id);
       return client;
     }
     const headers = await resolveSecretReferences(
@@ -400,18 +399,17 @@ export class McpClientPool {
         requestInit: { headers },
         ...fetchOpt
       });
-      closeOnAbort(signal, streamable);
-      await client.connect(streamable);
+      await startTransport(client, streamable, signal, config.id);
       return client;
     } catch (err) {
+      await client.close().catch(() => undefined);
+      // A cancelled entry must not fall through and open a second transport.
+      if (signal.aborted) throw cancelledError(config.id);
       log.debug("Streamable HTTP failed, trying SSE", {
         server: config.id,
         error: describeError(err, secrets)
       });
-      await client.close().catch(() => undefined);
     }
-    // A cancelled entry must not open a second transport on the way out.
-    if (signal.aborted) throw cancelledError(config.id);
     const { SSEClientTransport } = await import(
       "@modelcontextprotocol/sdk/client/sse.js"
     );
@@ -420,8 +418,7 @@ export class McpClientPool {
       requestInit: { headers },
       ...fetchOpt
     });
-    closeOnAbort(signal, sse);
-    await sseClient.connect(sse);
+    await startTransport(sseClient, sse, signal, config.id);
     return sseClient;
   }
 }
@@ -458,6 +455,29 @@ function closeOnAbort(
   const close = () => void transport.close().catch(() => undefined);
   if (signal.aborted) close();
   else signal.addEventListener("abort", close, { once: true });
+}
+
+/**
+ * Start one transport, refusing a startup its entry has already dropped.
+ *
+ * Everything before this point is awaited — resolving `${SECRET}` references,
+ * importing the SDK's transport module — so the user can disable the server,
+ * or the deadline can expire, while this connection is still being prepared.
+ * Closing an unstarted stdio transport does not mark it closed, so a
+ * `closeOnAbort` that fired during those awaits would not stop the `start()`
+ * below from spawning the command afterwards. The check and the start share
+ * one tick: `Client.connect` reaches `transport.start()` before its first
+ * await, so nothing can abort between them.
+ */
+async function startTransport(
+  client: Client,
+  transport: Parameters<Client["connect"]>[0],
+  signal: AbortSignal,
+  serverId: string
+): Promise<void> {
+  if (signal.aborted) throw cancelledError(serverId);
+  closeOnAbort(signal, transport);
+  await client.connect(transport);
 }
 
 /** Every page of `tools/list`. */
--- a/packages/agents/src/tools/external-mcp-tools.ts
+++ b/packages/agents/src/tools/external-mcp-tools.ts
@@ -1,0 +1,528 @@
+/**
+ * External MCP servers on the toolbelt.
+ *
+ * A user configures a server (Blender, Hugging Face, a local script) and its
+ * tools become ordinary {@link Tool}s named `mcp_<server>_<tool>`. Because
+ * they are plain belt tools, every loop reaches them the same way: the base
+ * provider loop and the Responses loop dispatch through the harness
+ * `executeTool`, the Claude Agent SDK wraps the belt in its in-process MCP
+ * server, and a CodeAct guest imports them from
+ * `@nodetool-ai/sandbox-nodetool/mcp`.
+ *
+ * Connections live in an {@link McpClientPool}: one client per server config,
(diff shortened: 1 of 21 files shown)


packages/agents/src/tools/external-mcp-tools.ts
```diff
@@ -0,0 +1,528 @@
...
+      });
+      await client.close().catch(() => undefined);
+    }
+    const { SSEClientTransport } = await import(
+      "@modelcontextprotocol/sdk/client/sse.js"
+    );
+    const sseClient = new Client({ name: "nodetool", version: "1.0.0" });
+    await sseClient.connect(
+      new SSEClientTransport(url, { requestInit: { headers }, ...fetchOpt })
+    );
```
georgi: F1 [P1] Bound transport startup and allow removal to cancel it

An HTTP server that rejects the Streamable HTTP POST, then opens an SSE stream without sending an `endpoint` event, leaves this await pending indefinitely. In SDK v1.27.1, `Client.connect()` awaits `super.connect()`/`transport.start()` before starting the timed initialize request, so its default request timeout does not cover this stage. `discover()` awaits this promise while holding the pool's serialized queue, and chat awaits discovery before assembling its belt. Disable/delete/save subsequently queue behind it, so the user cannot recover through Settings. Even without discovery, `closeEntry()` waits for the unresolved client. Add a startup deadline that closes the actual transport, plus cancellation that can interrupt pending connections without waiting behind discovery. A regression test should keep SSE open without an endpoint and verify chat skips the server and removal completes.

packages/agents/src/tools/external-mcp-tools.ts
```diff
@@ -0,0 +1,528 @@
...
+ * counts: a literal the user typed into the config is not a secret to blank,
+ * and blanking `true` or `1` would eat unrelated text.
+ */
+function rememberResolved(
+  raw: Record<string, string>,
+  resolved: Record<string, string>,
+  secrets: Set<string>
+): void {
+  for (const [key, value] of Object.entries(resolved)) {
+    if (value && value !== raw[key]) secrets.add(value);
```
georgi: F2 [P2] Redact individual referenced secrets, not just the expanded header

For the advertised `Authorization=Bearer ${HF_TOKEN}` setup, this records only `Bearer <token>`. An upstream error such as `Invalid token <token>` therefore passes through `describeError`/`redactedError` unchanged into logs, probe results, or the tool answer. I reproduced this with the extracted helpers: a remembered `Bearer test-secret` leaves `Invalid token test-secret` intact. Record each value returned by the secret resolver during recursive expansion as well as the final header, so the underlying HF_TOKEN is redacted independently of its surrounding template.

georgi (commented):
Reviewed a717d5d. Three actionable findings below: F1 can wedge chat and MCP configuration changes, F2 leaves referenced credentials exposed in diagnostics, and F3 silently loses resource-link results.

Validation: ran isolated checks against the PR's extracted pool/redaction/result-normalization code and inspected the pinned MCP SDK v1.27.1 transport startup path. The returned PR workflow runs are green. I did not run the full repository suite locally.

packages/agents/src/tools/external-mcp-tools.ts
```diff
@@ -0,0 +1,528 @@
...
+    } else if (block.type === "image" && isString(block.data)) {
+      out.images.push({
+        data: block.data,
+        mimeType: isString(block.mimeType) ? block.mimeType : "image/png"
+      });
+    } else if (block.type === "resource" && isRecord(block.resource)) {
+      const res = block.resource;
+      if (isString(res.text)) parts.push(res.text);
+      else if (isString(res.uri)) parts.push(`[resource ${res.uri}]`);
+    }
```
georgi, line 608: F3 [P2] Preserve resource_link blocks in remote tool answers

The normalizer handles embedded `resource` blocks but silently drops MCP `resource_link` blocks. For example, `{content:[{type:'resource_link',uri:'https://example.com/render.glb',name:'render.glb'}]}` normalizes to empty text/no images, and `ExternalMcpTool.process()` returns `{}`. A successful file-producing tool whose result is a resource link thus gives the agent no URI to use or show. Preserve at least the URI, name, MIME type, and description, either as structured output or a text representation, and cover a resource-link-only result in the tests.

georgi:
@claude address comments

packages/agents/src/tools/external-mcp-tools.ts
```diff
@@ -0,0 +1,693 @@
...
+      // fills and blocks a server that logs before it answers `initialize`.
+      // Only a bounded line reaches the log, never the environment.
+      stdio.stderr?.on("data", (chunk: Buffer | string) => {
+        const text = redact(chunk.toString(), secrets)
+          .trim()
+          .slice(0, MAX_STDERR_CHARS);
+        if (text) log.debug("MCP server stderr", { server: config.id, text });
+      });
+      closeOnAbort(signal, stdio);
+      await client.connect(stdio);
```
georgi: [P2] Stop before starting a transport when cancellation already happened

If removal or the startup deadline fires while `resolveSecretReferences()` is awaiting a secret, the pool drops this entry successfully, but `connect()` resumes here afterward. `closeOnAbort()` sees the already-aborted signal and closes the not-yet-started transport, then the next line starts it anyway. The SDK's stdio `close()` does not mark the transport permanently closed, so this still spawns the configured command after the user disabled/deleted it. The same pattern exists around the asynchronous HTTP transport imports. I reproduced the ordering with a deferred secret resolver and an injected transport: after `sync([])` completed, releasing the secret produced one transport start while `serverIds()` remained empty. Check `signal.throwIfAborted()` immediately before each transport start (after the intervening awaits), or make the startup helper reject an already-aborted signal. Add a regression that cancels while secret resolution is pending and asserts zero transport starts after releasing it.

georgi (commented):
Re-reviewed b943ab0 against the previous findings. The original stalled-startup/removal cases, nested bearer-token redaction, and resource-link forwarding now pass isolated checks against the extracted implementation. One cancellation race remains, detailed inline.

Validation used injected transport boundaries and the SDK's stdio startup/close implementation, not the full repository suite. The latest PR workflow runs (Test, Page Load Smoke, E2E Workflow Runner, Docs CI, Docs Lint) all report `action_required`, so this commit does not yet have passing CI verification.

georgi:
@claude address new comment

PR #5668, opened 2026-09-06
feat(timeline): audio-driven motion, subject mattes, adjustment clips, hold and path presets

georgi:
## What changed

The timeline gains three outcomes: motion that responds to sound, subjects that can be isolated and reused, and treatments applied across a section. Custom animations can place keyframes on the media clock (`timeBase: "source"`), so a curve baked from a signal stays aligned through speed, in-point and time remap, and trim and split re-slice it instead of stretching it. On that base: `bake_audio_animation` maps an audio clip's loudness envelope or onsets onto a target clip's scale, opacity or offset, with an inspector panel and route; `isolate_subject` runs BiRefNet v2 video with `output_mask` and attaches the returned luma mask to the source clip as a `generatedMatte`, so it shares the clip's window by construction, with stale detection, versions, knobs, a matte view and a route; adjustment clips treat the composite of every layer beneath them within their window, bottom-up when stacked and scoped to a group's own surface, in both compositors and the editor. Smaller pieces: a `hold` easing, seeded-noise `shake` and `float`, a `followPath` preset sampled by arc length, the `set_baked_animation` and `set_generated_matte` ops shared by every host, validator checks for adjustment and stale-matte cases, and a FAL factory fix so a single-asset endpoint keeps its declared secondary outputs (the `mask_video` field was dropped before).

## Verification

- [x] `npm run test:affected` — the protocol schema changed, so this was the full run. Packages: every suite green except `image-nodes`, which needs a Vulkan driver this sandbox lacks (documented in AGENTS.md), plus two load-sensitive cases (`runtime/host-binaries`, `blender-nodes/settings`) that pass on rerun. Web: 15664 passed, two suites timed out under load and pass on `--onlyFailures` rerun. Electron: 692 passed. Mobile: 1108 passed.
- [x] `npm run typecheck` — web and electron clean. Mobile reports errors in untouched app-runtime fixtures because this sandbox resolves the published `@nodetool-ai/protocol` rc.11 rather than the workspace build.
- [x] `npm run lint` — exit 0, warnings only and all pre-existing.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — 12/12 after two fixes surfaced by the first run: the url-egress audit needed the isolate-subject fetch inventoried, and the secrets audit needed the isolate-subject route to pass `availableSecrets`.

Inverted checks, each observed red then restored (details in the commit messages and test headers): source-time sampling neutered in `sample.ts` reds 5 source-curve tests; re-slice disabled in `trimClip`/`splitClip` reds 4; `hold` mapped to linear reds 4; the validator's source-window push gated off reds exactly its test; stacked adjustments' z swapped reds "runs stacked adjustments bottom-up"; `strength < 1` inverted in `canvas2d.ts` reds the strength pixel test; `isGeneratedMatteStale` negated reds 4 of 6 validator cases; the audio-drive demo's expected peak moved by 999 ms fails with `expected 1027 to be less than or equal to 40`; the failure branch in `isolate_subject` forced off reds "keeps the previous ready matte selected when the generation fails".

## Agent capabilities

- [x] `npm run capabilities:check` passes (table regenerated for `bake_audio_animation`, `isolate_subject`, and the changed `edit_timeline` description)
- [x] `bake_audio_animation` is covered by `timeline-audio-drive` (`timeline-audio-drive-demo` + `capabilities-bake-audio-animation`); `isolate_subject` by `timeline-generated-matte` (`capabilities-isolate-subject`); both entries are in `packages/cli/src/harness/registry.ts` and `docs/harnesses.md`

## New checks

- [x] Each new check was inverted once and observed failing; the list is under **Verification**
- [x] The url-egress and secrets audits already assert they found their targets; the new validator checks each have a red case and a green case

## Not in this PR

- A live BiRefNet run was not possible here (no FAL key). The provider seam is injected and the mask format (grayscale H.264 mp4, luma-matted) is taken from the manifest; the quality pass on hair, fast motion and occlusion still needs a keyed run.
- Point tracking (attach a clip to a tracked point) is not started: no tracking model is in the provider manifests.

https://claude.ai/code/session_013in93XM9r3h7RaFz62YGwN

---
_Generated by [Claude Code](https://claude.ai/code/session_013in93XM9r3h7RaFz62YGwN)_


web/src/stores/timeline/TimelineStore.ts
```diff
@@ -2240,6 +2316,235 @@ export const createTimelineStore = (
...
+            // The editor may have moved to another sequence while the matte
+            // ran; loading this one over it is the clobber every reload path
+            // avoids.
+            if (get().sequenceId !== sequenceId) return undefined;
+            get().applyAgentEdit({
+              tracks: (sequence.tracks ?? []) as TimelineTrack[],
+              clips,
+              markers: (sequence.markers ?? []) as TimelineMarker[]
+            });
+            get().setBaseUpdatedAt(sequence.updatedAt);
```
georgi: F1 [P1] Preserve pending editor changes when adopting the generation result

The sequence-id check only protects navigation to another timeline. If the user edits a clip, adds a marker, or deletes a track while the request/query is in flight, this replaces all three document arrays with the server snapshot. For example, edit another clip after the GET captures its response but before it resolves, within the 750 ms autosave debounce: `applyAgentEdit` discards that edit, then `setBaseUpdatedAt` marks the replacement as the synchronized document. The audio bake completion at lines 2356–2361 has the same problem. These paths bypass the dirty-document merge already used by `useTimelineExternalSync`. Merge the returned change against the saved baseline/current draft, or apply only the generated fields with a conflict check, instead of adopting the whole snapshot. Add a deferred-response test that edits an unrelated clip before completion and verifies the edit survives.

packages/agents/src/capabilities/timeline-audio-bake.ts
```diff
@@ -0,0 +1,219 @@
...
+): BakedCurvePoint[] {
+  const { targetClip } = options;
+  const windowStartMs = targetClip.startMs;
+  const windowEndMs = targetClip.startMs + targetClip.durationMs;
+  const mapped: BakedCurvePoint[] = [];
+  for (const point of points) {
+    const timelineMs =
+      audioSourceMsToTimelineMs(options.audioClip, point.timeMs) +
+      options.offsetMs;
+    if (timelineMs < windowStartMs || timelineMs > windowEndMs) continue;
```
georgi: F2 [P2] Interpolate the target boundaries before dropping curve points

The incoming curve has already been simplified, so absence of a keyframe inside the target does not mean absence of overlap. Reproduction: a linear envelope over 0–4000 ms simplifies to `(0, 1), (4000, 2)`; a target spanning timeline 1000–3000 ms then receives `[]`, and `bake_audio_animation` incorrectly returns "does not overlap". With only one retained point, the boundary portions instead become flat and lose the measured motion. Clip the piecewise-linear curve geometrically, inserting interpolated values at the target's start/end before discarding outside vertices. The example should yield target-source points `(0, 1.25), (2000, 1.75)`.

packages/timeline/src/render/canvas2d.ts
```diff
@@ -444,21 +504,143 @@ export function drawTimelineFrame<TSource>(
...
+
+  ctx.save();
+  ctx.setTransform(1, 0, 0, 1, 0, 0);
+  ctx.globalAlpha = Math.max(0, Math.min(1, adjustment.opacity));
+  ctx.globalCompositeOperation = "source-over";
+  if (shape && !shapeApplied) clipMask(ctx, shape, w, h);
+  if (wipe && !softWipe) clipWipeRect(ctx, w, h, wipe);
+  ctx.filter = filterForEffects(adjustment.effects, undefined, lifts);
+  try {
+    ctx.drawImage(scratch.surface, 0, 0, w, h);
```
georgi, line 689: F3 [P2] Interpolate the treated composite instead of source-over compositing it twice

The scratch surface contains the accumulation's existing alpha, so drawing it source-over that same accumulation increases opacity wherever the surface is translucent. With a real Canvas 2D surface containing a 50%-opaque pixel and an enabled neutral color effect (`contrast: 1`) at adjustment opacity 1, alpha changes from 127 to 191. This affects adjustments inside group surfaces and transparent exports, including softened subject edges. A fully applied color treatment must replace the underlying composite there, and partial/masked treatments should interpolate original and treated premultiplied RGBA using adjustment coverage. The GPU `blendStack` adjustment branch uses the same source-over approach and needs the corresponding correction. Add a neutral-adjustment alpha-invariance test on a transparent surface.

packages/agents/src/capabilities/timeline-audio-bake.ts
```diff
@@ -0,0 +1,219 @@
...
+  const points: KeyframePoint[] = [{ timeMs: windowStartMs, value: lo }];
+  let lastTimeMs = windowStartMs;
+
+  for (const onsetMs of [...onsetsMs].sort((a, b) => a - b)) {
+    if (onsetMs <= lastTimeMs || onsetMs > windowEndMs) continue;
+    const riseMs = Math.max(
+      onsetMs - Math.max(0, options.attackMs),
+      lastTimeMs + EPSILON_MS
+    );
+    if (riseMs >= onsetMs) continue;
```
georgi: F4 [P2] Keep beat peaks when attack_ms is zero

Zero is accepted by the route and inspector, but it makes `riseMs === onsetMs`, so this guard skips every onset. Calling `beatCurve([1000, 2000, 3000], { attackMs: 0, releaseMs: 150, outputRange: [1, 1.2], windowMs: [0, 4000], tolerance: 0.01, maxPoints: 4096 })` returns only the two resting endpoints. Choosing an instantaneous attack therefore disables beat-driven motion. Handle zero attack explicitly, retaining the peak with a hold/step or a minimal pre-onset interval, and cover it with a test that asserts each onset still reaches the high value.

georgi (commented):
Reviewed ebea68e3407928b6ce71895f3879a536da84095a. Four actionable findings below, including a pending-editor-edit overwrite that should be fixed before merge.

Validation: executed the new audio curve functions against a simplified linear envelope and zero-attack beat train, and executed the Canvas 2D adjustment function against a real @napi-rs/canvas surface. Those reproduced F2–F4. F1 follows from tracing the completion handlers, applyAgentEdit, and the existing autosave/external-sync paths. This was targeted validation, not a full monorepo test run or a live FAL generation.

packages/agents/src/capabilities/timeline-audio-bake.ts
georgi: Fixed in 67d7a981. The curve is now clipped geometrically: every point is mapped to timeline ms, an interpolated value is inserted at the target's start and end wherever a segment crosses them, and only then are outside vertices dropped. Your example yields `(0, 1.25), (2000, 1.75)`, pinned by a test, plus a between-two-keyframes case and a genuine no-overlap case that still refuses.

---

packages/agents/src/capabilities/timeline-audio-bake.ts
georgi: Fixed in 67d7a981. With `attack_ms: 0` the rise point now sits one epsilon (the module's ascending floor) before the onset and the peak lands on the onset itself, so each beat is a step to the high value. Equal timestamps were not usable because `evalCurve` collapses a zero-span segment. Your call now yields a peak at 1000, 2000 and 3000 ms; a test pins it and a second one pins the unchanged non-zero attack shape.

---

web/src/stores/timeline/TimelineStore.ts
georgi: Fixed in 2c8726ea. Both completions now go through one helper that runs the existing `mergeTimelineDocuments` three-way merge: base is the document the action sent, draft is the current store, server is the fetched copy, and the merge is scoped with synthesized update-clip ops for the clips the route wrote so the generated field lands while unrelated edits win. The isolate loop rolls its base forward each poll round. Write-back is a single `applyAgentEdit`, so it stays one undo entry, and `setBaseUpdatedAt` receives the merged base the way the external-sync hook does. Two deferred-response tests, one per action, rename a clip, add a marker and delete a track while the GET is pending and assert all three survive; each failed against the old adopt-the-snapshot code.

---

packages/timeline/src/render/canvas2d.ts
georgi: Fixed in 7b98493e on both compositors. The treated result is now mixed by coverage rather than composited over: `out = original * (1 - c) + treated * c` on every premultiplied channel, alpha included, with `c = opacity × mask × wipe`, so a full treatment replaces and a partial or masked one interpolates. Canvas 2D clears the destination and draws the treated scratch as the source at full coverage, and lerps the snapshot against the treated readback per pixel otherwise. The GPU path grades the accumulation premultiplied and a new mix pass blends it with the untreated accumulation by a coverage texture, wipe profile included. Tests pin your case (neutral colour effect at opacity 1 on a 50%-alpha pixel stays at alpha 127, previously 191), the half-opacity midpoint, and a masked region, on Canvas 2D and on the GPU; the GPU case ran against lavapipe, not skipped.

---

web/src/stores/timeline/TimelineStore.ts
```diff
@@ -2240,6 +2421,261 @@ export const createTimelineStore = (
...
+            // the placeholder `mark` just wrote onto the clip. The placeholder
+            // is this action's own optimistic state, not a user edit — leaving
+            // it out of the base would make the clip read as edited on both
+            // sides, and the merge would refuse the matte the run produced.
+            const sent = syncedSnapshotOf(beforeSave);
+            const marked = get().clips.find((c) => c.id === clipId);
+            let base: TimelineSyncedDoc = marked
+              ? {
+                  ...sent,
+                  clips: sent.clips.map((c) => (c.id === clipId ? marked : c))
```
georgi: F1 [P1] Build the optimistic base from the sent clip, not the live clip

This still loses edits to the target made while the initial `timeline.update.mutate` is pending. `marked` is read from the live store after the await, so it includes those unsaved edits, and copying the entire clip into `base` tells the merge engine they already existed on the server. Reproduction: save a clip named `Original`, rename it to `User rename` before that save resolves, then receive the ready matte on the server's `Original` clip. The actual `adoptServerSequence` + merge engine treats the draft as unchanged from this synthetic base and restores `Original`, losing the rename. Build this baseline from the target in `sent.clips` and overlay only this action's optimistic `generatedMatte`, preserving every other sent field. Add a deferred initial-save test, since the deferred GET tests exercise a later timing window.

web/src/stores/timeline/TimelineStore.ts
```diff
@@ -1191,6 +1273,105 @@ const syncedSnapshotOf = (
...
+    width: sequence.width ?? base.width,
+    height: sequence.height ?? base.height
+  };
+
+  const { doc, nextBase } = mergeTimelineDocuments(
+    base,
+    draft,
+    server,
+    clipWriteOps(touchedClipIds)
+  );
```
georgi: F5 [P1] Handle same-clip merge conflicts instead of silently dropping the generated result

`mergeTimelineDocuments` merges a timeline clip atomically, not by field. If the user renames or moves the target after the initial save, the draft clip differs from the base and the returned clip differs because its matte/animation was generated. The merge keeps the draft and returns the finished server clip in `conflicts`, but this helper discards that list and advances `baseUpdatedAt` anyway. In a focused execution, a renamed target with the optimistic `{assetId: '', status: 'generating'}` kept that placeholder after adopting a server clip with `{assetId: 'paid-mask', status: 'ready'}`; the merge returned one conflict, which the helper ignored. The action returns success while the inspector remains generating, and a subsequent autosave can overwrite the finished matte. Audio bakes can likewise vanish when their target was edited. Merge the generated fields separately where safe and surface genuine conflicts through the existing conflict flow, rather than dropping them. Cover edits to the target itself, not just unrelated clips.

georgi (commented):
Follow-up review of bc4a908a52377155b0718e35fd57f6af83c07262.

F2 and F4 are fixed in the focused reproductions: the inner target now receives interpolated points (0, 1.25) and (2000, 1.75), and zero attack retains all three beat peaks. F3's Canvas 2D reproduction now preserves alpha 127; the GPU implementation also uses the requested premultiplied mix on inspection, although I did not execute the GPU path this round.

F1 is only partially fixed. The new adoption helper preserves unrelated edits but still mishandles edits to the target clip in two distinct timing windows, detailed inline. I reproduced both using the current adoption helper and current merge engine with controlled base/draft/server snapshots. These should be addressed before merge. Validation was targeted, not a full suite or live FAL run.

web/src/stores/timeline/TimelineStore.ts
georgi: Fixed in d4dd8593. The isolate base is now built from the target in `sent.clips` with only the generating placeholder overlaid (the placeholder body moved into a pure `placeholderMatte` used by both the store write and the base), so nothing is read back off the live store after the save await. Your reproduction is pinned by a deferred-initial-save test per action: rename during the pending save, then adopt the server's clip; the rename survives and the generated field arrives. Both went red against the previous base construction.

---

web/src/stores/timeline/TimelineStore.ts
georgi: Fixed in d4dd8593. After the document merge, each touched clip gets a field-scoped pass (`adoptGeneratedClipField` in `merge.ts`). When the draft's generated field still equals the base's, the draft changed other fields only: the server's `generatedMatte` (or the one baked animation found by `findBakedAnimationIndex`) is overlaid onto the draft clip, the clip's conflict is dropped, and its `nextBase` entry becomes the server clip so a later autosave cannot overwrite it. When the draft changed that field too, the draft stands, the server clip is offered through the same conflict banner the external-sync hook uses (via the shared `timelineConflictKey`, accept overlays only the generated field), and the action returns `pendingUserResolution: true` instead of plain success. Your placeholder case is pinned: a renamed target now takes `{assetId: "paid-mask", status: "ready"}`; a cleared matte and a deleted baked curve land in the banner. All six store cases went red against the previous helper.

---

web/src/stores/timeline/merge.ts
```diff
@@ -214,3 +218,159 @@ export function mergeTimelineDocuments(
...
+      resolved.add(clipId);
+      continue;
+    }
+
+    rebased.set(clipId, baseClip);
+    pending.push(
+      merged.conflicts.find(
+        (conflict) =>
+          conflict.unit.kind === "clip" && conflict.unit.id === clipId
+      ) ?? {
```
georgi, line 353: F6 [P2] Treat matching draft and server generated fields as already resolved

This branch assumes that any draft field differing from the original base contests the result, but the draft may already contain exactly that result. For example, live document sync can load the completed matte before this action's GET finishes, while the action still holds its pre-run base. With a generating matte in base and the identical ready matte in both draft and server, `mergeTimelineDocuments` correctly returns zero conflicts, then this pass manufactures one pending conflict and restores the old base. The action reports `pendingUserResolution: true` even though the result is already applied. Check draft-field equality with the generated server field before treating it as a conflict, and advance the generated field's base when they agree. Add a converged base/draft/server test and a completion-after-live-sync case.

georgi (commented):
Re-reviewed 3d3b80c62dfc83cd9837cf26939c032e11bbb1af. The previous F1 baseline construction is corrected, and the F5 field-scoped adoption now preserves a target rename while applying the ready matte. A genuine cleared-matte conflict is also retained for resolution.

Two narrower P2 regressions remain in the new field-adoption logic, detailed inline. Focused executions of the current merge engine, adoption pass, and audio-field selector reproduced both. No full suite, GPU run, or live provider generation was performed this round.

web/src/stores/timeline/TimelineStore.ts
```diff
@@ -86,17 +91,120 @@ import type { Asset } from "../ApiTypes";
...
+  const indexIn = (clip: TimelineClip): number =>
+    findBakedAnimationIndex(
+      clip.animations,
+      AUDIO_BAKED_ANIMATION_KIND,
+      property
+    );
+  return {
+    valueOf: (clip) => {
+      const index = indexIn(clip);
+      return index < 0 ? undefined : clip.animations?.[index];
```
georgi: F7 [P2] Select the audio animation produced by this bake, including append mode

`findBakedAnimationIndex` returns the first audio bake driving the property, which is not necessarily the animation this request produced. With an existing scale bake A and `replace: false`, the server returns `[A, B]` and B's `animationId`. If the user renames the target during the request, the atomic merge keeps the draft `[A]`; this selector reads A on base, draft, and server, so `adoptGeneratedClipField` decides nothing was generated and never adopts B. The focused reproduction leaves only A, reports a clip conflict but `pending: false`, and accepting that conflict also overlays A, so the offered result cannot recover B. Thread the returned `animationId` into the field selector and overlay that exact animation, appending when appropriate. Cover append mode with an existing same-property bake and a concurrent target edit.

web/src/stores/timeline/merge.ts
georgi: Fixed in 27a18e50. `adoptGeneratedClipField` now checks the draft's generated field against the server's before treating a draft/base difference as a conflict. When they agree structurally the clip is resolved: the draft stays untouched, no conflict is offered, `pendingUserResolution` is not set, and the clip's `nextBase` entry advances to the server clip so the base carries the finished value. Pinned by a converged base/draft/server unit case and a completion-after-live-sync store case per action, each of which failed before the branch existed.

---

web/src/stores/timeline/TimelineStore.ts
georgi: Fixed in 27a18e50. The bake action passes the route's `animationId` into `bakedAudioAnimationField(property, animationId)`, so `valueOf` selects the animation with that id (the property selector is only the fallback when no id was returned) and the overlay writes exactly that animation: replace the same-id entry if the draft has one, otherwise append, carrying every other animation through. The banner's accept path uses the same spec. Your append case is pinned: an existing scale bake A, `replace: false`, and a rename during the request now end with `[A, B]`, the rename kept and no conflict; replace mode with a concurrent rename still replaces A. The append case and the id-selector unit case both failed under the property selector.

---

PR #5669, opened 2026-09-06
feat(timeline): add the WT-1, BL-1 and DR-1 instruments to midi tracks

georgi:
## What changed

A midi track could play one voice — a saw through a lowpass. This ports the three FableSynth instruments ([github.com/georgi/fablesynth](https://github.com/georgi/fablesynth)) into the offline renderer as three more members of the `MidiInstrument` union, so a track can play any of them: `wavetable` (WT-1) is two morphing wavetable oscillators plus a sub and noise through a Cytomic filter swept by a second envelope, with the tables built from harmonic spectra and band-limited into a mip ladder so a high note does not alias; `bass` (BL-1) is a monophonic acid line where a note at or above `accentVelocity` hits harder and sweeps further and a note overlapping the one before it glides into pitch without retriggering the envelopes; `drum` (DR-1) is sixteen pads from MIDI 36, each a tuned one-shot with a pitch envelope, coloured noise and a ring modulator, ringing for its own decay rather than the note's length. Eight presets ship with them, reachable from the track header's voice picker and from `set_track_instrument` by id, and the track instrument panel gains the filter/drive/level controls for the new voices and the kit's pad list. Not ported: stereo (this renderer is mono), the LFOs and mod matrix, the FX racks, and the plugins' own sequencers — a track's effects live on the track and its notes live in the clip.

Everything funnels through `renderInstrumentEvents`, so the browser preview, the audition, the CLI render and the `RenderTimeline` node all get the new voices from one switch. `instrumentSignature` now walks the instrument instead of listing its fields, because these voices carry nested oscillators, envelopes and pads and a hand-written list would go stale on the first field added — the render cache would then serve audio from before the change.

## Verification

- [x] `npm run test:affected` — the timeline, protocol, agents and web-timeline suites pass (`packages/timeline` 960 passed, `packages/protocol` 1310 passed, `packages/agents` 4423 passed, web timeline 1278 passed). Two pre-existing environment failures are unrelated to this diff: `packages/image-nodes` vignette tests (no Vulkan driver for WebGPU in this container) and `packages/video-nodes` `timeline-time-remap-decode` (ffmpeg hook timeout).
- [x] `npm run typecheck` — web and electron clean. Mobile fails only because `mobile/node_modules` is not installed in this container (`Cannot find module 'react-native'`), which is unrelated to the diff.
- [x] `npm run dev:nodetool -- harness gate --base main` — 8/8 selfchecks pass, `mappingViolations: []`.

Two gate failures were fixed after the first push, both this diff's own:

1. `capabilities:check` reported `capability-table.ts is stale` — listing the new voices in the `edit_timeline` spec moved its contract fingerprint. Regenerated with `npm run capabilities:sync`. (An earlier revision of this description wrongly called that pre-existing; it was not.)
2. The gate then reported a mapping violation: `edit_timeline`'s contract moved while its coverage mapping stood still. Answered the way the rule asks — with the test, in the third commit.

New tests in `packages/timeline/tests/midi.engines.test.ts` measure the rendered samples rather than the parameters: the mip ladder holds nothing above the harmonic a level allows, a WT-1 note at C7 has no off-harmonic content within 60 dB, BL-1 renders two same-frame notes as the later one alone and keeps its envelope running through a slide, and a DR-1 pad answers to one note and is silent above the kit. `tests/midi.protocolCompat.test.ts` now parses every shipped preset through the protocol schema and asserts a preset exists for all four instrument types.

## Agent capabilities

- [x] `npm run capabilities:check` passes.
- [x] The re-declared contract names a suite that covers it: `packages/agents/tests/timelines-op-input.test.ts` now drives `ui_timeline_set_track_instrument` with `wt1-bloom-pad`, `bl1-acid` and `dr1-tr-void`, asserts the preset is resolved and stored whole rather than kept as a reference, and asserts the kit keeps its sixteen pads and its base note through the round trip.

No capability is added and no schema changes — `set_track_instrument` takes the same shape, with more voices reachable through it.

## New checks

The alias test is the one new check that could pass by measuring nothing, so it carries its own control: the same analyzer is run over a naive (deliberately aliasing) saw at the same pitch and must report the foldover above −30 dB, while the wavetable render must be below −60 dB. An earlier version of the test read the analysis window's own sidelobes as aliasing (−32.7 dB with a Hann window); the Blackman-Harris window and the control were added because of it.

---

https://claude.ai/code/session_017cAXfhNs91xuFq95uwfK1C

### DIFF
--- a/packages/agents/src/capabilities/timelines.specs.ts
+++ b/packages/agents/src/capabilities/timelines.specs.ts
@@ -213,10 +213,16 @@ export const EDIT_TIMELINE_SCHEMA: JsonSchema = {
         "the clip hides notes instead of deleting them. " +
         'set_track_instrument takes {"track", "instrument"}, either a named ' +
         'voice — {"preset": "bass"}, one of saw-lead, square-lead, soft-pad, ' +
-        "pluck, bass, bell — or the synth spelled out: {type: " +
+        "pluck, bass, bell, the FableSynth WT-1 wavetable patches " +
+        "wt1-prime-lead, wt1-bloom-pad, wt1-vox-morph, wt1-chime-bell, the " +
+        "BL-1 acid basslines bl1-acid, bl1-deep, bl1-rubber, and the DR-1 " +
+        "drum kit dr1-tr-void — or the synth spelled out: {type: " +
         '"subtractive", waveform, attackMs, decayMs, sustain, releaseMs, ' +
-        "cutoffHz, resonance, gainDb}. get_state reports the preset a track's " +
-        "instrument matches as presetId. " +
+        "cutoffHz, resonance, gainDb}. dr1-tr-void plays one drum per note " +
+        "from 36 (kick, kick 2, snare, clap, rim, closed hat, open hat, ride, " +
+        "three toms, crash, two percs, vox, glitch) and is silent above 51. " +
+        "get_state reports the preset a track's instrument matches as " +
+        "presetId. " +
         'set_tempo takes {"bpm", offset_ms?, beats_per_bar?, beat_unit?} and ' +
         "rescales every midi clip around offset_ms — halving the BPM doubles " +
         "each midi clip's start and length — while picture and audio stay " +
--- a/packages/agents/tests/timelines-op-input.test.ts
+++ b/packages/agents/tests/timelines-op-input.test.ts
@@ -12,7 +12,7 @@ import {
   initTestDb
 } from "@nodetool-ai/models";
 import { createCapabilityRun, UNGATED } from "../src/capabilities/invoke.js";
-import { ANIMATED_PROPERTIES } from "@nodetool-ai/timeline";
+import { ANIMATED_PROPERTIES, findInstrumentPreset } from "@nodetool-ai/timeline";
 import {
   createTimelineToolBridge,
   type TimelineAnimationBakeRequest,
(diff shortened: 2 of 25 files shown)


georgi:
Review of `48905704efa730c5df08e913bf05693f9800045c`: one actionable finding.

### [P2] BL-1 can silently drop notes after a nested overlap

[`packages/timeline/src/midi/engines/bl1.ts:55–56`](https://github.com/nodetool-ai/nodetool/blob/48905704efa730c5df08e913bf05693f9800045c/packages/timeline/src/midi/engines/bl1.ts#L55-L56)

`monoOrder()` keeps the maximum gate-off of all preceding notes, but `renderBassVoices()` gates the envelope using only the current event's `gateOffFrame`. Those two definitions of an active note disagree after a shorter note takes over from a longer one.

Reproduced with the shipped BL-1 Acid patch at 48 kHz, all velocities 100:

- Pitch 40: 0–1000 ms.
- Pitch 47: 200–300 ms.
- Pitch 52: 500–700 ms.

The second note releases the envelope at 300 ms, and its 80 ms release finishes before the third onset. The third note is nevertheless marked `slide: true` because `previousGateOff` is still 1000 ms. It therefore reuses the `done` envelope and immediately exits the rendering loop. Its entire 500–700 ms window is exactly zero. Shortening the first note to 250 ms makes the third note audible, with a raw engine peak of approximately 0.493.

Make slide detection agree with the gate/envelope lifecycle. Under the documented next-note-takes-over behavior, track the immediate predecessor's gate rather than the historical maximum, and add this nested-overlap sequence as a regression test. If held-note fallback is intended instead, it needs an active-note/gate model that also keeps the envelope alive.

Validation: executed the PR's BL-1 and shared DSP source locally with Node's TypeScript transformation, changing only runtime import extensions. The four reported PR workflow runs are green. I did not run the full repository test suite locally.

georgi:
Confirmed and fixed in `19ff366b`.

Your repro is exact: I added the sequence (40 over 0–1000 ms, 47 over 200–300 ms, 52 over 500–700 ms, BL-1 Acid at 48 kHz, velocity 100) as a test first and watched it fail — `peak` over the 500–700 ms window was `0`, not merely small.

Fixed the way you describe under the documented next-note-takes-over behavior: `monoOrder` now measures a slide against the note that actually held the voice rather than the longest gate seen so far, so `previousGateOff = event.gateOffFrame` instead of `Math.max(...)`. That also makes the invariant provable rather than incidental — a slide now always starts while its predecessor is still gated, and `Adsr` only enters `release`/`done` once the gate goes false, so an inherited envelope can never be a finished one. Both facts are in the comment above `monoOrder`.

Verified: the new regression test passes, `packages/timeline` 961 passed, the two agent timeline suites 110 passed, `npm run lint` exit 0, and `harness gate --base main` 8/8 with `mappingViolations: []`.

---

PR #5739, opened 2026-09-11
feat(providers): add video upscale, interpolate and outpaint task types

georgi:
## What changed

Video had one catch-all transform task, `video_to_video`, so an upscaler, a frame interpolator and an outpainting model were indistinguishable in the catalog and unreachable as their own provider calls; images had `upscale_image` and `inpainting` capabilities but no outpainting at all. This adds four capabilities to the provider seam — `upscale_video`, `interpolate_video`, `outpaint_video` and `outpaint_image` — each with its own params type, `BaseProvider` method, capability detection and `dispatchCapability` case, plus an entry in the unit-billed capability set and the project spend categories. Models are assigned across every provider that ships suitable endpoints: fal (11 upscalers, 3 interpolators, 4 video outpainters, 2 image outpainters), kie (`topaz/video-upscale`), replicate (5 video upscalers, `bria/expand-image`), atlascloud (`atlascloud/video-upscaler`) and topaz (`topaz/video/interpolate`), with `upscaleVideo` / `interpolateVideo` / `outpaintVideo` / `outpaintImage` implemented on fal, kie, replicate and atlascloud.

Two tagging corrections fall out of it. A task keyword only claims a specialized task when the manifest entry declares a video input, so `grok-imagine/upscale` (which upscales a prior task by id) and fal's frame-list interpolation endpoint no longer surface as models that take a clip; and a LoRA trainer carries the same keywords while performing none of them, so it stays a plain source-consuming endpoint. Image outpainting endpoints become `outpaint`-only, keeping them out of the generation pickers the way `upscale` and `segment` already are. The `inpainting` tag is added rather than substituted: `buildImageModels` already derives it from a declared mask field, so the name check only catches endpoints whose manifest entry carries no mask field but whose id says what they are, and they stay valid answers to an `image_to_image` request.

Scope note: the agents' `SUPPORTED_CAPABILITIES` list for `find_model` is left alone — it already omits `upscale_image`, `inpainting`, `video_to_video` and `lip_sync`, so adding only these four would have been inconsistent with that surface's existing coverage.

## Merge with main (commit 2)

`main` moved under this branch and extracted `modelMatchesTask` / `STRICT_MODEL_TASKS` out of `useModelsByProvider.ts` into `hooks/modelTaskMatching.ts`, conflicting with the strict-task entries added here. Resolved by taking main's extraction and carrying the four new strict tasks into the new module. Widening `VideoModelTask` also broke `CLIP_TASK_LABELS`, an exhaustive `Record<VideoModelTask, string>` in `StoryboardBoard`; it is now keyed off `requiredVideoTasksForShots`'s own return type — the only thing that indexes it — which also drops the `video_to_video` and `lip_sync` entries that were already unreachable.

## Verification

Re-run in full on the merge commit (`681bae82`):

- [x] `npm run lint` — exit 0
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 8/8 selfchecks passed`, exit 0
- [x] `npm run typecheck` — the only 4 errors are `ProjectLifecycleActions.tsx` (3) and `projectStarters.test.ts` (1). **Both are pre-existing on `main`**: a checkout of pristine `a7ae40b8` produces the same 4 errors in the same 2 files, so this diff adds none. They are not this PR's to fix and no fix for them exists to port yet.
- [x] `npm run test:affected` — 107/115 tasks. The one failing task is `@nodetool-ai/deploy` `exports all public API`, which times out at **5055 ms against a 5000 ms limit** on a bare `await import("../src/index.js")` under parallel load. It passes 3/3 standalone, and `packages/deploy` references none of the code this PR touches.

Before the merge, the same four checks were clean on `b34f9d1a`: `test:affected` 115/115, `typecheck` 0 errors, `lint` 0 errors, gate 6/6.

One further pre-existing `main` failure worth naming: `web` `src/demo/timeline/__tests__/timelineChrome.test.tsx > shows the editor's actions, inert` fails on pristine `a7ae40b8` too (verified in a clean worktree), and arrived here with the merge.

Three container-level failures were fixed rather than waved through, since `AGENTS.md` documents each as an environment problem and not a broken test:

- 52 `image-nodes` failures, all `No WebGPU adapter available (Node/Dawn)`. After `apt-get install -y mesa-vulkan-drivers`: `Test Files 6 passed (6) / Tests 90 passed (90)`.
- `video-nodes` `timeline-time-remap-decode` hook timeout and `runtime` `video-frame-fallback` `spawn ffmpeg ENOENT`. After `apt-get install -y ffmpeg`, both pass.
- Two `cli` `run-dsl` tests timing out at 30 s under parallel load (50 s and 39 s). Run alone: `Test Files 1 passed (1) / Tests 10 passed (10)`.

A stale `@nodetool-ai/protocol` (rc.11) nested under `mobile/node_modules` also shadowed the workspace build and produced 36 mobile typecheck errors; the same 36 appeared on the unmodified base tree, and removing the nested copy cleared them.

## Agent capabilities

Not applicable — this adds provider capabilities on the `BaseProvider` seam, not agent/MCP capabilities. No capability's declared contract (name, description, input schema, permission category, `needsToolCallId`) changes.

## New checks

The diff adds `packages/runtime/tests/providers/edit-task-capabilities.test.ts` (11 cases) covering all three seams the new tasks cross: the capability a provider advertises, the `runProviderPrediction` dispatch, and the catalog tagging.

- [x] Two of the new checks were inverted once and observed failing, then the sources were restored:
  - Replacing the video-input filter with `|| true` in `buildVideoModels`:
    ```
    FAIL  tests/providers/edit-task-capabilities.test.ts > editing task model tagging >
      drops a transform task from a model that declares no video input
    Tests  1 failed | 10 passed (11)
    ```
  - Making `coerceOutpaintPadding` return its bag unconditionally:
    ```
    FAIL  tests/providers/edit-task-capabilities.test.ts > editing task dispatch >
      leaves padding unset when the caller names no side
    Tests  1 failed | 10 passed (11)
    ```
- [x] The existing `generation-seam-audit` is the audit that asserts it found its targets, and it caught a registry this diff had missed — `MEDIA_METHODS` did not know the four new media methods, which would have let a call to any of them slip past the seam unrecorded. It now lists them and passes.

Three pinned expectations changed on purpose, because the behaviour they pinned is what this PR replaces: `fal-ai/topaz/upscale/video` and its siblings now tag `upscale_video` instead of `video_to_video`, `.../outpaint` tags `outpaint_video`, and `.../frame-interpolation` tags `interpolate_video`.

https://claude.ai/code/session_01829riCEERjYpRSiswKuvxd


georgi:
**`Quality Gate (npm run check) / test-packages-agents` is failing, and it is not this PR's.**

7 tests fail in `packages/agents`, all with `Error: Database not initialized. Call initDb() or initPostgresDb() first.`:

- `tests/view-image-tool.test.ts` — 5 cases, via `viewImageImpl` → `findRunAsset` → `Asset.find`
- `tests/tools/asset-tools.test.ts` — 2 cases (`:222`, `:338`)

**Why it isn't this PR's.** The same 7 tests fail identically on pristine `main` (`a7ae40b8`), checked out in a clean worktree and run with the same command. This branch touches no file in `packages/agents` and nothing in the `Asset` / `db` path; its only `packages/models` edit adds four capability→category entries to `project-summary.ts`.

**Root cause.** `a7ae40b8` ("fix: enforce project workspace boundaries") introduced `findRunAsset` in `packages/agents/src/capabilities/assets.ts`:

```ts
async function findRunAsset(run, assetId): Promise<AssetRow | null> {
  const { Asset } = await import("@nodetool-ai/models");
  const asset = await Asset.find(userIdOf(run.context), assetId);
  return asset?.project_id === resolveProjectId(run, {}) ? asset : null;
}
```

Asset resolution now goes through the database and is filtered by `project_id`. That commit updated `tests/capabilities-assets.test.ts` for the new requirement (it added `beforeEach(() => { initTestDb(); })` and a `seedAsset` helper), but `view-image-tool.test.ts` and `tools/asset-tools.test.ts` were not updated, so they still run with no database.

**Proposed patch (not applied here — it belongs to `a7ae40b8`, not to this PR).** The two suites need the same treatment `capabilities-assets.test.ts` got. Adding only the DB init is not sufficient — I tried it:

```ts
import { initTestDb } from "@nodetool-ai/models";
beforeEach(() => { initTestDb(); });
```

That clears the `Database not initialized` error but the 6 cases then fail with `TypeError: Cannot read properties of undefined (reading 'uri')`, because `findRunAsset` also requires the asset row to exist with a `project_id` matching `resolveProjectId(run, {})`. So each case additionally needs its asset seeded, mirroring `seedAsset` in `capabilities-assets.test.ts`:

```ts
await new Asset({
  id: "abc123",
  user_id: USER,
  project_id: "default",
  name: "abc123.png",
  content_type: "image/png"
}).save();
```

I reverted that trial rather than widen this PR with a fix to another commit's tests. No open PR currently fixes it, so there is nothing to port in.

**Re-run.** `rerun_failed_jobs` on run `34655500137` returned `403 This workflow is already running`, since other Quality Gate jobs are still in progress. The one re-run is unspent; I will use it once the run settles and will keep watching this PR until it is green and merged.

For completeness, two other pre-existing `main` failures reach this branch through the merge and are noted in the PR description: 4 web typecheck errors in `ProjectLifecycleActions.tsx` / `projectStarters.test.ts`, and `web/src/demo/timeline/__tests__/timelineChrome.test.tsx > shows the editor's actions, inert`. Both reproduce on untouched `a7ae40b8`.

---

georgi:
**Two more Quality Gate checks failed. Both reproduce on pristine `main`; neither is this PR's.**

Same method as the `test-packages-agents` comment above: run the failing suites at `a7ae40b8` in a clean worktree and compare.

### `Quality Gate (npm run check) / test-packages-websocket` — 5 failures, 2 files

| Suite | Failures | Cause |
|---|---|---|
| `tests/mcp-view-image.test.ts` | 4 | Same `findRunAsset` change from `a7ae40b8` as the agents failures — asset resolution is now DB-backed and `project_id`-filtered, so `view_image` returns `isError: true` and no `image` block. |
| `tests/sandbox-api-coverage.test.ts` | 1 | `a7ae40b8` added `restoreTabs: protectedProcedure` to `packages/websocket/src/trpc/routers/projects.ts` (line 182 of that commit's diff) without classifying it in `src/trpc/sandbox-coverage.ts`. The audit fails with `expected [ 'projects.restoreTabs' ] to deeply equal []`. |

Verified: both files produce the **identical 5 failures** on untouched `a7ae40b8`. This branch's commit touches no file under `packages/websocket` or `packages/agents` — `git diff b34f9d1a~1 b34f9d1a --stat -- packages/websocket packages/agents` is empty.

The second one is self-describing: the audit exists precisely to catch a new procedure that nobody classified, and it is doing its job. The fix is one entry in `src/trpc/sandbox-coverage.ts` saying which capability covers `projects.restoreTabs`, or why sandboxed code may not have it. That belongs with the commit that added the procedure.

### `Quality Gate (npm run check) / test-app` — 1 failure

`src/demo/timeline/__tests__/timelineChrome.test.tsx:52 > shows the editor's actions, inert` — `getByRole("button", { name: /project settings/i })` finds nothing. 1 failed, 16003 passed.

This is the failure already named in the PR description: it fails on pristine `a7ae40b8` too (verified in a clean worktree before this branch was merged with main). `a7ae40b8` reworked `WorkspaceShell.tsx` and `ProjectSelector.tsx`, which is where that toolbar's buttons come from.

### Status

No open PR fixes any of these, so there is nothing to port in. I am not absorbing fixes for another commit's tests into this PR.

Every check that has failed on this head so far is one of these three, all traced to `a7ae40b8`. This branch's own four checks pass: `lint` exit 0, `harness gate` 8/8, `typecheck` clean apart from main's 4 pre-existing errors, and `test:affected` green apart from one 5055 ms-vs-5000 ms import timeout under parallel load that passes 3/3 standalone.

The single re-run remains unspent — `rerun_failed_jobs` on run `34655500137` still returns `403 This workflow is already running`. I will spend it once the run settles and keep watching until this PR is green and merged.

---

PR #5742, opened 2026-09-12
fix: give typecheck an explicit tsc heap ceiling

georgi:
## What changed

`npm run typecheck` let every `tsc` inherit Node's machine-derived default heap. That default is **2096 MB** on a 16 GiB host (smaller under a container memory limit), while `web`'s program needs **~4.5 GiB** to check — 10,646 files once MUI/Emotion structural assignability is counted. So a bare `tsc --noEmit` aborted with `JavaScript heap out of memory` (exit 134) instead of reporting type errors, and the failure read as a compiler bug rather than a missing flag. Anyone whose shell happened to carry a `--max-old-space-size` in `NODE_OPTIONS` never saw it, which is why it looked intermittent.

This routes every `tsc` in the root typecheck chain — `web`, `electron`, `mobile`, and the `packages/protocol` build inside `typecheck:mobile` — through a new `scripts/run-tsc.mjs`. The runner applies `typeScriptBuildEnv`, the same `NODETOOL_TSC_HEAP_MB` policy `npm run build:packages` already uses, so the repo keeps one heap knob and a `--max-old-space-size` the caller set in `NODE_OPTIONS` still wins. It resolves the nearest `node_modules/typescript/bin/tsc` rather than the root one, because `mobile` has a separate dependency tree pinned to TypeScript 6.0.3 while the root is on 5.9.3 — matching what npm's own bin resolution did before.

Profiling found no safe way to shrink the program itself, so the ceiling is the fix rather than a workaround. Check time is diffuse (the hottest file, `ThemeNodetool.tsx`, is 5s of 116s; `structuredTypeRelatedTo` dominates at 207s cumulative), and resolving `@nodetool-ai/*` through `dist` `.d.ts` instead of the `nodetool-dev` source condition saves only 11% (4609 MB → 4122 MB) while giving up typecheck fidelity against live package source. `incremental: true` was also measured: warm runs drop to 13s / 2072 MB, but cold peak *rises* to 5290 MB, and 2072 MB still sits on top of the 2096 MB default — so it does not remove the need for an explicit ceiling.

Letting the check run to completion immediately exposed 2 real type errors that the abort had been hiding (`documentSync.test.ts`, arrived on `main` via #5741). Fixed in `e3e12bfb` — see **Fallout** below.

## Verification

Measured with `--extendedDiagnostics` in `web/`, after `npm run build:packages`.

The reproduction, with the machine's default heap:

```
$ cd web && env -u NODE_OPTIONS node ../node_modules/typescript/bin/tsc --noEmit
 8: 0x1414ade v8::internal::Factory::NewFillerObject(...) [node]
 9: 0x18772bc v8::internal::Runtime_AllocateInYoungGeneration(...) [node]
exit=134
```

The same environment, through the runner:

```
$ env -u NODE_OPTIONS node ../scripts/run-tsc.mjs --noEmit
exit=0
```

- [x] `npm run test:affected` — the root `package.json` change selects every suite (`packages`, `web`, `electron`, `mobile`), which is CI's job. Ran the slices this diff can affect: `npm run test:scripts` (26 passed, 6 files), `packages/websocket/tests/build-typescript-workspace.test.ts` (6 passed, covers the `typeScriptBuildEnv` heap policy this reuses), and `web/src/stores/__tests__/documentSync.test.ts` (11 passed, the file changed here).
- [x] `npm run lint` — exit 0; only pre-existing `no-unused-vars` warnings on unrelated files.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 1/1 selfchecks passed`.
- [ ] `npm run typecheck` — `web` and `electron` both pass (exit 0). The chain still exits 2 on `typecheck:mobile`, for a pre-existing reason that is not this PR's and not heap-related: see **Known blocker**.

## Fallout: 2 type errors the abort was hiding

`documentSync.test.ts` declares `resolveSave: ((value: { updatedAt: string }) => void) | null = null` and assigns it only inside the promise executor. TypeScript narrows it back to its `null` initializer at both call sites, so `resolveSave?.({...})` typed the callee `never` and failed TS2349. The executor runs synchronously inside `save()`, and the `expect(save).toHaveBeenCalledTimes(1)` above each call proves it ran, so this switches to a definite-assignment declaration and a direct call. The `?.` was not merely redundant — it turned a never-assigned resolver into a silent no-op instead of a failed test.

## Known blocker (not this PR's)

`typecheck:mobile` reports ~36 errors: missing `@nodetool-ai/protocol` exports (`Shot`, `ShotStatus`, `CameraDirection`, `TRPC_MAX_BATCH_SIZE`, `processingMessageSchemas`) and field drift (`sync_mode`, `expose_as_tool`, `supports_dynamic_inputs`, `application_id`).

`mobile/package-lock.json` pins `@nodetool-ai/protocol` to the published registry tarball `0.7.0-rc.11` while the workspace is at `0.7.0-rc.40`, and `mobile/tsconfig.json` maps only the `/triggers` subpath to workspace source — so the bare specifier resolves to rc.11 through `mobile/node_modules`. CI's `cd mobile && npm ci` installs the same pinned tarball, so it reproduces there. This diff touches no mobile or protocol source, and `tsc --build packages/protocol` never fed mobile's types, so these errors are identical on `main` — previously hidden behind the heap abort.

Fixing it means changing how `mobile` consumes workspace packages (a `paths` mapping for the bare specifier, or a `file:../packages/protocol` dependency), which per `AGENTS.md` must stay in agreement across `mobile/metro.config.js`, `tsconfig.json`, and `jest.config.js`. That belongs in its own PR rather than widening this one.

## New checks

`scripts/__tests__/run-tsc.test.mjs` drives the runner against a stub `tsc` that reports the heap it was actually given, so it cannot pass on a runner that never hands the flag to the child. It covers the default ceiling, `NODETOOL_TSC_HEAP_MB`, deferring to an explicit `NODE_OPTIONS` flag, exit-code propagation, and the nearest-`node_modules` resolution.

- [x] The check was inverted once and observed failing. Replacing `env: typeScriptBuildEnv()` with `env: process.env` in the runner:

```
 × raises the child heap to the default ceiling and forwards args
 × honours NODETOOL_TSC_HEAP_MB
 Tests  2 failed | 6 passed (8)
```

  Restored, then re-run: `Tests 8 passed (8)`.
- [x] The resolution tests assert a concrete found path rather than only absence — one pins the real TypeScript this repo builds with, and a separate case asserts `null` when nothing exists, so the walk-up cannot pass by matching nothing.

## Agent capabilities

No capability added and no declared contract changed.

https://claude.ai/code/session_011LrV2XcGTFGjMnkH4YQzXt


georgi:
Merged `main` in (`ef3f264a`) and re-verified. Two updates, one of which is a blocker that is not this PR's.

**`web` typecheck now passes clean** — exit 0 under a default-heap environment, which is the first time this chain has completed rather than aborting. The 4 errors I listed in the PR body were fixed on `main` in the meantime; the merge brought in 2 new ones (`documentSync.test.ts` TS2349, from #5741), fixed in `e3e12bfb`. `resolveSave` is declared `(...) | null = null` and only assigned inside the promise executor, so TypeScript narrows it to `null` at both call sites and `resolveSave?.(...)` types the callee `never`. The executor runs synchronously inside `save()` and the `expect(save).toHaveBeenCalledTimes(1)` above each call proves it ran, so it now uses a definite-assignment declaration and a direct call. The `?.` was also masking a real failure mode: a never-assigned resolver was a silent no-op rather than a failed test. All 11 tests in that file pass.

**`mobile` typecheck is red, and it is not this PR's failure.** `npm run typecheck` still exits 2 on the `typecheck:mobile` leg with ~36 errors — missing `@nodetool-ai/protocol` exports (`Shot`, `ShotStatus`, `CameraDirection`, `TRPC_MAX_BATCH_SIZE`, `processingMessageSchemas`) and field drift (`sync_mode`, `expose_as_tool`, `supports_dynamic_inputs`, `application_id`).

The cause is dependency wiring, not heap and not this diff:

- `mobile/package-lock.json` pins `@nodetool-ai/protocol` to the **published registry tarball `0.7.0-rc.11`** (`resolved: https://registry.npmjs.org/@nodetool-ai/protocol/-/protocol-0.7.0-rc.11.tgz`), while the workspace is at `0.7.0-rc.40`.
- `mobile/tsconfig.json` maps only the `@nodetool-ai/protocol/triggers` subpath to workspace source. The bare specifier resolves through `mobile/node_modules`, so it gets rc.11.
- CI runs `cd mobile && npm ci`, which installs that same pinned tarball, so this reproduces there.

This diff touches no mobile or protocol source — only the `typecheck` script strings — and `tsc --build packages/protocol` never fed mobile's types either way, since mobile does not resolve protocol from the workspace. So these errors are identical on `main`, and they were previously hidden behind the heap abort this PR removes.

I am not fixing it here, because the fix is an architectural change to how `mobile` consumes workspace packages (either a `paths` mapping for the bare specifier or a `file:../packages/protocol` dependency), and per `AGENTS.md` that mapping has to stay in agreement across `mobile/metro.config.js`, `tsconfig.json`, and `jest.config.js`. That is a separate PR, and widening this one would bury the heap fix. I have raised it with the author for a decision.

---

georgi:
All 26 checks on `e3e12bfb` report success and `mergeable_state` is `clean`, but that green is not real, and the reason explains why the heap abort survived so long.

**The heap fix is confirmed working in CI.** From the `typecheck` leg's log, `typecheck:web` ran through the runner and passed in 101s (08:26:11 → 08:27:52), and `typecheck:electron` passed. That is the check completing rather than aborting.

**`typecheck:mobile` failed in that same job, and the job still reported success.** The log carries ~36 `##[error]` annotations — the exact `sync_mode` / `supports_dynamic_inputs` / `expose_as_tool` / `TRPC_MAX_BATCH_SIZE` / `Shot` / `ShotStatus` / `CameraDirection` / `Screenplay` / `processingMessageSchemas` / `application_id` errors described above, confirming that diagnosis in CI. Then `##[endgroup]` printed and the job concluded green.

The cause is the matrix Run step:

```yaml
run: |
  echo "::group::${{ matrix.check }}"
  ${{ matrix.command }}
  echo "::endgroup::"
```

GitHub's default shell is `bash -e`, and POSIX `-e` is ignored for a command in an `&&` list that is not the last one. So when `npm run typecheck` — first in a nine-command `&&` chain — fails, the shell does not exit; it falls through to `echo "::endgroup::"`, which succeeds and becomes the script's exit status. Reproduced directly:

```
$ cat t.sh
echo "::group::typecheck"
false && echo "second" && echo "third"
echo "::endgroup::"

$ bash -e t.sh; echo "EXIT=$?"
::group::typecheck
::endgroup::
EXIT=0
```

Note that `set -eo pipefail` does **not** fix this — the problem is the trailing statement, not the pipe. This is the same class of bug already documented in `quality-guard.yml`, in a different shape.

**Blast radius.** Three legs chain commands with `&&`, so every command but the last cannot fail the build:

| Leg | Chained | Actually gating |
|---|---|---|
| `typecheck` | 9 | only `capabilities:check` |
| `deps` | 6 | only `check-execution-boundary.mjs` |
| `test-packages-core` | 3 | only the `turbo run test` |

That is 18 checks of which 15 are vacuous, including `npm run typecheck` itself, `validate:examples`, `codegen:dsl:check`, `typecheck:examples`, `check-workspace-deps`, `check-circular-deps`, `check-key-boundaries`, `check-lockfile-sync`, `test:oxlint-rules` and `test:scripts`. `lint`, `test-app`, `bundle`, the single-`turbo` shards and `harness-gate` (its `if` is the final statement) propagate correctly and are unaffected.

**A verified fix**, keeping the log grouping:

```yaml
run: |
  echo "::group::${{ matrix.check }}"
  trap 'echo "::endgroup::"' EXIT
  ${{ matrix.command }}
```

Making the command the script's last statement means its status becomes the script's status, and the trap still closes the group. Checked in all three shapes:

```
failing && chain   → EXIT=1   (endgroup printed)
passing && chain   → EXIT=0   (endgroup printed)
single failing cmd → EXIT=1   (endgroup printed)
```

**I have not pushed this.** Applying it turns the gate red immediately — `typecheck:mobile` fails today, and the other 14 unmasked checks are unaudited — so the first PR to carry it would red-wall the repo until the mobile protocol drift is fixed. That sequencing is the author's call, not something to bundle into a heap fix. Raised with the author for a decision; the sane order is mobile drift first, then the Run step, then this can go in on a genuinely green gate.

---

PR #5755, opened 2026-09-13
fix(runtime): report OpenRouter errors that arrive on a 200

georgi:
## What changed

OpenRouter commits an HTTP 200 before its upstream provider runs, so any failure after that point rides the response body instead of the status line — its [API reference](https://openrouter.ai/docs/api-reference/errors) says to check the body for an `error` field even on a 200. The OpenAI-compat client checked only the status, so such a body fell through to `decodeChatCompletion`, which found no `choices` and threw `openrouter returned no choices` — a diagnosis that named neither the endpoint nor the reason. This landed on `view_image` and `understand_video` specifically because they are the two tools that put pixels in the request (`understand_video` via ffmpeg frame sampling, since OpenRouter's wire format carries only `text` and `image_url` content parts), so an image-routing refusal such as `No endpoints found that support image input` is raised for them and for nothing else. `errorFromOkBody` now inspects a 200 completion body for a top-level `error` and for one inside the first choice (`finish_reason: "error"`), the streaming path reads that same choice-level shape, and a numeric `error.code` in the 100–599 range becomes the error's status so `isRateLimitError` still classifies a 429 that never reached the status line.

## Verification

Inverted once: with the three source files reverted and the new tests kept, all four new cases failed, including `names the upstream reason when an image request fails on a 200` (which asserts `not.toContain("returned no choices")`):

```
× throws the upstream reason when a 200 body carries an error
× throws when a 200 body reports the error inside its first choice
× throws on a mid-stream choice error with no top-level error field
× names the upstream reason when an image request fails on a 200
Tests  4 failed | 40 passed (44)
```

- [x] `npm run test:affected` — `122/127 tasks successful`. Two failures, both reproduced identically on the base commit with the diff stashed: `image-nodes` (no Vulkan ICD; passes 266/266 after `apt-get install mesa-vulkan-drivers` with `VK_DRIVER_FILES=/usr/share/vulkan/icd.d/lvp_icd.json`) and `base-nodes > directed-campaign-kit.test.ts > rejects a wrong reference role and missing product before a paid call` (pre-existing). `packages/runtime` on its own: 2222 passed, 13 skipped, 0 failed after installing ffmpeg.
- [x] `npm run typecheck` — clean for every root workspace. The only errors are `mobile/` missing-module errors from its separate, uninstalled dependency tree; `tsc --noEmit` in `packages/runtime` exits 0.
- [x] `npm run lint` — 1 error, `packages/atlascloud-nodes/src/atlascloud-factory.ts` `anti-slop(no-runtime-typeof)`, identical with the diff stashed. This diff adds none.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 17/17 selfchecks passed`.

## Agent capabilities

No capability is added and no declared contract changes.

## New checks

- [x] The check was inverted once and observed failing; the failing output is in **Verification** above
- [x] Not an audit — these are unit tests over fixed response bodies, and a case asserting an ordinary completion carrying `error: null` still decodes guards against the check firing on everything

https://claude.ai/code/session_01L7JxkuDVdiHZhnvam2h4Rw

---
_Generated by [Claude Code](https://claude.ai/code/session_01L7JxkuDVdiHZhnvam2h4Rw)_


georgi:
`Quality Gate (npm run check) / lint` is failing on `a6fdfcd`, and the failure is not this PR's.

The one error is in a file this diff does not touch:

```
packages/atlascloud-nodes/src/atlascloud-factory.ts:713:15: error anti-slop(no-runtime-typeof)
Found 75 warnings and 1 error.
```

`main` is red for the same reason. I checked out the base commit `cf4ea43` with this branch's diff absent and ran `npm run lint` there: identical single error, exit 1. That is direct evidence for what a job re-run would only infer, so I have not spent a re-run on a deterministic lint failure. `git log -L 707,720` dates the offending lines to `e69b9b5` ("feat(atlascloud): add latest image and video models"), which landed a `typeof value === "object"` narrowing into a tree that `.oxlintrc.anti-slop-enforced.json` lists under the enforced `anti-slop/no-runtime-typeof` override. No revert or fix PR exists, so there is nothing to port.

Rather than widen this PR into an unrelated package, here is the patch, verified locally — `npm run lint` exits 0 repo-wide, `tsc --noEmit` in `packages/atlascloud-nodes` exits 0, and its 86 tests pass. It follows the file's own convention (`isStringValue`/`isNumberValue` take `NodeValue`, not `unknown`, so consolidating into a predicate does not just move the finding to `no-unknown-parameters`):

```diff
@@ -165,6 +165,21 @@ function isStringValue(value: NodeValue): value is string {
   return typeof value === "string";
 }
 
+/** Narrow a node property value to an image or video reference. */
+function isVisualReference(
+  value: NodeValue
+): value is { [key: string]: NodeValue; type: "image" | "video" } {
+  if (
+    typeof value !== "object" ||
+    value === null ||
+    Array.isArray(value) ||
+    value instanceof Uint8Array
+  ) {
+    return false;
+  }
+  return value.type === "image" || value.type === "video";
+}
+
 /** Narrow a node property value to a number. */
 function isNumberValue(value: NodeValue): value is number {
   return typeof value === "number";
@@ -707,14 +722,7 @@ export function createAtlasNodeClass(spec: AtlasManifestEntry): NodeClass {
       for (const group of requiredGroups) {
         const values = input[group];
         const hasVisualReference =
-          Array.isArray(values) &&
-          values.some(
-            (value) =>
-              typeof value === "object" &&
-              value !== null &&
-              "type" in value &&
-              (value.type === "image" || value.type === "video")
-          );
+          Array.isArray(values) && values.some(isVisualReference);
         if (!hasVisualReference) {
           throw new Error(
             `${specRef.title}: connect at least one reference image or video`
```

The `Uint8Array` and `Array.isArray` guards are load-bearing: `NodeValue` admits both, and without them `tsc` rejects the `value.type` access.

Say the word and I will carry this into this PR, or it can land on its own to unblock `main` for every open PR. This PR's own checks are otherwise clean, and I am keeping it watched until it is green.

---

georgi:
`Run workflow suite in browser` is also failing on `a6fdfcd`, and this one is not this PR's either. `main` is red for it.

Five workflows fail, all from the directed-campaign kit:

```
compose-directed-campaign-formats: failed (Node "layout-elements" failed: The accepted creative contract is invalid.; Node "layout-finalize" failed: The accepted creative contract is invalid.)
propose-three-campaign-directions: failed (Node "direction-brief" failed: product name is required.)
render-a-directed-campaign-hero: failed (Node "hero-prepare" failed: The campaign plan is invalid. Propose directions again.)
reopen-a-directed-campaign: failed (Node "validate-record" failed: media.info: could not read a document ref with no uri, asset_id, or data)
revise-an-accepted-campaign-hero: failed (Node "revision-prepare" failed: media.bytes: could not read a image ref with no uri, asset_id, or data)
```

I ran `npm run test:e2e-runner` from `web/` on the base commit `cf4ea43` with this branch's diff absent: the same five workflows fail with byte-identical messages. Again direct evidence rather than a re-run, so no re-run spent.

These are node-level contract failures inside the campaign nodes (a missing product name, an invalid plan, media refs carrying no `uri`/`asset_id`/`data`). This PR changes one thing — how `packages/runtime/src/providers/openai-compat/` reads an error out of an HTTP 200 chat-completion body — and none of these workflows makes a chat-completion call, so there is no path from the diff to these failures.

The same regression has a unit-level twin that is also red on `main`: `packages/base-nodes` › `tests/directed-campaign-kit.test.ts` › `rejects a wrong reference role and missing product before a paid call`. The likely origin is the directed-campaign work in `f918112`, `ac89c3b` and `ca79f7c`; I have not root-caused it, since it is a different subsystem from this PR and fixing it here would widen the PR well past its subject.

So both red checks on this PR — this one and `Quality Gate (npm run check) / lint`, covered in my previous comment — are pre-existing `main` breakage with no fix PR to port. Nothing about this PR's own subject is failing: `packages/runtime` passes 2222 tests, `tsc --noEmit` is clean there, and `harness gate --base origin/main` reports 17/17. I am keeping this PR watched until it is green.

---

georgi:
Third red check, `Quality Gate (npm run check) / test-packages-nodes`, and this one is a different kind from the other two.

```
@nodetool-ai/game-nodes › tests/game-nodes.test.ts:316
  › ExportGodotProject › writes the golden resources byte for byte
Error: Test timed out in 5000ms.
Tests  1 failed | 44 passed (45)
```

A timeout, not an assertion. It passes here on this exact commit `a6fdfcd`: `npm run test --workspace=packages/game-nodes` gives 45/45, and the test itself runs in **309 ms** against vitest's default 5000 ms ceiling — a 16x margin. The package sets no `testTimeout`, so the default applies.

The same job log shows the runner was heavily contended while this ran: `[vitest-pool]: Timeout terminating forks worker for .../image-nodes/tests/lib-image-processing.test.ts`, `Slow sandbox worker round trip {"workerRoundTripMs":1645}` in code-nodes, and `extracts red channel` taking 14 s. A 309 ms test blowing a 5 s budget under that load is the runner, not the diff — and this PR changes only how `packages/runtime/src/providers/openai-compat/` reads an error out of an HTTP 200 chat body, which a Godot project export never touches.

This is exactly the case a re-run is for, so I have spent my one re-run on it (`rerun_failed_jobs` on run `34747801796`). If it fails a second time I will treat it as real and root-cause it rather than call it a flake again. The lint job in that same run will stay red — it is the pre-existing `main` breakage from my earlier comment, and a re-run cannot change a deterministic lint error.

Status of the three red checks on this PR:

| Check | Cause | Mine? |
|---|---|---|
| `Quality Gate / lint` | `atlascloud-factory.ts` `no-runtime-typeof`, from `e69b9b5` | No — red on base `cf4ea43`, patch proposed above |
| `Run workflow suite in browser` | directed-campaign kit, 5 workflows | No — reproduced byte-identical on base `cf4ea43` |
| `Quality Gate / test-packages-nodes` | game-nodes 5 s timeout under runner load | No — 45/45 locally on this commit, 309 ms vs 5000 ms; re-run spent |

---

georgi:
Correcting the record on `test-packages-nodes`, and a root cause for the directed-campaign failures.

**The game-nodes timeout was a flake, and the re-run confirms it.** On attempt 2 `@nodetool-ai/game-nodes` passed; 73 of 75 tasks succeeded. So the "treat it as real if it fails twice" commitment from my last comment does not trigger — it did not fail twice.

**Attempt 2 failed on a different package**, `@nodetool-ai/base-nodes`, with two failures in `tests/directed-campaign-kit.test.ts` — the file I already flagged as red on `main`:

1. `rejects a wrong reference role and missing product before a paid call` — `ExecutionPreflightError: Provider "openai" needs "OPENAI_API_KEY"`.
2. `executes deterministic layouts, escapes exact copy, preserves original artifacts, and restores the portable record` — `Test timed out in 30000ms`. It passes here in **7.8 s**, so this is the same runner-contention family as the game-nodes one.

**Root cause of the first one, verified.** The test builds its context with a fake OpenAI provider:

```ts
const fake = createFakeContext({ providers: { openai: provider }, jobId: … });
const session = await ExecutionSession.create({ …, context: fake.context, … });
```

but `ExecutionSession.create` calls `assertPreflight`, whose credential check reads *server* credentials rather than the providers injected on the context. So the fake is ignored and the run is refused for a missing `OPENAI_API_KEY` — before the graph can produce the reference-role/missing-product error the test actually asserts.

The decisive check: with a throwaway value in the environment the whole file passes.

```
$ npx vitest run tests/directed-campaign-kit.test.ts
Tests  1 failed | 7 passed (8)

$ OPENAI_API_KEY=sk-test-not-a-real-key npx vitest run tests/directed-campaign-kit.test.ts
✓ rejects a wrong reference role and missing product before a paid call 164ms
Tests  8 passed (8)
```

That is why it is green wherever a developer has a key in the environment and red on CI. It has nothing to do with this PR, and it reproduces on base commit `cf4ea43` with this branch's diff absent.

`PreflightOptions` already carries a `providerConfiguration?: ProviderConfigurationChecker` seam, so the honest fix is for a run whose context supplies a provider to pass a checker that accounts for it, rather than for the test to depend on ambient environment. I have not pushed that: it is in `packages/execution` / `packages/base-nodes`, and carrying it here would widen a PR about OpenRouter error decoding into an unrelated subsystem. Happy to do it in its own change.

Standing summary for this PR, all verified rather than assumed:

| Check | Cause | Mine? |
|---|---|---|
| `Quality Gate / lint` | `atlascloud-factory.ts` `no-runtime-typeof` from `e69b9b5` | No — red on base; verified patch proposed above |
| `Run workflow suite in browser` | directed-campaign kit, 5 workflows | No — reproduced byte-identical on base |
| `Quality Gate / test-packages-nodes` | base-nodes preflight needs an ambient `OPENAI_API_KEY`, plus one 30 s timeout under load | No — root cause above; passes with any key set |
| `Quality Gate / quality` | aggregator over the legs above | No independent failure |

Everything this PR could plausibly affect is green on CI: `typecheck`, `harness-gate`, `bundle`, `build`, `docker`, `test-packages-core`, `test-packages-agents`, `test-packages-websocket`, `test-app`, `Workflow Integration Tests`, `Workflow Runner Browser E2E`. Keeping it watched.

---

PR #5760, opened 2026-09-13
feat(atlascloud): accept signed prediction callbacks

georgi:
## What changed

AtlasCloud can POST a prediction's terminal state back instead of making the run poll for it ([AtlasCloud webhooks](https://atlascloud.ai/docs/en/webhooks)), but nothing sent a `webhook_url` and nothing served the callback, so every run polled the prediction endpoint every 3s until it finished. This adds the ingress side the way the fal callback works and the wait side the way KIE's does. A signed route at `/api/providers/atlascloud/webhook` verifies Ed25519 over `<timestamp>.<raw body>` against the provider's JWKS (cached for a day, refreshed once on an unknown key id, deliveries older than five minutes rejected), then hands the prediction to an in-process registry the submitting run is waiting on. `atlasSubmit` sends `webhook_url` when `NODETOOL_PUBLIC_URL` resolves to a public https host, which is the only address AtlasCloud accepts — an http, loopback, or private value is ignored rather than sent to be refused.

Two design points. **The callback carries no durable state.** fal's callbacks commit to `GenerationWebhookDelivery` because fal runs through the durable generation lifecycle, which is fal-only by construction (`createFalGenerationLifecycleHooks` returns early for any other provider, and the recovery worker decodes fal outputs). AtlasCloud has no durable attempts, so a durable inbox for it would be rows nothing reads. **The callback never replaces the prediction endpoint.** Delivery is at-least-once and best effort with a ~30-minute safety net, so `atlasAwaitResult` races the callback against a reconciliation poll every 15s and aborts the loser. A callback that never arrives costs latency, not the run, and one this process cannot match (a duplicate, or a prediction another replica submitted) is acknowledged with `matched: false` so AtlasCloud stops retrying it.

Callers moved from `atlasPoll` to `atlasAwaitResult`: `AtlasCloudProvider.runJob` and the node-pack factory. `atlasPoll` itself gained an abort check at the top of its loop, because `sleep` resolves rather than throws on abort and the settled race would otherwise burn the request's whole retry budget on doomed fetches.

## Verification

All four required checks pass on the current head (`64f80c39`), and all 26 CI checks are green, `Quality Gate / quality`, `harness-gate`, `typecheck`, `lint`, and CodeQL included.

- [x] `npm run test:affected` — 112/112 tasks, exit 0
- [x] `npm run typecheck`
- [x] `npm run lint` — exit 0
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — selects the new `atlascloud-callbacks` surface

New suites:

```
npm run test --workspace=packages/websocket -- atlascloud-webhook   # 20 passed
npm run test --workspace=packages/runtime   -- atlascloud-webhook   # 23 passed
npm run test --workspace=packages/runtime   -- url-egress           # 11 passed
npm run test --workspace=packages/cli       -- harness              # 198 passed
```

Two failures during the first local run were environment gaps rather than defects, and I fixed the environment rather than skipping anything, per [dev-environment.md](docs/dev-environment.md#webgpu-on-a-headless-machine). A missing Vulkan ICD failed 52 `image-nodes` tests with `No WebGPU adapter available (Node/Dawn)`; after `apt-get install mesa-vulkan-drivers`, `image-nodes` passes 26/26 files and 266/266 tests. A missing ffmpeg failed `tests/providers/video-frame-fallback.test.ts` to load with `spawn ffmpeg ENOENT`; after installing ffmpeg it passes 13/13.

The negative cases are what prove the verifier does something. Each of these is a test in `packages/websocket/tests/atlascloud-webhook.test.ts` asserting a rejection, not a happy path: a body altered after signing (`invalid_signature`), a signature from a key the JWKS does not publish (`invalid_signature`), a timestamp 301s old (`stale_timestamp`), each of the four required headers removed in turn (`missing_header`), a `session_id` that is not the signed webhook id, a status AtlasCloud does not define, a signed body that is not JSON, and a body over the route's limit (413, with the sink never called). An unreadable JWKS answers 503 rather than 400, so AtlasCloud retries instead of giving up.

One test exists purely to catch drift: the route's `ATLASCLOUD_WEBHOOK_PATH` is asserted equal to the runtime's `ATLAS_WEBHOOK_PATH`, which is what the submitted `webhook_url` is built from. A mismatch between them would be a 404 only AtlasCloud ever sees.

## Agent capabilities

## New checks

The diff adds no repo-wide check, rule, or audit. It adds two registry entries that existing audits verify: a `fixedHost` row in the URL egress inventory for the JWKS fetch (`packages/runtime/tests/url-egress-audit.test.ts` fails on an unclassified `fetch` in the file), and an `atlascloud-callbacks` harness plus surface in `packages/cli/src/harness/registry.ts`. The registry test caught my first docs pointer as unresolvable and I fixed the pointer rather than the test:

```
FAIL tests/harness-registry.test.ts > docs pointers resolve
AssertionError: atlascloud-callbacks: no heading in docs/configuration.md starts with "NODETOOL_PUBLIC_URL"
```

It also adds one performance regression test, for the CodeQL fix in `64f80c39`. `replace(/\/+$/, "")` on the callback base is quadratic in the length of a slash run that is not at the end (1.0s at 50k slashes, 14.8s at 200k). `withoutTrailingSlashes` walks it off in one pass, and the test asserts a 50k run resolves in under 250ms. I inverted the fix to confirm the test catches a regression:

```
× walks a long slash run in linear time 812ms
AssertionError: expected 808.8577700000001 to be less than 250
```

That inversion mattered: my first attempt at the test used a base that *ends* in slashes, which is the fast path (0.1ms) and passed against the defect. The pathological shape is a long run followed by a non-slash.

https://claude.ai/code/session_01Rqpjq8je6UT6tV3bwYKkSZ


packages/runtime/src/providers/atlascloud-transport.ts
```diff
@@ -85,17 +95,47 @@
...
+ * AtlasCloud accepts only a public https URL, so a `NODETOOL_PUBLIC_URL` that is
+ * http, loopback, or an RFC1918 address resolves to undefined here: the run
+ * polls instead of sending a submission AtlasCloud would refuse.
+ */
+export function atlasWebhookUrl(
+  env: Record<string, string | undefined> = process.env
+): string | undefined {
+  const base = env["NODETOOL_PUBLIC_URL"]?.trim();
+  if (!base) return undefined;
+  const url = `${base.replace(/\/+$/u, "")}${ATLAS_WEBHOOK_PATH}`;
```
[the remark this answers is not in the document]
georgi: Fixed in 64f80c39.

The alert is correct. `replace(/\/+$/, "")` retries the whole slash run from every start position when the run is not at the end, so it is quadratic in the run's length. Measured on `https://nodetool.example.com` + n slashes + `x`:

```
 50000 slashes then char: 1004.9ms
100000 slashes then char: 3743.4ms
200000 slashes then char: 14770.6ms
```

A base that genuinely ends in slashes is the fast path (0.1-0.3ms at the same sizes), which is why my first attempt at a regression test passed against the defect. The pathological shape is a long run followed by a non-slash.

Scope: the value is `NODETOOL_PUBLIC_URL`, set by the operator rather than supplied on a request, so this is not reachable from the callback surface. It is still a defect in a URL builder and the linear form is three lines, so it is fixed rather than dismissed.

`withoutTrailingSlashes` now walks the run off in one pass. `packages/runtime/tests/providers/atlascloud-webhook-transport.test.ts` asserts a 50k-slash run resolves in under 250ms; I inverted the fix to confirm the test catches a regression:

```
× walks a long slash run in linear time 812ms
AssertionError: expected 808.8577700000001 to be less than 250
```

---

packages/runtime/src/providers/atlascloud-transport.ts
```diff
@@ -178,6 +237,72 @@ export async function atlasPoll(
...
+  opts.signal?.addEventListener("abort", onAbort, { once: true });
+
+  const callback = registerAtlasWebhookWait(
+    predictionId,
+    windowMs,
+    settled.signal
+  );
+  const reconcile = atlasPoll(apiKey, predictionId, {
+    pollInterval: reconcileInterval,
+    maxAttempts: Math.max(1, Math.ceil(windowMs / reconcileInterval)),
```
georgi: [P2] F1 (Correctness): do not let reconciliation exhaust the wait early

With callbacks enabled, `AtlasCloudProvider.runJob` converts a 10-second timeout and a 3-second poll interval into `maxAttempts: 4`, so this becomes `ceil(12000 / 15000) === 1`. `atlasPoll` makes its first request immediately and does not sleep after its last attempt. If that response is `processing`, it throws `job timed out` immediately, wins `Promise.race`, and removes the callback waiter despite nearly the whole 12-second window remaining. An isolated reproduction made one HTTP request and then rejected; a success callback delivered 20 ms later could no longer match. The current test resolves the callback synchronously, before that first poll can settle, so it misses this case.

Keep the overall timeout deadline-based, prevent the reconciliation attempt budget from rejecting before that deadline, and retain useful fallback polling for windows shorter than 15 seconds. Add delayed-callback and `processing`-to-`completed` polling tests.

packages/runtime/src/providers/atlascloud-transport.ts
```diff
@@ -178,6 +237,72 @@ export async function atlasPoll(
...
+  const maxAttempts = opts.maxAttempts ?? 600;
+  const windowMs = pollInterval * maxAttempts;
+  const reconcileInterval = Math.max(
+    pollInterval,
+    ATLAS_WEBHOOK_RECONCILE_INTERVAL_MS
+  );
+
+  const settled = new AbortController();
+  const onAbort = (): void => settled.abort(opts.signal?.reason);
+  opts.signal?.addEventListener("abort", onAbort, { once: true });
```
georgi: [P2] F2 (Correctness): propagate an already-aborted caller signal

Adding an `abort` listener does not deliver an abort that already happened. When `opts.signal.aborted` is true on entry, this new controller stays live, and both the registry and reconciliation poll receive that live signal instead of the cancelled caller signal. A reproduction with a pre-aborted controller still issued a poll and returned a completed prediction; the same call without `NODETOOL_PUBLIC_URL` rejected with the cancellation reason without issuing a request. This also affects cancellation that occurs before the wait starts after submission.

Check the caller's existing aborted state before registering or polling, or compose the caller and settlement signals with `AbortSignal.any`. Add wrapper-level tests for cancellation both before entry and during the wait, rather than testing the registry's pre-aborted handling alone.

packages/websocket/src/routes/atlascloud-webhook.ts
```diff
@@ -0,0 +1,455 @@
...
+export const ATLASCLOUD_WEBHOOK_PATH = "/api/providers/atlascloud/webhook";
+export const ATLASCLOUD_WEBHOOK_JWKS_URL =
+  "https://api.atlascloud.ai/api/v1/webhooks/jwks.json";
+export const ATLASCLOUD_WEBHOOK_TIMESTAMP_WINDOW_SECONDS = 300;
+export const ATLASCLOUD_WEBHOOK_JWKS_CACHE_TTL_MS = 24 * 60 * 60 * 1000;
+export const ATLASCLOUD_WEBHOOK_MAX_BODY_BYTES = 1024 * 1024;
+
+const HEADER_ID = "x-atlascloud-webhook-id";
+const HEADER_TIMESTAMP = "x-atlascloud-webhook-timestamp";
+const HEADER_SIGNATURE = "x-atlascloud-webhook-signature-ed25519";
```
georgi, line 40: [P2] F3 (Provider contract): accept the alternative Ed25519 signature header

AtlasCloud's [webhook contract](https://www.atlascloud.ai/docs/en/webhooks) says to prefer `X-AtlasCloud-Webhook-Signature-Ed25519` when present and fall back to `X-AtlasCloud-Webhook-Signature` for pure Ed25519 deliveries after the HMAC migration. This implementation only reads the former, so a correctly signed delivery using the documented alternative gets `400 missing_header` before verification. I reproduced this with real Ed25519 keys: the same signature passes in the preferred header and fails when moved to the alternative header.

Read the alternative only when the preferred header is absent, then apply the same Ed25519 decoding and verification. Keep legacy HMAC-only signatures rejected. Add tests for both Ed25519 header forms and for precedence when both headers are present.

packages/websocket/src/routes/atlascloud-webhook.ts
```diff
@@ -0,0 +1,455 @@
...
+      }
+      return "bad_signature";
+    };
+
+    let outcome = matches(keys);
+    if (outcome !== "signed") {
+      // A signed callback can arrive immediately after AtlasCloud rotates keys.
+      // One bounded refresh handles that while preventing retry storms.
+      try {
+        keys = await this.refreshKeys();
```
georgi: [P2] F4 (Security): avoid a JWKS fetch for every invalid signature

This branch refreshes on `bad_signature` as well as `unknown_key`, regardless of the cache TTL. With a warm cache and a known `kid`, three sequential callbacks carrying syntactically valid but incorrect signatures caused three additional JWKS fetches in an isolated reproduction. No signing key is needed to trigger those fetches. `refreshPromise` only coalesces overlapping requests, so sequential requests to this public route bypass the day-long cache and can drive outbound provider requests up to the ingress limit.

Reject a bad signature against a known cached key without refreshing. Restrict rotation refreshes to unknown key IDs, with a cooldown or bounded negative cache for repeated misses, and test that invalid signatures cannot repeatedly force network access while the cache is fresh.

georgi (commented):
Reviewed head `64f80c39b9b17fd5b605e4ad75260a6fea92b9a4`. Four findings are attached inline: F1, premature reconciliation timeout; F2, loss of an already-aborted caller signal; F3, rejection of the documented alternative Ed25519 signature header; F4, forced JWKS refreshes for invalid signatures despite a warm cache.

Verification: inspected the diff, transport callers, registry, ingress/auth wiring, and relevant tests, and checked AtlasCloud's current webhook contract. Ran two isolated Node.js reproductions using the reviewed wait/verifier control flow. They reproduced all four findings, including a no-webhook cancellation control and real Ed25519 signing/verification. HTTP/JWKS responses and logging were injected or stubbed. These were not runs of the repository's Vitest suites. A local checkout was blocked by this environment's GitHub DNS resolution, so I did not rerun the workspace tests, lint, or typecheck. GitHub reports the existing Quality Gate as successful for this head.

packages/runtime/src/providers/atlascloud-transport.ts
georgi: Fixed in cb4500d6. Confirmed the arithmetic and the failure: with `pollInterval: 3000, maxAttempts: 4` the window is 12s, `ceil(12000 / 15000)` is 1 attempt, and `atlasPoll` neither sleeps after its last attempt nor treats `processing` as anything but exhaustion, so it rejected at once and the `finally` dropped the waiter.

The deadline now sits with the callback wait, which already holds `windowMs`. `atlasPoll` throws a typed `AtlasPollBudgetExhausted`, and the reconciliation half swallows that one error and stops instead of settling the race:

```ts
}).catch((error: unknown): Promise<AtlasPollResult> => {
  if (error instanceof AtlasPollBudgetExhausted) {
    return new Promise<AtlasPollResult>(() => undefined);
  }
  throw error;
});
```

Every other rejection — a terminal failure, a 4xx, an abort — is still the answer and propagates.

Two further changes from your note. The interval clamps to half the window, so a window shorter than 15s gets real fallback polling rather than one read:

```ts
const reconcileInterval = Math.max(
  pollInterval,
  Math.min(ATLAS_WEBHOOK_RECONCILE_INTERVAL_MS, Math.floor(windowMs / 2))
);
const reconcileAttempts = Math.floor(windowMs / reconcileInterval) + 1;
```

The `+ 1` is because `atlasPoll` reads before each sleep, so N attempts span N-1 intervals — the old `ceil` under-covered the window even when it did not collapse to one attempt. And a window that expires with neither half terminal now reports `AtlasCloud job timed out`, the same error the polling path gives, so a slow prediction fails identically in both modes rather than being reported as a webhook problem.

Three tests added, per your list: a delayed callback after the reconciliation budget runs out, `processing` → `completed` polling, and window expiry. I reverted the fix to confirm the first one catches the regression:

```
× keeps waiting when the reconciliation budget runs out first 35ms
AssertionError: expected false to be true          // waiter already dropped
AssertionError: promise rejected "AtlasPollBudgetExhausted: AtlasCloud job …" instead of resolving
```

---

packages/runtime/src/providers/atlascloud-transport.ts
georgi: Fixed in cb4500d6, using the `AbortSignal.any` option you suggested:

```ts
const settled = new AbortController();
const waitSignal = opts.signal
  ? AbortSignal.any([opts.signal, settled.signal])
  : settled.signal;
```

That composes rather than subscribes, so an already-aborted caller signal makes the composite aborted on creation. It also removes the listener bookkeeping — the old `addEventListener` plus `removeEventListener` in `finally` is gone, and the `finally` now only aborts `settled`.

Your diagnosis was exactly right, including the asymmetry: without `NODETOOL_PUBLIC_URL` the call went straight to `atlasPoll`, whose loop checks `signal.aborted` and rejects without a request, while the webhook path issued one. Both now reject before any request.

Two wrapper-level tests added as you asked, rather than relying on the registry's own pre-aborted handling: cancellation before entry (asserting `poll.calls()` is 0, so nothing was issued) and cancellation during the wait. Reverting to the listener form fails the first:

```
× honours a signal that was already aborted before the wait 9ms
```

---

packages/websocket/src/routes/atlascloud-webhook.ts
georgi: Fixed in cb4500d6. You are right about the contract and I was wrong about it.

I checked the docs again before changing anything, since this is a claim about an external spec. They do say it:

> "Once HMAC is retired the Ed25519 signature moves to `X-AtlasCloud-Webhook-Signature` — so prefer the `-Ed25519` header when present and fall back to `-Signature`"

and the header table row reads "In the pure Ed25519 scheme this carries the Ed25519 signature instead". My first read of that page returned a condensed summary that labelled the bare header "HMAC-SHA256 signature (legacy)" with no migration note, and I built the verifier on that. The fallback was missing because I had the contract wrong, not because I decided against it.

Implemented as you described — the alternative is read only when the preferred header is absent, then the same decode and verification path:

```ts
const signature =
  parseHeader(headers, HEADER_SIGNATURE) ??
  parseHeader(headers, HEADER_SIGNATURE_FALLBACK);
```

Legacy HMAC stays rejected without a separate branch: a hex HMAC is 64 characters, which fails the `{86,88}` base64url check that a 64-byte Ed25519 signature passes, so a migration-era delivery carrying only the HMAC gets `400 invalid_header` rather than being verified under the wrong scheme.

Three tests, covering both forms and the precedence case you named: the signature accepted in the fallback header, the `-Ed25519` header preferred when both are present (with a hex HMAC in the bare one, which would fail the callback if read first), and an HMAC-only delivery rejected. Reverting the fallback fails the first:

```
× accepts the signature in the documented fallback header 22ms
```

---

packages/websocket/src/routes/atlascloud-webhook.ts
georgi: Fixed in cb4500d6. Reproduced your measurement before changing it — three bad signatures against a warm cache and a known `kid` caused four fetches total, one cold-start plus one per attempt.

The refresh is now reachable only from an unknown key id, which is the only outcome that can mean the cache is stale. A signature that fails against a key AtlasCloud does publish is simply wrong and never touches the network:

```ts
if (outcome === "unknown_key" && this.claimRotationRefresh()) {
```

For the repeated-miss half I took the cooldown rather than a negative cache: one rotation refresh per 60s, so cycling made-up key ids cannot drive fetches either. A bounded negative cache would need sizing and eviction, and a cooldown cannot be exhausted by varying the id, which is the cheaper attack. A genuine rotation still resolves within one cooldown.

One subtlety worth naming: only refreshes an unknown key id provokes count toward the cooldown. The cold-start load and a TTL expiry are the verifier's own doing and cannot be driven by a caller, so they leave the budget intact. Without that split the cold-start fetch consumed the cooldown and the first rotation after boot was blocked for 60s — the existing rotation test caught exactly that while I was writing this.

Two tests, matching what you asked for. Three invalid signatures against a fresh cache leave `fetches` at 1 (the cold start alone), and three unknown key ids produce exactly one rotation refresh rather than three. Reverting to the old `outcome !== "signed"` branch fails both:

```
× does not refetch the JWKS for a signature that fails a known key
AssertionError: expected 4 to be 1
× allows one rotation refresh per cooldown for unknown key ids
AssertionError: expected 4 to be 2
```

---

georgi:
`Quality Gate (npm run check) / test-packages-nodes` failed on `cb4500d6`, and I do not believe it is this PR's. Re-run queued, which is the one re-run I'll spend on it.

What failed, in `@nodetool-ai/base-nodes`:

```
FAIL tests/directed-campaign-kit.test.ts > Directed Campaign Kit workflow contracts
     > executes deterministic layouts, escapes exact copy, preserves original artifacts,
       and restores the portable record
Error: Test timed out in 60000ms.

Test Files  1 failed | 35 passed (36)
     Tests  1 failed | 2023 passed | 11 skipped (2035)
```

Why I read it as runner pressure rather than a defect:

- It is a timeout, not an assertion. The same test passes locally on this exact commit in **8.23s of test time** (24.7s wall including transform and import), against a 60s limit.
- The same job logged `[vitest-pool]: Timeout terminating forks worker` for a *different* file in the package, `tests/example-workflows-execute.test.ts`. Two independent timeouts in one package points at the runner, not at either test.
- That suite reported `import 619.04s` for 36 files, which is not a healthy import budget.
- The diff cannot reach this test. It touches the AtlasCloud callback path only: the transport, its registry, the ingress route, and their tests. `directed-campaign-kit` exercises Code and control nodes through the kernel and never constructs an AtlasCloud provider. The one shared file is `packages/runtime/src/providers/index.ts`, where the change is three added export lines with no side effects.

The honest counterpoint: this job was green on the previous head `64f80c39` and red on `cb4500d6`, so it did change state across my push. I could not find a mechanism for that, and the local run plus the second timeout in the same job are what tip me to load. If the re-run fails the same way, I will treat it as real and root-cause it properly rather than re-running again.

Note for whoever owns that suite: a 60s limit on a test whose body takes ~8s locally leaves little headroom on a loaded runner. Raising it, or splitting the four behaviours in that one `it` into separate cases, would make this leg less sensitive to scheduling. I have not changed it here — it is outside this PR, and quieting a timeout is not something to do from inside an unrelated change.

---

PR #5763, opened 2026-09-13
fix(execution): close four durable fal recovery gaps

georgi:
## What changed

Four durable fal recovery defects found in a review of the recovery path. (1) A live render saved its asset and left the attempt with `provider_status: "succeeded"` and no next check, while `attachment_status` stayed `pending`; the attachment recovery query excludes exactly that row, so a crash between `runGeneration` and the storyboard write stranded a paid asset forever. Attempts now stay scheduled while a destination intent is unapplied, the settled outcome is projected onto the generation through `Prediction.settleAttachments` (a completed, ready generation holds no lease, so the write is fenced by the status it replaces), and an intent that names no output key now picks a media row rather than a structured one that could never attach. (2) The recovery decoder did not know the flat `audio_url` / `video_url` shapes the live fal extractors read, so those results became structured outputs the production finalizer marked ready with no download and no asset — a silent false completion; the flat aliases are decoded as media, an alias repeating its envelope's URL is deduped, and an audit keeps the decoder level with the live extractors. (3) After a lost generation-state write, a second pass found nothing retryable and went back to the provider; the worker now records the provider result on the attempt before the first output row is written and repairs the generation from its ready rows against that manifest, falling back to the provider when the row set is incomplete. (4) `runOnce` minted one lease window for a whole batch, so a late item could be claimed with an expiry already in the past, and the server's 30s timer could overlap passes; the window is now minted at each claim and one pass runs at a time per worker. The host finalizer and storyboard attach moved out of `server.ts` into `generation-recovery.ts` so the production wiring is testable, and that attach skips the storyboard write when the document already carries the asset.

## Verification

- [x] `npm run test:affected` — passes with `VK_DRIVER_FILES` pointing at lavapipe. Without a Vulkan ICD, 14 shader-backed `base-nodes` image tests fail with "No WebGPU adapter available (Node/Dawn)" (the environment gap in `docs/dev-environment.md`); both suites pass once the driver is installed. One `packages/cli` `run-dsl` case timed out at 30s under load and passes on its own re-run (10/10).
- [x] `npm run typecheck` — `typecheck:electron` clean. `typecheck:web` fails on `web/src/stores/workflowUpdates.ts:1583` (TS2367, `"booting"` comparison), which reproduces unchanged on the base commit and is untouched by this diff.
- [x] `npm run lint` — clean.
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — Gate: 8/8 selfchecks passed.

Each regression test was observed failing against the unfixed code:

- Reverting the lifecycle scheduling change: all three `generation-attachment-recovery` tests fail (`expected [] to deeply equal [ 'gen-attach-1' ]`, `expected "vi.fn()" to be called 1 times, but got 0 times`).
- Removing the `audio_url` alias: `fal-media-shape-parity` fails (`expected [ 'audio_url' ] to deeply equal []`) and the production-finalizer test fails because no download is attempted (`expected "vi.fn()" to be called with arguments: [ 'https://fal.media/clip.mp4', …(1) ]`).
- Reverting the repair and lease changes: `repairs a generation from saved outputs without another provider read`, `leases every item in a slow batch past the moment it is claimed`, and `runs one pass at a time per worker` all fail.

## Agent capabilities

Skipped: no capability added, and no declared capability contract changed.

## New checks

- [x] `packages/execution/tests/fal-media-shape-parity.test.ts` was inverted once (alias removed) and observed failing; the failing output is in **Verification** above.
- [x] The audit asserts the exact set of result fields it found in the live extractors, so it cannot pass by matching nothing.

https://claude.ai/code/session_012zBdNA3TTf89cGSUqYdETa

---
_Generated by [Claude Code](https://claude.ai/code/session_012zBdNA3TTf89cGSUqYdETa)_


georgi:
Three checks were red on 640bc94. None of them is this PR's; all three come from commits `main` picked up after this branch was cut (`96da0b5..94517a12`). I merged `main` in and pushed a fix for the one I could fix safely.

**1. `build-and-check` (Docs CI) — fixed here.** HTML-Proofer reported 16 internal-link failures, all in `_site/plans/provider-webhooks.html`, a page added to `main` in 7bf9d5c and untouched by this diff. Its links point at repository source (`../../packages/...`) and at `plans/generation-recovery.md`, which the built site excludes. Docs CI only runs on pull requests touching `docs/**`, so it never ran against that commit on `main`; building `main` alone locally reproduces the same 16 failures, while this branch's docs before the merge gave 0. `plans/provider-webhooks.md` is an internal implementation plan, so I excluded it from the site exactly as its sibling `plans/generation-recovery.md` is excluded (one line in `docs/_config.yml`). Verified with the CI command: 16 failures before, 0 after (1804 internal links, 859 files), with the agent-artifact and privacy-guard steps still passing.

**2. `Quality Gate / harness-gate` → `npm run validate:examples` — not this PR's, no fix ported.** It fails on `packages/base-nodes/nodetool/examples/nodetool-base/Generate a Native-Audio UGC Testimonial.json` and `packages/base-nodes/nodetool/examples/apps/ugc-product-video.app.json`, both added to `main` in 28ddec1 and byte-identical to `main` here:

```
error Property "model" requires a video_model to be selected [nodetool.video.ReferenceToVideo generate]
```

The `generate` node ships an empty placeholder — `{"type":"video_model","provider":"","id":"","name":"","supported_tasks":["reference_to_video"]}`. The patch is to fill `provider`, `id` and `name` with the model the example is meant to run (the shape `find_model` returns on its top-level `ref`), which is a content decision about which paid model the shipped example points at, so I am not guessing it here. Every other selfcheck in the gate passed, including this PR's own `generation-recovery` suite.

**3. `Run workflow suite in browser` — not this PR's, no fix ported.** The `brand-a-ugc-product-video` workflow from the same commit fails in the browser suite:

```
brand-a-ugc-product-video: failed (Node "build-timeline" failed: The transcription returned no word timestamps. Use whisper-1 with timestamps enabled.)
```

The fix belongs to that example's transcription step, not to this diff.

Evidence that 2 and 3 are red on the base branch: on `main`'s own push of 94517a12, **Test** (which runs the quality gate) and **E2E Workflow Runner** both failed, while both were green on 96da0b5 — the commit this branch was cut from. No re-run was spent: the cause is deterministic and reproduces locally on a clean `main` checkout.

I have not re-run anything else and will keep watching this PR. Once `main`'s example is fixed, checks 2 and 3 should clear here without another change from me.

---

PR #5767, opened 2026-09-13
fix(examples): restore the Seedance 2.5 model on the UGC testimonial graph

georgi:
## What changed

`nodetool harness gate` was red on `main`: the `validate` selfcheck failed because "Generate a Native-Audio UGC Testimonial" shipped an unselected `video_model`. The Seedance switch in 7a09763 updated the graph's description, node titles, and prompt to Seedance's `@Image1`/`@Image2` reference syntax, but blanked `provider`, `id`, and `name` instead of assigning the new model — the recipe prose already said "The finished sample used Seedance 2.5 on AtlasCloud" while the graph pointed at nothing. `validate:examples` exists to catch exactly this drift; its header names "an unselected model" as the motivating case. This assigns `atlascloud / bytedance/seedance-2.5/reference-to-video`, regenerates the `ugc-product-video` bundle and the marketing recipe entries (which now correctly surface the required `ATLASCLOUD_API_KEY`), and updates the test to assert the shipped model. The app's `ModelSelect` binding is untouched, so the model stays user-selectable and simply gains a working default.

Note for reviewers: `packages/base-nodes/AGENTS.md` says example model fields "must ship empty" so `applyDefaultModels` can fill them per user. That rule is contradicted by both the gate and the corpus — 185 of 186 model fields across `nodetool-base/*.json` ship populated, and this graph was the only empty one. Reconciling that doc is left out of this PR deliberately.

## Verification

- [x] `npm run test:affected` — passes except two failures in `packages/runtime` that reproduce identically on a clean `origin/main` checkout (see Pre-existing below). `runtime` is only selected here because the generated marketing file is unmapped, which widens selection to all suites.
- [ ] `npm run typecheck` — one failure, pre-existing on `origin/main` and untouched by this diff (see below)
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — `Gate: 6/6 selfchecks passed`

Also run:

- `npm run dev:nodetool -- harness gate --all` — `Gate: 32/32 selfchecks passed` (was 31/32 before this change)
- `npm run validate:examples` — `310/310 graphs validate cleanly` (was `308/310`)
- `npm run test --workspace=packages/base-nodes -- ugc-product-video` — 5 passed
- `npx tsx marketing/scripts/generate-recipes.mjs --check` — clean
- `npm run check:agents-docs`, `npm run check:lockfile` — pass

The check was observed failing before the fix:

```
Gate: 3/4 selfchecks passed
  FAIL validate (exit 1): npm run validate:examples

308/310 graphs validate cleanly.
  FAIL  .../nodetool-base/Generate a Native-Audio UGC Testimonial.json
    error Property "model" requires a video_model to be selected [nodetool.video.ReferenceToVideo generate] (property)
  FAIL  .../apps/ugc-product-video.app.json › Generate a Native-Audio UGC Testimonial
    error Property "model" requires a video_model to be selected [nodetool.video.ReferenceToVideo generate] (property)
```

### Pre-existing failures, not from this diff

Each reproduces on a clean `origin/main` checkout and is left alone:

1. `npm run typecheck` — `web/src/stores/workflowUpdates.ts:1583` compares a prediction status against `"booting"`, which no longer exists anywhere in `packages/protocol`. Fixing it is a product call (add the status back, or drop the dead branch).
2. `packages/runtime` `tests/url-egress-audit.test.ts` — 2 failures: `packages/websocket/src/server.ts: fetchExternalMedia not called`. The egress inventory lists that file as guarded but it no longer calls the screening function; resolving it needs a security judgment about whether the egress moved or the guard was dropped.

Three further suites (14 image tests, and `tests/providers/video-frame-fallback.test.ts`) failed only from missing sandbox packages, not code: they pass after `apt-get install mesa-vulkan-drivers` (the Vulkan ICD that `AGENTS.md` prescribes for headless WebGPU) and `ffmpeg`.

## Agent capabilities

## New checks

https://claude.ai/code/session_01AQKk3jFrESLoSF7VoMt6N2

---
_Generated by [Claude Code](https://claude.ai/code/session_01AQKk3jFrESLoSF7VoMt6N2)_


georgi:
`Typecheck, lint & build` failed on 9ad8ee7 at the `gen:templates` drift check:

```
##[error]templateEntries.generated.ts / public/templates is stale. Run 'npm run gen:templates' in marketing/ and commit the result.
 marketing/src/data/templateEntries.generated.ts | 99 ++++++++++++++++++++++++-
```

This one isn't from this PR. The stale row is `Assemble a UGC Product Video`, whose graph is present on `origin/main` while its entry is absent from `templateEntries.generated.ts` there — so the generator reports drift on the base branch too. This PR's change is a `video_model` field on a different graph, and it does not feed the template catalog at all: the regenerated diff is exactly the one missing entry plus a comma reflow, with none of this PR's edits in it.

The fix exists and is trivial, so I ported it in (2763e81) rather than waiting on the base: `npm run gen:templates` in `marketing/`. Re-running the generator afterwards produces no further diff, which is the same assertion CI makes. It will no-op once `main` carries the regeneration.

Unrelated to this check, one pre-existing failure is still expected in `npm run check`: `web/src/stores/workflowUpdates.ts:1583` compares a prediction status against `"booting"`, which no longer exists in `packages/protocol`. It reproduces on a clean `origin/main` checkout and isn't something this PR should decide — it needs a call on whether to restore the status or drop the dead branch.

---

georgi:
`Run workflow suite in browser` also failed, and this one is red on the base branch rather than from this PR.

```
brand-a-ugc-product-video: failed (Node "build-timeline" failed: The transcription returned no word timestamps. Use whisper-1 with timestamps enabled.)
```

The same job on `main` at `aea51e1` — this branch's merge-base — fails with the byte-identical message, same workflow and same node ([run 34773772179](https://github.com/nodetool-ai/nodetool/actions/runs/34773772179)). It is also not the workflow this PR touches: `Brand a UGC Product Video` is a sibling in the same bundle, and the failing node is `build-timeline`, not the `generate` node whose `video_model` this PR fills in.

Cause, for whoever picks it up: `build-timeline` (`nodetool.code.Code`) throws that message itself when its `words` input is empty. `words` comes from `transcribe` (`openai.audio.Transcribe`), which under the e2e suite's scripted provider returns no word-level timestamps. So it is a fixture/model-selection gap in the suite, not a regression in graph content.

I have not fixed it here. It needs a call on whether the example should pin an ASR model that returns word timestamps or the scripted provider should supply them, and that is a separate change from this PR's one-field fix — widening this PR into another workflow's model selection is exactly the kind of edit that should be decided deliberately. No re-run spent: the base-branch run is stronger evidence than a retry would be.

---

PR #5769, opened 2026-09-13
feat(timeline): Final Cut-style audio fades

georgi:
## What changed

Audio fades in the timeline editor were a pair of numbers in the inspector: linear only, drawn as a static triangle on the clip, and dropped entirely by the ffmpeg export. This brings them up to what [Final Cut offers](https://support.apple.com/de-de/guide/final-cut-pro/ver39580fed/mac). A clip the mixer sounds now carries a grip at each top corner — drag it inward to fade in or out, back to the corner to remove it, arrow keys to move a focused one. Right-clicking a grip picks the curve from Linear, S-curve, +3 dB and -3 dB, stored on the clip as `fadeInShape`/`fadeOutShape` (absent stays linear, so existing documents sound as they did) and editable in the inspector too. Alt+T applies half-second fades to the selection, clamped to what a short clip has room for, and the clip menu does the same for one clip. The preview schedules the authored curve instead of a straight ramp and resumes it mid-fade on a seek, and the ffmpeg export applies fades at all — each shape maps to the `afade` curve with the same gain function, so a render matches what was previewed. One behaviour change beyond the feature: fade ramps no longer draw on image and overlay clips, because a fade is an audio envelope the compositor renders nothing from, and they now draw on midi clips, which the mixer does fade.

Design notes worth flagging:

- **-3 dB is ffmpeg's `iqsin`**, the exact inverse of the +3 dB quarter sine, rather than a curve sitting exactly 3 dB under linear at the midpoint. Picking curves ffmpeg already implements is what makes preview and export agree sample-for-sample; the alternative was a `volume` expression per clip.
- **A time-remapped clip's stretches are summed back into one stream before fading.** A stretch covering only part of a ramp cannot carry it — `afade` always runs a full 0→1 — so `amix=…:normalize=0` reassembles the clip and the fade goes on that. Unfaded clips keep the flat chain they had.
- **Mobile is untouched.** `mobile/src/documents/` carries its own plain-TS clip types; unknown fields survive its JSON round-trip, so a shape authored on desktop is not lost there, but mobile cannot set one.

## Verification

CI on `7000564`, 29 checks: **27 green, 2 red**, and both red ones are pre-existing failures that also affect `main`. Details in the comments; summary here.

Green, including every check this change could plausibly touch: `typecheck`, `lint`, `build`, `bundle`, `harness-gate`, `test-app`, `test-packages-nodes`, `test-packages-websocket`, `test-packages-agents`, `docker`, `deps`, `typescript-6-fallback`, `Workflow Integration Tests`, `Workflow Runner Browser E2E`, `Playwright smoke tests`, `Load every route without errors`, CodeQL.

Red, neither caused by this diff:

1. **`Run workflow suite in browser`** — `brand-a-ugc-product-video` dies on its own Code node's `rawWords.length === 0` guard when transcription returns no word timestamps. Identical on `main` across three consecutive runs since ~16:17Z, hours before this branch existed. Re-run once; failed again, so it is real rather than a provider flake. No fix exists to port. [Detail](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656690677).
2. **`test-packages-core`** — fails on `@nodetool-ai/runtime#test`, the url-egress audit. The durable FAL recovery download moved from `packages/websocket/src/server.ts` to `generation-recovery.ts` in `aea51e1` (#5763) and the inventory did not follow; the egress is still screened by `fetchExternalMedia` in its new home, so no guard was dropped. Latent on `main` behind a turbo cache hit, which this diff invalidated. Two-line patch proposed, deliberately not pushed here. [Detail](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656775668). `quality` is the aggregate of this one, not an independent failure.

`7000564` also carries a **port of #5767**, byte-identical to its head, to clear the `validate:examples` breakage that `harness-gate` was failing on. That job is now green. It no-ops when #5767 merges. [Detail](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656765425).

Locally, exercised directly:

- `packages/timeline` — 94 files, 1195 tests, including the new `tests/audioFade.test.ts` (curve shape, monotonicity, the dB relationships the names claim, clamping, resume sampling).
- `packages/video-nodes` — 39 files, 344 tests, including new cases pinning `afade=t=in:st=0.000:d=0.500:curve=qsin`, the sub-mix on a remapped clip, and that an unfaded clip's chain gains no `afade`.
- `packages/protocol` — 86 files, 1491 tests.
- `web` timeline and stores — 140 suites, 1224 tests, including the new fade-handle gesture suite, `AudioGraph.fades.test.ts` and `TimelineStore.fades.test.ts`.
- Two packages had to be installed before the suites could run at all: `mesa-vulkan-drivers` (the Vulkan ICD `AGENTS.md` prescribes for headless WebGPU) and `ffmpeg`.

The preview/export parity claim was checked against a real ffmpeg rather than asserted: rendering a constant tone through `afade` with each of `tri`, `qsin`, `hsin` and `iqsin` and comparing the PCM envelope to `fadeShapeGain` gives a maximum difference of 0.00000 over 1000 samples for all four. A unit test pins the same relationship against ffmpeg's published gain functions so the mapping cannot drift.

One correction to an earlier revision of this description: it reported `npm run typecheck` as failing on `web/src/stores/workflowUpdates.ts:1583` and called that pre-existing. The failure is real in this sandbox and does reproduce there on a clean `origin/main` checkout, but CI's `typecheck` job passes on `main` and on this head, so it is an artifact of the local environment and not a repository breakage. The checklist above now reflects the CI result.

## New checks

- [x] `audioFade.test.ts` asserts the protocol's `clipFadeShapeEnum` carries exactly the shapes the curve math implements — the one place the two duplicated lists could drift. Inverted by removing `"minus3dB"` from the protocol tuple, then restored. Observed failing:

  ```
   FAIL  tests/audioFade.test.ts > the protocol's fade-shape enum > carries exactly the shapes the curves implement
  AssertionError: expected [ 'linear', 'plus3dB', 'sCurve' ] to deeply equal [ 'linear', 'minus3dB', …(2) ]
  ```

  Worth knowing for anyone relying on it: the test imports the protocol through its package export, which resolves to `dist/`, so it catches the drift only once `@nodetool-ai/protocol` is rebuilt. Turbo's `test` → `build` dependency does that in CI; editing the tuple and rerunning vitest alone does not (that inversion passed until I rebuilt).

- The curve/ffmpeg parity test iterates `CLIP_FADE_SHAPES` and asserts a curve was found for each, so it cannot pass by matching nothing.

https://claude.ai/code/session_015Y6do3Fy6Vh433v7tzhnck


georgi:
**`Run workflow suite in browser` is failing, and it is not this PR's.** Standing down on it, with the evidence:

```
brand-a-ugc-product-video: failed (Node "build-timeline" failed:
  The transcription returned no word timestamps. Use whisper-1 with timestamps enabled.)
```

That string is thrown by the example graph's own Code node, in `packages/base-nodes/nodetool/examples/nodetool-base/Brand a UGC Product Video.json`, on the `rawWords.length === 0` guard. It fires when the upstream transcription returns no words. This PR touches clip fade envelopes — `packages/timeline`, the ffmpeg audio filter chain, protocol schemas and the timeline editor UI. It adds no transcription code path, changes no Code node, and does not edit that graph.

The same failure, byte for byte, is on `main` itself:

| Run | Commit | Result |
| --- | --- | --- |
| [34773772179](https://github.com/nodetool-ai/nodetool/actions/runs/34773772179) | `aea51e1` (main, 18:14Z) | same message |
| [34771956441](https://github.com/nodetool-ai/nodetool/actions/runs/34771956441) | `8a6a901` (main, 17:38Z) | failure |
| [34768131450](https://github.com/nodetool-ai/nodetool/actions/runs/34768131450) | `94517a1` (main, 16:22Z) | failure |
| [34766326611](https://github.com/nodetool-ai/nodetool/actions/runs/34766326611) | `96da0b5` (main, 15:49Z) | last green |

So `E2E Workflow Runner` has been red on `main` since roughly 16:17Z, about six hours before this branch existed. Nothing to port: I looked for an open PR fixing it and found none — [#5767](https://github.com/nodetool-ai/nodetool/pull/5767) repairs a different red check (the `video_model` on the UGC testimonial graph, which is `validate:examples`/`harness-gate`), not the transcription path this one dies on.

I have re-run the failed job once ([run 34787406817](https://github.com/nodetool-ai/nodetool/actions/runs/34787406817)). Three consecutive deterministic failures on `main` argue it is real rather than a provider flake, but the node does call a live transcription service, so the re-run is the one piece of evidence the `main` history cannot supply. That re-run is now spent.

A root-cause fix looks like it belongs to whoever owns that graph: either the transcription node stopped returning `timestamp` arrays in the shape the Code node reads (`Array.isArray(item.timestamp)`), or the model behind it no longer emits word-level timestamps at all. Fixing it here would mean editing an unrelated example graph on a fades PR, so I have left it alone. Happy to take it as its own change if wanted.

`Quality Gate / harness-gate` is still running on this head. If it fails, expect it to be the `validate:examples` breakage that #5767 fixes — also red on `main` — and I will follow up here rather than open a second thread.

---

georgi:
**`Quality Gate / harness-gate` was the second red check, and I have pushed a fix for it** (`7000564`).

Only one selfcheck failed — `validate` (`npm run validate:examples`). The other seven passed, including this PR's own `timeline-validate` and `node-pack-parity`:

```
"harnessId": "validate", "command": "npm run validate:examples", "ok": false, "exitCode": 1
```

That is the unselected `video_model` on the two UGC testimonial graphs, red on `main` too, and [#5767](https://github.com/nodetool-ai/nodetool/pull/5767) already fixes it. Rather than wait for it to merge, I ported it: the seven files here are **byte-identical to #5767's head** (`aafa5bd` on `claude/fix-harness-gate-102msu`) — I verified each by sha256 — so this no-ops when #5767 lands instead of conflicting with it. If #5767 merges first, this commit simply disappears into the merge; if this merges first, #5767 becomes empty.

Verified before pushing:

- `npm run validate:examples` → `310/310 graphs validate cleanly` (was `308/310`)
- `npm run dev:nodetool -- harness gate --base origin/main` → `Gate: 10/10 selfchecks passed`, `validate` among them
- `npm run test --workspace=packages/base-nodes -- ugc-product-video` → 5 passed (the port includes that test's updated assertion, which would otherwise fail against the now-populated model)
- `npm run lint` → clean
- `npm run test:affected` → only the `packages/runtime` `url-egress-audit` failure that reproduces on a clean `origin/main`

This is the one deliberate widening of the PR. It is a ported fix, not new work, and I would rather hand back a green branch than a red one blocked on someone else's merge.

Still outstanding and **not** fixed by this push: `Run workflow suite in browser`, covered in my [earlier comment](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656690677) — the transcription/word-timestamps failure that is also red on `main` and has no fix to port.

---

georgi:
**Third red check: `Quality Gate / test-packages-core`.** Also not this PR's, but this one has a real root cause worth fixing, and I am proposing rather than pushing it because it touches the security egress inventory.

The job fails on one workspace:

```
Tasks:  85 successful, 87 total
Failed: @nodetool-ai/runtime#test
```

That is `packages/runtime/tests/url-egress-audit.test.ts`, two assertions:

```
keeps every guarded file actually calling its screening function
  → ["packages/websocket/src/server.ts: fetchExternalMedia not called"]
classifies every file that calls a screening function
```

**The guard moved; it was not dropped.** `packages/websocket/src/server.ts` now contains no `fetch`, no `safeFetch` and no `fetchExternalMedia` at all. The durable FAL output recovery download left that file in `aea51e1` ("fix(execution): close four durable fal recovery gaps", #5763) and now lives at `packages/websocket/src/generation-recovery.ts:201`:

```ts
const response = await fetchExternalMedia(sourceUrl, {
  signal: AbortSignal.timeout(60_000)
});
```

So the egress is still screened, in its new home. The inventory simply did not follow it, which is exactly why both assertions fail at once: the old path no longer calls its screening function, and the new path calls one with no entry (`generation-recovery.ts` appears zero times in both `packages/runtime/tests/url-egress-inventory.ts` and `docs/url-egress-inventory.md`). No unguarded egress is introduced by this, and no security hole is open.

The proposed patch is a path change in two places:

```diff
--- a/packages/runtime/tests/url-egress-inventory.ts
+++ b/packages/runtime/tests/url-egress-inventory.ts
@@
   guardedMedia(
-    "packages/websocket/src/server.ts",
+    "packages/websocket/src/generation-recovery.ts",
     "Durable FAL output recovery download",
     "A persisted FAL result URL is copied into NodeTool storage during recovery.",
     "provider-response"
   ),
```

```diff
--- a/docs/url-egress-inventory.md
+++ b/docs/url-egress-inventory.md
@@ -107 +107 @@
-| durable FAL output recovery | `packages/websocket/src/server.ts` | provider response |
+| durable FAL output recovery | `packages/websocket/src/generation-recovery.ts` | provider response |
```

I have not pushed it. `AGENTS.md` makes this inventory a security artifact and its audit the enforcement, so "the guard moved" deserves a human signing off rather than an audio-fades PR quietly re-pointing it — even though the evidence above is unambiguous. Say the word and I will push it here, or open it as its own PR where it belongs; the latter is my preference since it is `#5763` follow-up, not fade work.

Note on why this surfaced now: `test-packages-core` was green on `main`'s last run because `@nodetool-ai/runtime#test` was a turbo cache hit there (63 of 87 tasks cached). This PR's diff invalidated that entry, so the suite actually ran and the pre-existing breakage appeared. It reproduces on a clean `origin/main` checkout with this branch stashed — I checked before opening the PR.

Also confirming for the record: the `Run workflow suite in browser` re-run **failed again** with the same transcription error, which settles it as real rather than a provider flake, as suspected in my [first comment](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656690677). My one re-run is spent and I will not re-run it again. `Quality Gate / quality` is the aggregate of the above, not an independent failure.

All three of these landed on `d9ee552`. The harness-gate fix in `7000564` is running now.

---

georgi:
**#5767 merged, so I brought `main` in (`a9e9fed`) and the ported commit has become the no-op it was meant to be.**

The PR diff against `main` is now back to **33 files, +1716/−77** — purely the audio-fades change. None of the seven files I ported appear in it any more:

```
$ git diff origin/main HEAD --name-only | grep -E "marketing/|ugc-product-video|Native-Audio"
(nothing)
```

That closes the one widening concern I raised when pushing `7000564`. The commit stays in the branch's history, contributes nothing to the diff, and the merge was clean — no conflict on those files, since both sides held identical bytes.

Re-verified on the merged head before pushing:

- `npm run dev:nodetool -- harness gate --base origin/main` → `Gate: 8/8 selfchecks passed`
- `npm run validate:examples` → `310/310 graphs validate cleanly`
- `web` timeline + stores → 140 suites, 1224 tests, all passing
- `npm run lint` → clean

`main` also brought in #5768's sketch-store refactor; it touches nothing this PR does, and the suites above confirm it.

Unchanged and still outstanding, both pre-existing and both red on `main` as well as here:

1. `Run workflow suite in browser` — the transcription word-timestamps failure. No fix to port, re-run spent.
2. `test-packages-core` — the url-egress audit. Two-line patch proposed [above](https://github.com/nodetool-ai/nodetool/pull/5769#issuecomment-5656775668), still not pushed; it wants a human's read on a security artifact, and I think it belongs in its own PR as #5763 follow-up.

Everything this PR is responsible for is green.

---

PR #5774, opened 2026-09-14
Timeline: add ClipTake model (Phase 1 of AI video editing PRD)

georgi:
## What changed

Adds Phase 1 of a 5-phase "P0 AI Video Editing" PRD: a formal take model for timeline clips, so a shot can hold several generated alternatives that are compared and switched between without altering editorial state (position, duration, effects, transitions). This is the nondestructive foundation the PRD's later phases (subject tracking, smart reframe, auditions UI polish, professional interchange export, generative timeline editing) build on — only this phase is included here; the rest is intentionally out of scope for this PR.

Rather than introducing a parallel `ClipTake` type, this extends the existing per-clip generation-history record (`ClipVersion`/`TimelineClip.versions`) and its existing switch mechanism (`TimelineStore.restoreVersion`) with take semantics: `label`, `source`, `provider`, `model`, `prompt`, `negativePrompt`, `parentTakeId` on `ClipVersion`, and an explicit `activeTakeId` on `TimelineClip` kept in sync with `currentAssetId`. All new fields are optional and mirrored in the protocol Zod schemas so they survive PATCH round-trips.

New shared logic lives in `packages/timeline/src/takes.ts` (`selectTake`, `renameTake`, `deleteTake`), used by both the `TimelineStore` browser actions and four new headless agent ops (`list_takes`, `select_take`, `rename_take`, `delete_take`) routed through the existing `edit_timeline` op executor. `deleteTake` refuses to remove the sole take or the currently active take while others exist, matching the PRD's "deleting an active take requires another take to be selected first" rule.

The Take Browser (`ClipVersionHistory.tsx`) now shows take labels and adds rename/delete affordances per tile, with delete disabled (with a tooltip) on the active take.

## Verification

- `npm run test --workspace=packages/timeline` — 1144 passed, 61 skipped (GPU tests skip: no Vulkan driver in this environment, pre-existing)
- `npm run test --workspace=packages/agents` — 4555 passed, 13 skipped
- `npm run test --workspace=packages/protocol` — 1499 passed
- `npm run lint` — passes (pre-existing warnings only, none in touched files)
- `npm run dev:nodetool -- harness gate --base origin/main` — 8/8 selfchecks passed
- `npm run typecheck` — fails on 2 pre-existing errors unrelated to this diff (`web/src/stores/sketch/SketchSessionStore.ts`, `web/src/stores/workflowUpdates.ts`, last touched by unrelated commits); confirmed via `git diff` that this PR's commit does not touch either file

## Agent capabilities

- [x] `npm run capabilities:check` passes — "capability table is current (289 capabilities)"
- [x] The four new ops are additions to the existing `edit_timeline` capability's op vocabulary (not new capabilities), covered by `packages/agents/tests/timeline-op-parity.test.ts` (which asserts every `TIMELINE_OP_NAMES` entry has a matching bridge tool) and by new cases in `packages/timeline/tests/takes.test.ts`

---

https://claude.ai/code/session_01SZGAvNkfxnzTtybWJ2mQWS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SZGAvNkfxnzTtybWJ2mQWS)_


packages/timeline/src/takes.ts
```diff
@@ -0,0 +1,100 @@
...
+  const index = versions.findIndex((v) => v.id === takeId);
+  if (index === -1) {
+    return { clip, error: `No take "${takeId}" on this clip.` };
+  }
+  const version = versions[index]!;
+  const isActive = clip.activeTakeId
+    ? clip.activeTakeId === takeId
+    : clip.currentAssetId !== undefined &&
+      clip.currentAssetId === version.assetId;
+
```
georgi: [P2] Reconcile the active-take alias before using it to authorize deletion

`activeTakeId` is not kept in sync by every existing asset writer. In particular, `landDirectGen` in `web/src/hooks/timeline/useTimelineDirectGenJob.ts` appends a version and patches `currentAssetId`, but never updates this new field; `TimelineStore.patchClip` simply spreads that patch.

Reproduction: generate takes A and B, select A with `restoreVersion`/`select_take`, then directly generate C. The clip now has `activeTakeId = A` but `currentAssetId = C.assetId`. `ClipVersionHistory` and `list_takes` mark both A and C active, so clicking A cannot switch playback back to it. More seriously, this ID-first guard allows `delete_take(C)` and leaves the playing asset with no take record, while refusing deletion of inactive A.

Keep the alias synchronized when direct generation lands, and validate or derive active identity from the authoritative `currentAssetId` when the fields disagree. Add a regression covering take selection followed by `landDirectGen`, asserting exactly one active take and refusal to delete the take actually playing.

packages/timeline/src/takes.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (!version || version.status !== "success") return clip;
+
+  const restoredHash = version.dependencyHash;
+  const status: TimelineClip["status"] =
+    clip.dependencyHash === restoredHash ? "generated" : "stale";
+
+  return {
+    ...clip,
+    currentAssetId: version.assetId,
+    activeTakeId: version.id,
```
georgi, line 59: [P2] Do not replace a 3D clip's glTF source with a bake-history asset

The newly exposed `select_take` accepts any clip with a successful version, but `ClipVersion` is also used for 3D bake history. `bake_model3d_clip` stores the rendered movie in `versions` and `model3dStyle.bake`, while deliberately keeping the original glTF in `currentAssetId`.

Running `bake_model3d_clip`, then `list_takes`, then `select_take` on the returned bake ID reaches this assignment and replaces the glTF asset ID with the movie asset ID. The clip remains `mediaType: "model3d"`, and `model3dStyle.bake` is unchanged. Because `computeModel3DBakeHash` includes `currentAssetId`, the previously valid bake becomes stale and the scene falls back to live 3D rendering with a video where the glTF should be.

Either reject generic take selection for `model3d` clips in this phase, or implement bake-aware selection that preserves the source model and handles the bake record separately. Add an op-level bake -> list -> select regression that verifies the original glTF cannot be overwritten.

georgi (commented):
Reviewed head `1174e187ecd2163d0a9ddbe8776134db43030e82`, including the full diff and the affected generation, store, and 3D-bake paths. I found two P2 correctness issues that I recommend fixing before merging; inline comments include reproductions and proposed regression coverage.

1. Direct generation can leave `activeTakeId` pointing at a previous take. The UI then marks two takes active, while `delete_take` can remove the record for the asset that is actually playing.
2. The new generic `select_take` operation accepts a 3D clip's bake history and replaces its glTF source with a rendered video asset.

Validation: GitHub reports successful Test, E2E Workflow Runner, and Page Load Smoke runs for this head. I also ran isolated JavaScript reproductions using the extracted `selectTake`/`deleteTake` implementations and fixtures matching the relevant state writes in `landDirectGen` and `bake_model3d_clip`. Both reproduced the reported state corruption. These were focused helper-level checks, not a local run of the full repository suite or browser/render integration tests.

georgi (commented):
Both P2 findings fixed in 35004c0c.

**Active-take alias reconciliation**: `landDirectGen` now sets `activeTakeId` alongside `currentAssetId` when it lands. More importantly, added `activeTakeIdOf(clip)` (`packages/timeline/src/takes.ts`) that derives active identity from `currentAssetId` — the field every renderer/export path actually reads — falling back to the stored `activeTakeId` alias only when no version's `assetId` matches. `deleteTake`, the `list_takes` op, and `ClipVersionHistory`'s active-tile detection all now go through this helper, so the whole class of "some other writer forgot to sync the alias" bugs is closed, not just the `landDirectGen` reproduction. Regression tests added in `packages/timeline/tests/takes.test.ts` (`activeTakeIdOf` derivation, `deleteTake` refusing the take actually playing when the alias is stale) and extended the existing reconnect-settle test in `useTimelineDirectGenJob.test.tsx` to assert `activeTakeId` tracks the landed version.

**model3d bake protection**: took your suggested minimal fix — `selectTake` now refuses (no-ops) on `mediaType: "model3d"` clips, and the `select_take` op throws a clear error naming why, rather than letting a bake asset overwrite the glTF `currentAssetId`. Added the exact op-level bake → list_takes → select_take regression you asked for in `packages/timeline/tests/ops.model3d.test.ts`, asserting the glTF survives the attempt.

All timeline/agents/protocol suites plus the harness gate pass on the new head.

---

PR #5779, opened 2026-09-14
Timeline: add MediaTrack subject tracking (Phase 2 of AI video editing PRD)

georgi:
## What changed

Phase 2 of the "P0 AI Video Editing" PRD, building on the take model that landed in #5774. It adds `MediaTrack` — a source-time-anchored tracking primitive for a subject moving through a clip — plus the bindings that let another clip follow one. This is infrastructure other edits attach to, not a single effect; Phase 3 (smart reframe) and Phase 4 (generative object removal/replacement) are the PRD's declared consumers, and neither is in this PR.

**Data model.** `MediaTrack` / `MediaTrackSample` / `TrackBinding` in `packages/timeline/src/types.ts`, mirrored as Zod schemas in `packages/protocol/src/api-schemas/timeline.ts` and threaded through the `TimelineDocument` mappers in `packages/models` — without that last part the field is silently dropped on the first PATCH.

**Track math** (`packages/timeline/src/mediaTrack.ts`). Sample interpolation reuses `evalCurve`'s exact rule (held flat before the first and after the last sample, linear between) rather than inventing different edge-case behavior. Trim/split re-slicing follows `animation/sourceCurves.ts`'s boundary-synthesis recipe. Staleness mirrors `isGeneratedMatteStale`. The `markGenerating` / `afterFailure` / `applyResult` trio mirrors `generatedMatte.ts`, preserving its invariant: a failed regenerate never destroys a working result.

**Scene model.** A live `position` / `position_scale` binding folds into the same transform composition `offsetX/Y` and `scale` already use inside `resolveAnimatedLayerProps`, so every host — preview, export, server render, agent frame preview — resolves a bound layer identically instead of one host growing its own rule.

**Ops, reachable in both hosts.** `list_tracks`, `bind_to_track`, `unbind_track`, `delete_track_object` are implemented in `ops/apply.ts` *and* registered in the bridge that `edit_timeline` actually dispatches through, with document persistence wired in `applyOps`. Wiring only the first would have left them passing tests while returning "no such operation" in production.

Two things this pass found rather than added: the parity test was passing **vacuously** (the four ops were not yet in `TIMELINE_OP_NAMES`, so "covers every op" asserted over a set that excluded them), and once they were added it went red on a real divergence — `applyTimelineOp` re-sliced a clip's `MediaTrack` on trim/split while the bridge did not. The fix extracts `resliceTracksForTrimmedClip` / `resliceTracksForSplitClip` so both hosts call one implementation, which is what the parity table exists to enforce.

## Stated plainly, so review isn't misled

- **The `track_object` provider runner is not implemented.** It is a documented seam behind `Deps` injection that throws by default. I searched fal-, replicate- and video-nodes and confirmed no tracking-capable provider exists in this codebase (the `TRACK_FORWARD`/`TRACK_BACKWARD` hits are camera-dolly enums, unrelated). The capability's state machine, persistence, CAS and failure handling are real and tested with injected fakes; the provider call is not.
- **Region selection is numeric input, not a canvas drag tool.** Building drag-to-select in the viewer is a separate interactive-canvas feature. The Tracking panel's "Track subject" button is disabled with a caption saying no provider is wired.
- **Only `position` and `position_scale` binding modes affect rendering.** `transform`, `mask`, `effect_region` and `reframe` are typed so a later phase needs no schema migration, and `bind_to_track` refuses them explicitly rather than accepting a binding that silently does nothing.
- **A clip follows at most one track** in this pass.

## Verification

Run on this branch after merging latest `main`:

- `npm run build:packages` — 64/64
- `npm run test --workspace=packages/timeline` — 1184 passed, 61 skipped (GPU tests; no Vulkan ICD, pre-existing)
- `npm run test --workspace=packages/agents` — 4572 passed, 13 skipped
- `npm run test --workspace=packages/protocol` — 1499 passed
- `npm run test --workspace=packages/models` — 959 passed
- `npm run lint` — exit 0 (pre-existing warnings only)
- `npm run capabilities:check` — table current (290 capabilities)
- `npm run dev:nodetool -- harness gate --base origin/main` — **10/10**, `mappingViolations: []`
- `npm run typecheck` — 3 pre-existing errors in files this diff does not touch (`SketchSessionStore.ts` ×2, `workflowUpdates.ts`)

One honest note: the **first** gate run reported `data-models` failing (exit 1). It did not reproduce — two subsequent full gate runs passed 10/10 and the suite passes 959/959 directly. I could not attribute it to this diff, whose only `packages/models` change is an optional field plus its mappers, covered by those passing tests. Flagging it rather than omitting it.

## Agent capabilities

- [x] `track_object` is a new capability with real coverage (`packages/agents/tests/capabilities-track-object.test.ts` — success, reuse, staleness-triggered regenerate, failure-preserves-prior-result, CAS conflict, refusals), so it needs no gap note
- [x] The four new `edit_timeline` ops are covered by `timeline-op-parity.test.ts` (fixtures driving both hosts) and `timelines-op-input.test.ts` (DB round trips: bind → list → unbind, and delete-unbinds-followers)

---

https://claude.ai/code/session_01SZGAvNkfxnzTtybWJ2mQWS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SZGAvNkfxnzTtybWJ2mQWS)_


georgi (commented):
Reviewed head `fc23051aa70f0df69dc88033530522518c529592` against `bc44410476d58e5165d64c6a370a2cb42971d5c3`.

I would fix the five inline findings before merging: two P1 integration defects (document saves discard tracks, and browser preview/export never receive them), plus three P2 correctness defects (sampling the follower's clock, discontinuous smoothing, and controls that do not commit changes to an existing binding).

The missing tracking provider, disabled region-selection starter, and explicitly deferred binding modes are acknowledged scope limitations, not review findings. The issues below are reproducible with a seeded ready MediaTrack and do not require a provider.

Verification: inspected the complete diff and relevant save/render callers at this exact head. Ran isolated Node.js reproductions of the relevant algorithms, transcribed from the source: an ordinary document rebuild drops mediaTracks; omitting the fifth resolver argument disables the binding; a trimmed source produces -400px instead of 0px at the tested frame; smoothing jumps approximately 250px across a sample boundary on a 1000px canvas. These were not repository integration tests or browser tests. I could not run the repository suites because the execution environment could not resolve github.com for checkout.

packages/protocol/src/api-schemas/timeline.ts
```diff
@@ -1399,7 +1481,11 @@ export const timelineDocument = z.object({
   setup: timelineSetup.optional(),
   /** Sequence this one was retargeted from. Without this field Zod strips it
    * on every PATCH and the lineage is lost on the first save. */
-  templateId: z.string().nullable().optional()
+  templateId: z.string().nullable().optional(),
+  /** Subject/object tracks (P0 AI Video, Phase 2). Without this field Zod
+   * strips it on every PATCH, so a track and every clip following it are
+   * lost on the next save. */
+  mediaTracks: z.array(mediaTrack).optional()
```
georgi, line 1488: [P1] Preserve mediaTracks in the actual document save handler

Adding this schema field and the model mappers is not sufficient to make it survive a normal save. At this head, `packages/websocket/src/trpc/routers/timeline.ts` rebuilds `merged` from an explicit list ending in `tempo` and `setup`, then assigns `fields.document = JSON.stringify(merged)`. It never copies `mediaTracks` from either `input.document` or the current document. Consequently, opening a sequence containing a ready track and saving any document edit deletes all its tracks, while the clips retain their now-dangling `trackBinding`s. Even explicitly supplying mediaTracks in the request does not help.

Preserve/update this field in that handler, including explicit `[]` deletion, and carry it through the browser's load/merge/save state (`TimelineStore` and `buildTimelineDocumentPayload` currently omit it too). Add a seeded-track round-trip test through `timeline.update`, not just `edit_timeline`, covering both an unrelated document patch and an explicit tracks update.

packages/timeline/src/render/sceneModel.ts
```diff
@@ -1544,16 +1556,72 @@ const IDENTITY_TRANSFORM: ClipTransform = {
...
+  };
+}
+
 export function resolveAnimatedLayerProps(
   layer: { clip: TimelineClip; transform?: ClipTransform; opacity: number },
   currentTimeMs: number,
   canvas: RenderCanvas,
-  cache?: AnimationCompileCache
+  cache?: AnimationCompileCache,
+  mediaTracks?: MediaTrack[]
```
georgi: [P1] Wire the new tracking argument into browser preview and export

The browser's shared compositor adapter, `web/src/components/timeline/preview/compositeLayers.ts::buildCompositeLayer`, still calls this function with only four arguments. Its options do not carry mediaTracks either. Therefore `resolveTrackBindingOffset` always returns early at `!mediaTracks` in both live preview and in-browser export, regardless of a valid clip binding. Only the agent frame-preview callers were updated in this diff, so the same document can move in the agent preview but remain static in the browser/export.

Thread the document tracking context through the browser state and shared compositor callsites before treating the binding as rendered. Add a host-level parity test using a seeded ready track; directly calling this helper with the fifth argument, as the new tests do, cannot catch the missing wiring.

packages/timeline/src/render/sceneModel.ts
```diff
@@ -1544,16 +1556,72 @@ const IDENTITY_TRANSFORM: ClipTransform = {
...
+  canvas: RenderCanvas
+): { x: number; y: number; scale: number; rotation: number } | undefined {
+  const binding = clip.trackBinding;
+  if (!binding || !mediaTracks) return undefined;
+  if (binding.mode !== "position" && binding.mode !== "position_scale") {
+    return undefined;
+  }
+  const track = mediaTracks.find((t) => t.id === binding.trackId);
+  if (!track) return undefined;
+  const sourceMs = clipSourceMsAt(clip, currentTimeMs);
```
georgi: [P2] Sample the tracked clip's source clock, not the follower's

Here `clip` is the text/image/shape layer carrying the binding, whereas the samples are anchored to the source of `track.clipId`. The owner's in-point, start time, speed and time-remap are never consulted. For example, take a video starting at timeline 0 with `inPointMs: 5000`, a text follower starting at 0, and track samples at source 5000ms (x=0.1) and 9000ms (x=0.9). At timeline 2000ms the video shows source 7000ms, so the follower should be centered. This code instead samples the text's source 2000ms, holds the first sample, and places it at -400px on a 1000px canvas.

Resolve the owner via `track.clipId` from the render context and use `clipSourceMsAt(owner, currentTimeMs)`. Add coverage with different owner/follower start times and in-points, plus a retimed owner. Matching zero-based clocks in the current tests hides this defect.

packages/timeline/src/mediaTrack.ts
```diff
@@ -0,0 +1,317 @@
...
+    };
+  }
+  // The exact sample at sourceMs itself still gets one more blend pass, so a
+  // track sampled past its own last recorded point holds the smoothed value
+  // rather than snapping back to the raw last sample `sampleMediaTrackAt` held.
+  const blendExact = (a: number | undefined, b: number | undefined) =>
+    a === undefined ? b : b === undefined ? a : a * clampedFactor + b * (1 - clampedFactor);
+  return {
+    sourceMs,
+    x: blendExact(smoothed.x, exact.x),
```
georgi: [P2] Avoid discontinuities when a sample enters the smoothing history

The history already includes samples with `sourceMs <=` the requested time, then this final pass blends the exact sample again. As a result, the filtered curve jumps at recorded sample times even when the original curve is continuous. With samples `(0ms, x=0)` and `(1000ms, x=1)` and smoothing 0.5, the implementation returns approximately 0.4999995 at 999.999ms but 0.75 at 1000ms. That is a roughly 250px jump in 0.001ms on a 1000px canvas, caused by smoothing rather than the track.

Compute the filtered value once per recorded sample and interpolate between neighboring filtered values, or use another time-continuous formulation. Add assertions at each sample boundary and immediately on both sides; merely asserting that the exact-time value is smaller than the raw sample does not catch this.

web/src/components/timeline/Inspector/ClipTracking.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+    const binding = clip.trackBinding;
+    const [trackIdInput, setTrackIdInput] = useState(binding?.trackId ?? "");
+    const [mode, setMode] = useState<LiveBindMode>(
+      binding?.mode === "position_scale" ? "position_scale" : "position"
+    );
+
+    const handleModeChange = useCallback((value: string) => {
+      if (value === "position" || value === "position_scale") setMode(value);
+    }, []);
```
georgi: [P2] Commit mode and track-id edits for an existing binding

Once a clip has a binding, changing Mode only calls `setMode`; editing Track id similarly only calls `setTrackIdInput`. Neither updates the store. The only action that commits those local fields, `handleBind`, is attached to the Follow button rendered exclusively under `!binding`, so a bound clip has no way to apply those edits. For example, follow track A in Position mode, switch to Position + scale, then save/reload: the stored binding is still Position. Typing track B also leaves it following A. Adjusting another knob happens to commit the mode, but still uses the old `binding.trackId`.

Commit these changes through `bindToTrack` when already bound, or expose an explicit Apply/Rebind action. Add a test that changes an existing binding and checks the store and undo history, not only initial bind/unbind.

packages/protocol/src/api-schemas/timeline.ts
georgi: Fixed in a91920e. `mediaTracks` now round-trips through the browser store, document payload, autosave, external sync, and server patch merge. Omitted values are preserved, while an explicit `[]` clears the tracks. Regression coverage was added for both cases.

packages/timeline/src/render/sceneModel.ts
georgi: Fixed in a91920e. Preview and export now pass the document's media tracks and clips into binding resolution, including the preview compositor, browser renderer, agent preview, and video-node compositor. A browser parity test covers a bound follower.

packages/timeline/src/render/sceneModel.ts
georgi: Fixed in a91920e. Binding resolution now looks up the track owner via `track.clipId` and samples that owner's source clock. Tests cover a follower with a different start time and an owner with time remapping.

web/src/components/timeline/Inspector/ClipTracking.tsx
georgi: Fixed in a91920e. Existing bindings now commit track ID and mode changes through `bindToTrack`, preserving the remaining binding options and creating undo history. The component test verifies both edits persist and produce two undo entries.

packages/timeline/src/mediaTrack.ts
georgi: Fixed in a91920e. Smoothing now computes one EMA value per recorded sample, then interpolates between those filtered samples. This removes the jump at exact sample timestamps, with before/at/after boundary coverage.

PR #5786, opened 2026-09-15
Add generative timeline editing controls and contracts

georgi:
## Summary

- Add generative timeline editing APIs and protocol contracts.
- Implement timeline generative operations and agent evaluation coverage.
- Add inspector controls and panels for direct and generated clip workflows.
- Extend timeline, agent, and UI tests for generative editing behavior.

## Testing

- Added and updated unit and component tests covering timeline generation, agent edits, and AI edit controls.
- Full validation not run (not requested).


packages/agents/src/evals/surfaces/timeline.ts
```diff
@@ -1673,6 +1701,164 @@ export function createTimelineToolBridge(
...
+          : editRange
+            ? editRange.endMs - editRange.startMs
+            : clip.durationMs;
+        const version = makeClipVersion({
+          id,
+          createdAt: now,
+          workflowUpdatedAt: now,
+          jobId: `generative-edit:${id}`,
+          assetId: `generative://${clip.id}/${id}`,
+          status: "success",
```
georgi, line 1821: [P1] Keep fabricated successful takes out of production edit_timeline

This is not confined to evaluations: `capabilities/timelines.ts::applyOps` exposes every bridge tool except `get_clip_frames` and `generate_clip`, and `editTimeline` persists the resulting `state.documentClips`. Calling `edit_timeline` with an op such as `{ op: "generatively_edit_clip", target: "hero", operation: "restyle" }` therefore saves this `status: "success"` take even though no provider runs and no asset is persisted. Selecting that take subsequently makes `currentAssetId` point at the nonexistent `generative://...` value, breaking playback/export. Please inject a real generation/persistence implementation and append a successful take only after obtaining its asset, or exclude this operation from production until that implementation exists. A regression test should exercise the production capability/save path rather than only the in-memory bridge.

web/src/components/timeline/Inspector/AIEditControl.tsx
```diff
@@ -0,0 +1,160 @@
...
+    operation: wireOperation(operation),
+    prompt,
+    provider: clip.provider,
+    model: clip.model
+  };
+  if (operation === "extend") {
+    detail.direction = "end";
+    detail.durationMs = 1000;
+  }
+  window.dispatchEvent(new CustomEvent("nodetool:timeline-generative-edit", { detail }));
```
georgi: [P1] Connect the AI Edit event to a production generation handler

Both inspector panels submit exclusively through this emitter, but there is no production listener for `nodetool:timeline-generative-edit`; the listener added by this PR is only in `AIEditControl.test.tsx`. Neither the existing generation hook nor another changed production module consumes the request. Clicking Apply on a generated video therefore closes the menu without starting a job, adding a take, or reporting an error, for all six advertised operations. Register and lifecycle-manage a coordinator that validates and executes this request and records its result, or keep these controls unavailable until that path is implemented. The integration test needs to assert an actual generation call and candidate take, not just successful event dispatch.

web/src/components/timeline/Inspector/AIEditControl.tsx
```diff
@@ -0,0 +1,160 @@
...
+): void {
+  if (!clip.currentAssetId || typeof window === "undefined") return;
+  const detail: TimelineGenerativeEditRequest = {
+    target: clip.id,
+    clipId: clip.id,
+    sourceAssetId: clip.currentAssetId,
+    operation: wireOperation(operation),
+    prompt,
+    provider: clip.provider,
+    model: clip.model
```
georgi: [P2] Include the selected range and object track in edit requests

The emitted payload, request interface, and control props provide no `range` or `trackId`. Consequently, even after a consumer is connected, `planGenerativeOperation` rejects the UI's Replace range request with `replace_range requires a range.` and both object-edit requests with `requires trackId.` Supplying a ready current-asset MediaTrack in the document does not fix the latter because the request never identifies it. I reproduced these failures by feeding the captured event payloads into the new planner. Read/pass the applicable range and selected tracked-object ID when submitting, and disable operations whose prerequisites are unavailable. Add emitter-to-planner tests for these three operations.

packages/timeline/src/generative.ts
```diff
@@ -0,0 +1,261 @@
...
+    provider: result.provider,
+    model: result.model,
+    prompt: result.prompt,
+    negativePrompt: result.negativePrompt,
+    parentTakeId
+  };
+  const next: TimelineClip = {
+    ...clip,
+    versions: [...(clip.versions ?? []), version],
+    status: version.status === "success" ? "generated" : clip.status
```
georgi: [P2] Preserve the active clip status when only appending a candidate

With the default `activate` omitted, a successful result changes `clip.status` to `generated` even though `currentAssetId`, `activeTakeId`, and `lastGeneratedHash` all remain unchanged. For a stale clip whose `dependencyHash` differs from `lastGeneratedHash`, merely receiving an unselected candidate therefore reports the old active output as generated; a locked clip likewise becomes `{ locked: true, status: "generated" }`. I reproduced both cases using this function. Preserve the existing status when the result only appends a candidate, and derive any status change during explicit activation from the lock/dependency state rather than the candidate's success alone. Add stale and locked input cases to the take-composition tests.

georgi (commented):
Reviewed against head `71f8b50fb5ba9fb39242cc17710286b5d0a1cdfc`.

I found four correctness issues: two P1 blockers and two P2 issues. I would not merge this as-is. The inspector actions currently have no production consumer, while the new headless operation is reachable through production `edit_timeline` and can persist a successful take without generating a real asset.

The inline comments include reproduction conditions and suggested fixes. I traced the production operation dispatch, document-save path, and take selection. I also ran isolated checks using transpiled excerpts of the changed emitter/planner/take-composition functions: replace-range and both object-edit payloads fail the planner's required-input validation, and appending an unselected candidate incorrectly changes stale/locked status to generated.

These were isolated reproductions, not the repository test suites. A full checkout was blocked by GitHub DNS resolution in the execution environment, so I did not run the full unit/component suites.

PR #5787, opened 2026-09-15
feat(timeline): add interchange export

georgi:
## Summary
- add pure FCPXML and Premiere-compatible XML export for timeline sequences
- preserve active media takes, timeline timing, separate audio/video tracks, and markers
- report missing media and visual state that must be baked

## Verification
- npm run test --workspace=packages/timeline -- interchange.test.ts
- npm run build --workspace=packages/timeline
- npm run lint

## Known environment blockers
- aggregate typecheck reaches the mobile project, whose Expo dependencies are not installed
- affected tests and the capability harness need the missing better-sqlite3 native binary


packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+
+function renderFcpXml(sequence: InterchangeSequence, items: readonly {
+  track: TimelineTrack;
+  clip: TimelineClip;
+  asset: InterchangeAsset;
+  ref: string;
+}[]): string {
+  const assetLines = items
+    .map(({ clip, asset, ref }) =>
+      `      <asset id="${ref}" name="${escapeXml(asset.id)}" src="${escapeXml(asset.uri)}" duration="${time(asset.durationMs ?? 0, sequence.fps)}" hasVideo="${clip.mediaType === "video" ? "1" : "0"}" hasAudio="${clip.mediaType === "audio" || asset.hasAudio ? "1" : "0"}"/>`
```
georgi, line 114: [P1] Emit a document conforming to the declared FCPXML version

Every nonempty FCPXML export currently violates the declared 1.10 schema: `<asset>` requires one or more `<media-rep>` children, and `src` belongs on `<media-rep>`, not on `<asset>`. The generated marker fixture has a second schema violation because lines 129-132 insert `<marker>` directly under `<spine>`, whose content model does not allow markers. Apple's documentation explicitly says a nonconforming document is rejected on import. Emit `<asset ...><media-rep kind="original-media" src="..."/></asset>`, attach markers to a legal story element with the appropriate local timing, and validate generated fixtures against the full DTD rather than only asserting substrings.

Reference: [Apple's FCPXML 1.10 DTD](https://developer.apple.com/documentation/professional-video-applications/document-type-definition).

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+    })
+    .sort((left, right) =>
+      left.track.index - right.track.index || left.clip.startMs - right.clip.startMs
+    );
+}
+
+function activeAssetId(clip: TimelineClip): string | undefined {
+  const takeId = activeTakeIdOf(clip);
+  const take = clip.versions.find((version) => version.id === takeId);
+  return take?.assetId ?? clip.currentAssetId;
```
georgi, line 90: [P2] Keep currentAssetId authoritative when the take alias is stale

`activeTakeIdOf` falls back to the stored alias when `currentAssetId` is not represented in `versions`. Looking that alias up here and preferring its asset then reverses the documented source-of-truth rule in `TimelineClip`: the renderer uses `currentAssetId`, not an older recorded take. Reproduction: set `currentAssetId: "regenerated.mov"`, `activeTakeId: "old"`, and leave a successful `old.mov` version in history. Both exporters resolve `old.mov`, omit `regenerated.mov`, and return no warning. Setting the current asset to an unresolved ID also incorrectly reports `missingMedia: 0` because the old take is exported instead. Prefer an explicitly present `currentAssetId`; use a version fallback only when the current asset is absent. Add both regression cases.

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function frameDuration(fps: number): string {
+  const rounded = Math.max(1, Math.round(fps));
+  return `1/${rounded}s`;
+}
+
+function time(ms: number, fps: number): string {
+  const frames = Math.round((ms / 1000) * fps);
+  return `${frames}/${Math.max(1, Math.round(fps))}s`;
```
georgi, line 68: [P2] Preserve fractional frame rates in both target timebases

The numerator uses the actual `fps`, but the denominator and declared frame duration use its rounded integer. For `fps: 29.97` and `startMs: 600000`, the executed exporter produces `offset="17982/30s"`, which is 599.4 seconds rather than 600 seconds. Premiere has the same problem: it writes frame 17982 with `<timebase>30</timebase><ntsc>FALSE</ntsc>`. This shifts edits by 0.6 seconds at ten minutes and also changes source in-points/durations. Preserve a consistent rational rate, such as `30000/1001`, and encode the corresponding NTSC flag for xmeml. Add 23.976/29.97/59.94 long-timeline tests; the existing integer-rate tests cannot expose this.

Reference: [Apple's xmeml rate/timebase/ntsc definitions](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Elements/Elements.html).

georgi (commented):
Reviewed head `e8398678ec174180c83ed2a561f350e71d8229ed`. I would not merge this as-is. I found one P1 FCPXML import blocker and six P2 correctness or silent-loss issues, detailed inline.

Validation: executed an isolated Node harness against the exact new `interchange.ts` source (verified Git blob `dfc3fe38013efb72036adcf8dc4f677756b5012b`), using the runtime implementation of `activeTakeIdOf` from this head. Reproduced stale-take selection, fractional-frame-rate drift, silently excluded clip types, omitted playback state, absent Premiere audio components, and discarded Premiere markers. Inspected generated XML against the relevant declarations in Apple's FCPXML 1.10 DTD. I did not run the repository workspace suite or launch Final Cut Pro/Premiere, so this is not a native-import smoke test.

The three added tests only check selected strings and report fields. Add schema validation plus regression cases for these data-preservation failures; native importer/round-trip fixtures should also cover gaps and overlapping audio/video tracks.

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function exportableClips(sequence: InterchangeSequence): {
+  track: TimelineTrack;
+  clip: TimelineClip;
+}[] {
+  const byId = new Map(sequence.tracks.map((track) => [track.id, track]));
+  return sequence.clips
+    .filter((clip) => clip.mediaType === "video" || clip.mediaType === "audio")
```
georgi, line 77: [P2] Report excluded clip types before filtering them out

This filter removes image, text, shape, adjustment, group, MIDI, and model3d clips before either media resolution or `unsupported()` can inspect them. An image-only sequence with an available asset exports an empty sequence with `clipsExported: 0`, `missingMedia: 0`, and `warnings: []`; adding the same still/title to a video sequence silently drops it. Because the API explicitly tells callers to use the report to decide what needs baking, the caller cannot discover or collect these omitted elements. Supporting every type natively is not necessary for this pass, but every excluded rendered element needs an explicit per-clip warning or rejection before filtering. Add a mixed video/still/title regression that checks the report as well as the XML.

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+  const issues: string[] = [];
+  if ((clip.effects?.length ?? 0) > 0) issues.push("effects require baking");
+  if ((clip.animations?.length ?? 0) > 0) issues.push("animations require baking");
+  if (clip.generatedMatte?.status === "ready") issues.push("generated matte requires baking");
+  if (clip.transform || clip.crop || clip.mask || clip.matte) issues.push("advanced transforms require baking");
+  if (clip.blendMode && clip.blendMode !== "normal") issues.push("blend mode requires baking");
+  if (clip.opacity !== undefined && clip.opacity !== 1) issues.push("opacity requires baking");
+  if (clip.transitionIn) issues.push("transitions require baking");
+  if (clip.hidden || clip.muted || !track.visible || track.muted) issues.push("disabled media cannot be represented");
+  return issues;
```
georgi, line 103: [P2] Flag omitted playback state instead of returning a clean report

Neither serializer represents `speedMultiplier` or `timeRemap`, and this detector never warns about them. In the executed harness, adding `speedMultiplier: 2, speedBaked: false` or a reverse `timeRemap` produces byte-identical XML and an empty warning list. A two-second clip starting at source 0.5 seconds should consume four source seconds at 2x, but the Premiere output still encodes only a two-second source range and FCPXML has no time map. The caller therefore has no signal to bake the retime and gets a different edit. Audit this detector against actual rendering state: non-default `volumeDb`, audio fades, and captions likewise produced unchanged output without warnings. Emit native representations or explicit bake warnings, respecting already-baked speed via the existing source-rate/remap helpers.

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function renderPremiereXml(sequence: InterchangeSequence, items: readonly {
+  track: TimelineTrack;
+  clip: TimelineClip;
+  asset: InterchangeAsset;
+  ref: string;
+}[]): string {
+  const tracks = (mediaType: "video" | "audio") => sequence.tracks
+    .filter((track) => items.some((item) => item.track.id === track.id && item.clip.mediaType === mediaType))
```
georgi, line 142: [P2] Export the audio component of audio-bearing video assets

`tracks("audio")` only admits clips whose timeline `mediaType` is `"audio"`; `asset.hasAudio` is never consulted anywhere in the Premiere serializer. With one ordinary video clip whose resolver returns `hasAudio: true`, the generated sequence contains a video clipitem and an empty `<audio>` element, with no warning. Thus the movie's soundtrack is not placed in the exported sequence, even though the FCPXML path recognizes it. Emit corresponding audio clipitems/channel mappings, preferably linked to the video item, or explicitly report the audio as unrepresented. The existing test only supplies a separate music clip and does not cover embedded sound.

packages/timeline/src/interchange.ts
```diff
@@ -0,0 +1,191 @@
...
+      const start = Math.round((clip.startMs / 1000) * sequence.fps);
+      const end = Math.round(((clip.startMs + clip.durationMs) / 1000) * sequence.fps);
+      const inPoint = Math.round(((clip.inPointMs ?? 0) / 1000) * sequence.fps);
+      return `        <clipitem id="${ref}"><name>${escapeXml(clip.name)}</name><start>${start}</start><end>${end}</end><in>${inPoint}</in><out>${inPoint + end - start}</out><file id="file-${ref}"><name>${escapeXml(asset.id)}</name><pathurl>${escapeXml(asset.uri)}</pathurl></file></clipitem>`;
+      }).join("\n");
+      return `      <track><name>${escapeXml(track.name)}</name>\n${clips}\n      </track>`;
+    })
+    .join("\n");
+  return `<?xml version="1.0" encoding="UTF-8"?>\n<xmeml version="5"><sequence><name>${escapeXml(sequence.name)}</name><rate><timebase>${Math.round(sequence.fps)}</timebase><ntsc>FALSE</ntsc></rate><media><video><format><samplecharacteristics><width>${sequence.width}</width><height>${sequence.height}</height></samplecharacteristics></format>\n${tracks("video")}\n    </video><audio>\n${tracks("audio")}\n    </audio></media></sequence></xmeml>\n`;
```
georgi, line 154: [P2] Serialize sequence markers for the Premiere target

`sequence.markers` is never read on this path. Exporting a sequence with `{ timeMs: 2000, label: "Review & approve" }` produces exactly the same Premiere XML as exporting it without markers, and the report stays warning-free. This contradicts the PR's marker-preservation contract and silently removes editorial/review metadata. Write xmeml sequence `<marker>` elements with the label and correctly converted frame position, or report that markers are unsupported for this target. Add a Premiere marker regression; the current marker assertion only exercises FCPXML.

georgi:
Addressed in 766948ac9c.

- FCPXML assets now use media-rep; sequence markers are emitted as marker-bearing gaps rather than invalid spine children.
- currentAssetId is authoritative, including when the stored take alias is stale.
- 23.976, 29.97, and 59.94 use rational frame durations, and xmeml uses the matching NTSC timebase.
- Every non-audio/video clip and unsupported playback state now produces an explicit bake warning.
- Premiere writes sequence markers and reports embedded video audio that cannot yet be represented as a linked audio component.

Regression coverage now includes stale takes, unresolved current assets, NTSC long-timeline timing, Premiere markers, and omitted clip state. Verified with the focused timeline suite, timeline build, and lint.

PR #5819, opened 2026-09-18
fix(assets): show video thumbnails and scope Preview to its output handle

georgi:
## What changed

Video assets rendered as empty tiles in the editor's right panel and in the Library. `toAssetResponse` mints `thumb_url` for every video from a deterministic storage key without checking the object is there, and generating that object needs `ffmpeg` on PATH (`packages/websocket/src/lib/thumbnail.ts`) — so on any install without ffmpeg the URL 404s. `background-image` fails silently, and the video branch's `thumb_url || get_url` fallback was no better, because no browser paints an MP4 as a background image. Images never showed this because their `get_url` fallback *is* paintable. A new `useVideoThumbnail` hook probes the stored thumbnail once per URL and, when it does not load, decodes a poster frame through the same hidden-video/canvas extractor the timeline filmstrip uses (already shared, cached and concurrency limited); when neither yields a frame the tile shows the video icon rather than nothing.

Separately, a node with several media outputs (a video and its last frame) saved its assets under its node id alone, so a reloaded Preview wired to either handle read the same whole-node asset list and showed whichever asset the priority order picked — the reported "both Previews show the same thing". Autosave now stamps `metadata.output_name` with the top-level output handle each asset came from, and the Preview's DB fallback filters by the edge's `sourceHandle`; assets saved before this carry no handle, so those keep the previous behavior.

A third commit fixes an unrelated pre-existing CI failure that was blocking this PR — see below.

## Verification

All four mandated checks pass:

- [x] `npm run test:affected` — **exit 0**, "All affected suites passed" (127/127 Turbo tasks; web app suites 1109 + 692 passing)
- [x] `npm run typecheck` — **exit 0** (web, electron, mobile)
- [x] `npm run lint` — **exit 0**, zero errors
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — **exit 0**, 3/3 selfchecks, 310/310 graphs validate

Full `web` Jest suite: **1490 suites / 16278 tests passing, 0 failures.**

### Evidence for the video-thumbnail diagnosis

With no ffmpeg on PATH, the asset row is still written and still answers a `thumb_url`, but no thumbnail object exists:

```
WARN | nodetool.thumbnail | thumbnail generation failed
      {"contentType":"video/mp4","error":"Error: spawn ffmpeg ENOENT"}
```

After installing ffmpeg, the same path through `storeAssetWithThumbnail` with a real `testsrc` MP4 writes `u1/vid1_thumb.jpg`, 7155 bytes. So the server side is correct when ffmpeg is present and silently produces nothing when it is not — which is exactly the case the client-side fallback in this PR now covers.

### The pre-existing CI failure (fixed in `942e4d46`)

`Quality Gate / test-app` was red on `EntitySetupHost › generates a reference image from the entity description`, asserting `findByAltText("Selected entity reference")`. It was not caused by this PR: `web/src/components/setup/entity/**` is byte-identical to `origin/main`, the test fails with this PR's changes stashed, and it came in with `eb04d4e4`.

Root cause: Jest runs jsdom with `resources: "usable"`, so an `` src really is fetched — but jsdom has no image codec, so it fires `error` even for a valid `data:` URI. `ResponsiveImage` defaults to `showErrorFallback`, so a tick after mounting it replaces the `` with a broken-image icon and the alt text is gone. Probed directly: with a valid 1x1 GIF data URI the alt text is present on mount and absent 300ms later, with `data-testid="BrokenImageIcon"` in its place.

Fix: the reference slot carries a `data-testid` (which the error fallback keeps, since `ResolvedImage` spreads props onto it) and the test asserts that. The generated asset's identity is still pinned by the existing `updateAsset` assertion. The assertion was checked non-vacuous by forcing the preview branch off, which fails with `Unable to find an element by: [data-testid=entity-reference-preview]`.

### Separate finding, not fixed here

`ResolvedImage` keeps `error` in state and never resets it when `src` changes, so an image that errors once stays on the broken-image fallback even after a valid URL arrives. Confirmed by probe: after firing `error` on an empty-src image, re-rendering with a valid data URI still shows no ``. That is a real defect in a shared primitive but is outside this PR's scope and does not affect the CI failure above — happy to take it separately.

## Agent capabilities

## New checks

https://claude.ai/code/session_01T7SyBbtwD4qdBbGmLGaS6c


georgi:
## CI: `Quality Gate (npm run check) / test-app` is red on a failure this PR did not cause

`/ quality` is only the rollup — it reports `built: failure`, and the single real failure is in `test-app`:

```
Test Suites: 1 failed, 1489 passed, 1490 total
Tests:       1 failed, 3 skipped, 1 todo, 16283 passed, 16288 total

  ● EntitySetupHost › generates a reference image from the entity description
    src/components/setup/entity/__tests__/EntitySetupHost.test.tsx:248
    > 248 |   await screen.findByAltText("Selected entity reference")
```

### Why it is not this PR's

- `web/src/components/setup/entity/**` is **byte-identical** between this branch and `origin/main` — `git diff --name-only origin/main HEAD` lists only the 9 files this PR touches (assets grid, Preview node, autosave) plus the timeline-inspector files that came from main's own `73c8556d`.
- Checked out at the branch point with this PR's changes stashed, that same test fails identically. The only main commit since the branch point (`feat(timeline): optimize clip inspector`) touches `timeline/Inspector`, nothing in this path.
- The test was introduced by `eb04d4e4` (`feat(entity): allow AI reference generation`), which is upstream of this branch.
- Nothing in this diff touches `ResponsiveImage`, `useResolvedMediaUri`, or `resolveMediaUri`. The new `useVideoThumbnail` hook is imported only by `AssetItem` and `AssetListView`.

No fix for it exists on main or in an open PR, so there is nothing to port into this branch.

### Diagnosis so far, and a proposed direction

`ReferenceStep.tsx:190` renders the selected reference through `ResponsiveImage` with `showErrorFallback`. In the failure the DOM carries `data-testid="BrokenImageIcon"` where the image should be — the component is on the right branch (`assetId` is set) but no `` is mounted, so `findByAltText` has nothing to match.

Resolution itself looks sound on paper: the locator is `asset://generated-asset`, `useResolvedMedia` routes it through `useAssetStore.get` → `trpcClient.assets.get.query`, which the test mocks to an asset whose `get_url` is a valid `data:` URI, and `asResolvedMediaUrl` brands any non-empty string. That points at the error fallback being reached from the ``'s own `error` event rather than from failed resolution — jsdom never actually loads image bytes, so asserting on a rendered `` behind `showErrorFallback` is inherently fragile.

Two options, either contained to that suite:

1. Stub `ResponsiveImage` in `EntitySetupHost.test.tsx` the way the file already stubs `ImageRefPreview`, and assert the locator it received.
2. Assert on the reference region rather than on a live `` (e.g. a testid on the container), so the assertion does not depend on jsdom mounting image bytes.

I have deliberately **not** pushed either: this is another feature's test and outside what this PR was asked to change. Happy to take it in a separate PR if you want it — say the word.

### On re-running

I am not spending the one re-run on it. The failure is deterministic, not a flake: it reproduces locally on source identical to `origin/main`, and the job died well after test bodies ran (417s, 16283 passing). A re-run would repeat the same assertion failure and burn ~7 minutes of CI.

This PR's own surfaces are green: 1489 web suites pass, `packages/websocket/tests/asset-autosave-edges.test.ts` passes 34/34 including the new output-handle case, and lint reports nothing new in the touched files. I will keep watching this PR.

---

georgi:
Update to my comment above: the author asked me to fix that failure, so I did — it is no longer a stand-down.

Root cause was not what the assertion suggested. Jest runs jsdom with `resources: "usable"`, so an `` src really is fetched, but jsdom carries no image codec and fires `error` even for a valid `data:` URI. `ResponsiveImage` defaults to `showErrorFallback`, so a tick after mounting it swaps the `` for a broken-image icon and the alt text is gone. Probed directly: with a valid 1x1 GIF data URI the alt text is present on mount and absent 300ms later, with `data-testid="BrokenImageIcon"` in its place. The test could never have passed in this environment.

`942e4d46` gives the reference slot a `data-testid` — which the error fallback keeps, since `ResolvedImage` spreads props onto it — and asserts that instead. The generated asset's identity is still pinned by the existing `updateAsset` assertion. Checked non-vacuous: forcing the preview branch off fails with `Unable to find an element by: [data-testid=entity-reference-preview]`.

CI on `942e4d46` is green across all four workflows — Test (Quality Gate), E2E Workflow Runner, Page Load Smoke, User Journeys — and the PR is mergeable with no conflict.

All four mandated checks now pass locally as well; the npm registry outage that blocked them earlier has cleared:

- `npm run test:affected` — exit 0, "All affected suites passed" (127/127 Turbo tasks)
- `npm run typecheck` — exit 0 (web, electron, mobile)
- `npm run lint` — exit 0, zero errors
- `npm run dev:nodetool -- harness gate --base origin/main` — exit 0, 3/3 selfchecks, 310/310 graphs

Two failures I hit locally on the way were environment gaps that AGENTS.md documents as "do not skip", not code: `image-nodes` needed a Vulkan ICD (`mesa-vulkan-drivers`) and `video-nodes` needed real ffmpeg. With both installed those suites pass (266 and 7 tests).

Installing ffmpeg also let me close the loop on this PR's own diagnosis: a real `testsrc` MP4 through `storeAssetWithThumbnail` writes `u1/vid1_thumb.jpg` at 7155 bytes, where the same call without ffmpeg logs `spawn ffmpeg ENOENT` and writes nothing while `thumb_url` is still minted.

The `ResolvedImage` sticky-error defect noted in the PR description is still unfixed and still out of scope here.

---

PR #5824, opened 2026-09-19
Entities: stop 404ing the save read on a cross-project asset

georgi:
## What changed

The entity guided flow ended with `assets.get` returning `NOT_FOUND`. `useSaveEntity` read the asset it was about to tag with `project_id` set to the *destination* project, but the asset being tagged normally lives somewhere else at that moment: the flow's blank reference and every picker upload land in the loose bucket, and filing the asset under the project is the `projects.assignDocument` write further down the same mutation. Since the server scopes `assets.get` to a project when the input names one, that read 404ed on exactly the assets the mutation exists to move. `useDeleteEntity` carried the same read, scoped to the active project, so untagging an entity shown from another project failed the same way. Both reads now go unscoped; ownership is still enforced server-side by `Asset.find(ctx.userId, id)`, so nothing outside the caller's own library becomes reachable.

## Verification

New test `web/src/serverState/__tests__/useEntities.test.tsx` gives `assets.get` a fake that reproduces the server's project scoping. Both cases fail with the reported error before the fix:

```
● useSaveEntity › tags an asset that is not yet in the destination project
    Asset not found
      at Object.mutationFn (src/serverState/useEntities.ts:197:49)
● useDeleteEntity › untags an entity that lives outside the active project
    Asset not found
      at Object.mutationFn (src/serverState/useEntities.ts:256:49)
Tests: 2 failed, 2 total
```

After the fix: `Tests: 2 passed, 2 total`.

- [x] `npm run test:affected` — 300 suites, 2699 passed, 3 skipped
- [x] `npm run typecheck` — web and electron clean; the mobile pass fails only on its own uninstalled dependency tree (`expo`, `react-native`), untouched by this diff
- [x] `npm run lint`
- [x] `npm run dev:nodetool -- harness gate --base origin/main` — 1/1 selfchecks passed (76 files, 1379 tests)

## Agent capabilities

## New checks

https://claude.ai/code/session_016L1GDjtmV44N3hwwt2jVrE

---
_Generated by [Claude Code](https://claude.ai/code/session_016L1GDjtmV44N3hwwt2jVrE)_


georgi:
`Drive the app the way a user does` (User Journeys) failed on `ebc2cda`: 7 journey tests timed out on left-rail navigation, all at `pages.ts:265` waiting for `getByRole('button', { name: /^Apps\b/ })` after clicking `More` — `mini-app.spec.ts` (3), `library.spec.ts` (3), and the mini-app accessibility case.

Not this PR's: the diff touches only `web/src/serverState/useEntities.ts` and a new Jest test, and the same job failed the same way on this PR's base commit `fa9a99c` ([run 35442442980](https://github.com/nodetool-ai/nodetool/actions/runs/35442442980)). The rail moved Apps and Workflows under Documents, and the page objects still drove the old `More → Apps` path.

The fix already exists on `main` as `d1569c2` ("test(journeys): follow current document navigation"), which rewrites exactly those locators in `web/tests/journeys/pages.ts`. I merged `main` into this branch (`496ea9c`) to carry it, so CI re-runs against the corrected page objects; the merge is a no-op for this PR's own change. Validated before pushing: `npm run typecheck:playwright` clean, `npm run test:affected` 300 suites / 2699 passed, `npm run lint` clean.

---

